PR #7, opened 2021-03-10
chore: fix is404 value

sibbng:


brillout:
Thank you @sibbng.

Let me know if you have any questions about vite-plugin-ssr.

PR #13, opened 2021-03-21
Add React + Apollo example

stephensamra:
As discussed in https://github.com/brillout/vite-plugin-ssr/issues/8, here is an example that uses React and Apollo SSR.


brillout:
Neat.

Is there a way to move `new ApolloClient` outside of `  app.get("*", async (req, res, next) => {`? Doesn't seem right to initialize the client over and over again. Also I guess we can move it outside `async function hydrate() {` (doesn't make a difference but it's more idiomatic).

Also, I'd reduce the example to a bare minium:
 - No logo
 - No PageLayout
 - JS instead of TS
 - Etc.

If someone looks at the `examples/apollo` example, he'll want to see integration stuff and not unrelated things. (I'm actually planning to replace the `examples/react` example with a bare minimum one.)

Sorry for the late reply, I was heads down with implementing Client-side Routing. I normally reply within a day.

stephensamra:
I've simplified the example quite a bit and removed some unnecessary code/packages.

> Is there a way to move new ApolloClient outside of `app.get("*", async (req, res, next) => {`

This is actually done on purpose. According to [Apollo's docs on SSR](https://www.apollographql.com/docs/react/performance/server-side-rendering/#example), it is recommended that you create a new client per request:

> It's important to create an entirely new instance of Apollo Client for each request. Otherwise, your response to a request might include sensitive cached query results from a previous request.

Let me know if you have any other feedback.
Thanks.

brillout:
> This is actually done on purpose. According to Apollo's docs on SSR, it is recommended that you create a new client per request:

I guess we should add a comment about that on top of the client initilization code, at both places.

The test is not working. The react and vue tests are quite heavy; no need for such heavy test here and we can do something simple like `examples/redux/.test.spec.ts`.

I added instructions on how to run tests at https://github.com/brillout/vite-plugin-ssr/blob/master/CONTRIBUTING.md

We can also squeeze out more stuff, e.g. the scripts at `package.json` (see the redux example) and we can remove `const title = contextProps.title || "Demo: vite-plugin-ssr";`. Also we can remove `_error.page`.

LGTM otherwise.

Let me know if you run into any paper cuts while developing the example. I'm interested in a well polished contributing DX.

brillout:
@stephensamra Any luck? If you are busy let me know and I'll merge and finish it up. Thanks for the PR btw.

stephensamra:
Hey @brillout, I am a little busy but I plan on doing this sometime over the weekend. I'll update you either way. Thanks.

brillout:
Sounds good

stephensamra:
Hey @brillout, what version of node/npm should be used for this project? I'm currently using node v14.3.0 and npm v6.14.4 but I think that's causing some issues when trying to run npm install in the root directory (following https://github.com/brillout/vite-plugin-ssr/blob/master/CONTRIBUTING.md#run-test-suite):
```
This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```
I can switch versions using `nvm` but wondering what version I should switch to? Thanks.

brillout:
I believe npm v7 should do the trick, as they did a lot of breaking changes between v6 and v7.

My setup: node: `7.0.14`, npm: `v15.3.0`.

brillout:
You can use `npm test -- --root-dir examples/some-example/` instead of `npm test`. It will save you a lot of time. https://github.com/brillout/vite-plugin-ssr/blob/master/CONTRIBUTING.md#modify-example

stephensamra:
Thanks @brillout, that's working perfectly now 👍

stephensamra:
@brillout, I've cleaned up the example a bit more. The test is passing now, too. Let me know if it's all good or if you think there should be any more changes.

brillout:
Perfect.

Let me know if you have questions or critical(!) feedback regarding vps.

PR #15, opened 2021-03-24
feat: allow skip git init

thomasruiz:
Quick PR to allow the developer to skip the git repo initialization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,8 @@ yarn create vite-plugin-ssr
 
 Then choose between `vue`, `vue-ts`, `react`, and `react-ts`.
 
+You can use the `--skip-git` flag to skip the initialization of a new git repository.
+
 <br/><br/>
 
 
--- a/boilerplates/index.js
+++ b/boilerplates/index.js
@@ -108,7 +108,9 @@ async function init() {
     write(file)
   }
 
-  initGitRepo(root)
+  if (!argv['skip-git']) {
+    initGitRepo(root)
+  }
 
   console.log(`\nDone. Now run:\n`)
   if (root !== cwd) {
@@ -161,14 +163,17 @@ function initGitRepo(cwd) {
   if (!isGitInstalled()) {
     return
   }
+
   execSync('git init', { cwd })
   execSync('git add .', { cwd })
   execSync(
     [
-      'git commit',
-      '--message="boilerplate Vite w/ vite-plugin-ssr"',
-      '--author="Romuald Brillout <git@brillout.com>"',
-      '--date="Mon Mar 1 14:45:24 2021 +0100"'
+      'git',
+      '-c user.name="Romuald Brillout"',
+      '-c user.email="git@brillout.com"',
+      'commit',
+      '--no-gpg-sign',
+      '--message="boilerplate Vite w/ vite-plugin-ssr"'
     ].join(' '),
     { cwd }
   )


brillout:
I'm glad you did this. I've been wondering: why would someone want to skip git initialization?

thomasruiz:
Well, in my case, I have 2 email addresses, on for my company, and one for my other professional uses. In my global git config, I don't have my [user.email] set up, so I get the following error if I try to commit:

```
Author identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: no email was given and auto-detection is disabled
```

I always set the correct email before working on a new repo.

I guess there is also the case when people want to work with something other than git?

brillout:
> I get the following error

I did think of this:

```js
  execSync(
    [
      'git commit',
      '--message="boilerplate Vite w/ vite-plugin-ssr"',
      '--author="Romuald Brillout <git@brillout.com>"',
      '--date="Mon Mar 1 14:45:24 2021 +0100"'
    ].join(' '),
    { cwd }
  )
```

I'm surprised `git` still throws an error even though both the author name and author email are provided.

> I guess there is also the case when people want to work with something other than git?

If `git` is not installed, then there is no git initialization.

thomasruiz:
> I'm surprised `git` still throws an error even though both the author name and author email are provided.

Apparently, there is a difference between a commit author and the "committer identity". With the `--author` I get this:

```
Committer identity unknown
```

> If `git` is not installed, then there is no git initialization.

Yeah I saw that, but who doesn't have git installed by default? I know a few people who use mercurial for a specific project, but have git installed nonetheless.

brillout:
> Apparently, there is a difference between a commit author and the "committer identity". With the --author I get this:

Is there a fix around this? Would be neat to support people who don't have a global config.

> I know a few people who use mercurial for a specific project, but have git installed nonetheless.

But would they *start* a new project with mercurial?

thomasruiz:
> Is there a fix around this? Would be neat to support people who don't have a global config.

Apparently, you can use `-c` with `git` to use temporary git config. Using this should work?

`git -c user.name='Romuald Brillout' -c user.email='git@brillout.com' commit -m 'boilerplate Vite w/ vite-plugin-ssr'`

> But would they _start_ a new project with mercurial?

Maybe, who knows? :)

There are a few more issues for the first commit too: no README.md, no LICENSE.md,... I mean, usually, I end up amending the first commit before pushing anyway, and I can't be the only one who does that.

brillout:
> Apparently, you can use -c with git to use temporary git config

Should work I guess? Would you be up to update your PR to include this 😍 ?

> There are a few more issues for the first commit too

Ok you convinced me. But maybe we should use a `--skip-git` flag instead of a pesky prompt.

thomasruiz:
And there you go. I added the `--skip-git` flag which is actually a better option, but I don't know where to add it in the docs.

Also, I disabled gpg signing on the commit, otherwise it just fails again.

brillout:
Perfect, how about here https://github.com/brillout/vite-plugin-ssr#boilerplates

thomasruiz:
All done!

brillout:
Released.

`npm i -g create-vite-plugin-ssr@0.0.14 && npm init vite-plugin-ssr` or `yarn create vite-plugin-ssr`. (Only yarns auto busts the cache.)

brillout:
Btw. I'm curious: what is it you 1. like and 2. dislike about `vite-plugin-ssr`?

thomasruiz:
I'm just playing around right now, but I like how easy to use it is.

PR #21, opened 2021-03-30
Fix typo in `navigate` error

chrisvariety:
Just a small thing I noticed!

Also note the Readme appears to be incorrect in this area, it states `In Node.js, navigate() is a no-op.`, which is not correct.


brillout:
Neat.

> Also note the Readme appears to be incorrect in this area, it states In Node.js, navigate() is a no-op., which is not correct.

Fixed.

PR #23, opened 2021-03-31
indicate Node 15 requirement in contribution docs

gryphonmyers:
The project uses Node workspaces which were added in Node 15. I had my default set to Node 14 and the setup instructions didn't work. Probably best to indicate this requirement to potential contributors.


brillout:
👍

PR #30, opened 2021-04-05
fix: allow fetching new page props before url change

gryphonmyers:
This PR allows for fetching page props for a prospective new URL _before the URL has changed_. This is ideal/required when fetching data in client-side navigation hooks that occur before the navigation has completed.

Additionally, I removed the concept of "navigationChanged" for the following reasons:

- When our URL matches the initial URL, we probably DO want to serve the values cached to the DOM by the SSR process. E.g. even if we navigate off of the initial route, then go back to it, why would we make a network request for that data when it is still sitting in DOM (as it was when the app initialized)?
- It seemed strange to be setting navigation state in a method responsible for fetching data ("getPageProps" does not sound like it would/should have side effects).

### DIFF
--- a/src/client/navigationState.client.ts
+++ b/src/client/navigationState.client.ts
@@ -10,6 +10,9 @@ export const navigationState = {
   get isFirstNavigation() {
     return !navigationChanged && this.checkIfOriginalUrl(this.urlNow);
   },
+  checkIfCurrentUrl(url : string) {
+    return url === this.urlNow;
+  },
   checkIfOriginalUrl(url : string) {
     return url === this.urlOriginal;
   },
--- a/src/client/navigationState.client.ts
+++ b/src/client/navigationState.client.ts
@@ -1,14 +1,13 @@
 import { getUrlPathname } from './getUrl.client'
 
-let urlOriginal = getUrlPathname()
-let navigationChanged = false
+let urlOriginal = getUrlPathname();
 
 export const navigationState = {
-  markNavigationChange() {
-    navigationChanged = true
-  },
   get isFirstNavigation() {
-    return !navigationChanged && this.urlNow === this.urlOriginal
+    return this.checkIfOriginalUrl(this.urlNow);
+  },
+  checkIfOriginalUrl(url : string) {
+    return url === this.urlOriginal;
   },
   get urlOriginal() {
     return urlOriginal
--- a/src/client/navigationState.client.ts
+++ b/src/client/navigationState.client.ts
@@ -1,10 +1,14 @@
 import { getUrlPathname } from './getUrl.client'
 
 let urlOriginal = getUrlPathname();
+let navigationChanged = false
 
 export const navigationState = {
+  markNavigationChange() {
+    navigationChanged = true
+  },
   get isFirstNavigation() {
-    return this.checkIfOriginalUrl(this.urlNow);
+    return !navigationChanged && this.checkIfOriginalUrl(this.urlNow);
   },
   checkIfOriginalUrl(url : string) {
     return url === this.urlOriginal;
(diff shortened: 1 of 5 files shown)


brillout:
> E.g. even if we navigate off of the initial route, then go back to it, why would we make a network request for that data when it is still sitting in DOM (as it was when the app initialized)?

The idea here is to replicate what the browser would normally do. This is what the end-user expects, as it is what >98% of websites do. We can discuss an opt-in feature to tell vite-plugin-ssr to cache all pages and serve the cache whenever possible. (Out of consistency we should then cache all pages and not just cache the first page). If you don't need this urgently, then I'd suggest to open a new feature-request GitHub issue and see how many want this.

> This PR allows for fetching page props for a prospective new URL before the URL has changed

👍 Makes sense.

gryphonmyers:
> The idea here is to replicate what the browser would normally do. This is what the end-user expects, as it is what >98% of websites do.

Ok that's fair, how about this - I'll restore the functionality but remove the `markNavigationChange` call from `getPageProps`? It would then be up to a higher level function to call markNavigationChange (I just really don't think it should be this function...)

brillout:
I agree, and, if you can find a better design I'm all for taking it :-). I was actually struggling to make it intelligible and was and still am not happy about the current result.

gryphonmyers:
Can I suggest this might be the place that makes the most sense to set that state? https://github.com/brillout/vite-plugin-ssr/blob/master/src/client/router/useClientRouter.client.ts#L44

I will also note that part of why you may not have noticed the issue this PR addresses is the order of operations here - you're first changing the URL, then fetching and rendering the page for that URL. I'd suggest it should be in the reverse order. This is how routers typically work - you set the url as the last thing, once you're sure the navigation succeeded. For example, if there is an error during `fetchAndRender` we should not be on the new URL, we should be on the URL we were at before we encountered the error.

gryphonmyers:
^ I went ahead and shifted some bits around. I think this is better because we can explicitly tell the lower level function to use or not use the cached data, then we just use that parameter in the higher level function. This works for my purposes and hopefully preserves the existing functionality of the higher level routing solution.

BTW I am not taking any action on the order of operations mentioned above - I think it's out of scope for this PR.

brillout:
Perfect 👍 it's much better than before.

> we should be on the URL we were at before we encountered the error
> I'd suggest it should be in the reverse order

You're right, makes sense.

brillout:
Btw I renamed the commit messages to remove `fix:`. The idea of `fix:`  is to populate the `CHANGELOG.md`. This means `fix:` should only be used for commits that affect `vite-plugin-ssr` users. This PR only does internal changes that are have no effect whatsoever on `vite-plugin-ssr` users.

brillout:
FYI I improved the variables names: https://github.com/brillout/vite-plugin-ssr/commit/32302bbff93e74c85afec5f64e7c51fec90b30c1.

brillout:
Also, I changed the order; the URL is now changed after the new page data is fetched and right before the rendering of the new page.

gryphonmyers:
@brillout thanks - if you wouldn't mind publishing again, I'd love to get a version with these changes

brillout:
Any reason why `npm run link` doesn't suit you? This would be a big win for you as you'll then be able to change `vite-plugin-ssr` as you wish. If you want, you can still do all your changes in a PR draft so I can help out here and there.

https://github.com/brillout/vite-plugin-ssr/blob/master/CONTRIBUTING.md#modify-vite-plugin-ssr and then try the changes on `/examples/vue-router/` or an entirely new example.

gryphonmyers:
It does but I can't deploy with npm link

brillout:
Do you mean to deploy an app or deploy a new `vite-plugin-ssr` version?

My thinking here is that we would release a new `vite-plugin-ssr` version after we got a working POC.

PR #38, opened 2021-04-12
docs: Add alternative WindiCSS config

CanRau:
As a follow up to https://github.com/brillout/vite-plugin-ssr/pull/37


brillout:
I think we can show the `vite.plugin.js` example only and add a comment that it's also possible to define in `windi.config.js`.

brillout:
🙏

PR #40, opened 2021-04-13
Routing overridability

gryphonmyers:


gryphonmyers:
I see the following options for how we expose configuration of the matcher: 

- Plugin interface exposed when instantiating the plugin (e.g. pass `ssr({matchRoutes: vueRouterMatchRoutes})` into vite config).
- Additionally supported matchers are built into the source and configurable via enum (e.g. `ssr({ routeMatcher: 'vue-router' })`).
- This is sort of in addition to either of the above approaches, but perhaps users could select from matchers in specific *.route.js files (e.g. `export { route: '/:myParam?', matcher: 'vue-router' }`). This should allow for using multiple matchers within the same application, which would preserve the open-endedness that exists in other parts of the plugin.

brillout:
Looking forward to have a look at all of this!

> I do have a basic proof of concept working end-to-end

Can I see this somewhere? I'd love to see the end result of your work.

gryphonmyers:
Here let me do this - I'll get this setup working in the vue-router example and make those changes part of this PR.

gryphonmyers:
@brillout ^

brillout:
> Here let me do this - I'll get this setup working in the vue-router example and make those changes part of this PR.

Perfect.

I will finish my context props work and then have a look at your PR. Very curious to see the deep integration.

brillout:
Couple of notes:
 - I agree with some of your refactoring but let's focus this PR on vue integration. Including refactoring in a PR: makes it harder for me to comprehend what the PR changes, and makes it harder to resolve merge conflicts (which will actually occur after I merge my latest to changes to master).
 - We can create a seperate refactor PR before or after this PR, whatever works best for you. In general, refactor PRs should be merged very quickly to minimize merge conflicts.
 - I agree with splitting `route.shared.ts` into smaller files. But there are too many files; many files have only ~10 LOC functions. Instead we can create a `src/routing/utils/` that contains self-contained-low-level functions (similarly to `src/utils/`). The logic of each file in `src/routing/utils/` should be self contained. Having many files in `src/routing/utils/` is fine but `src/routing/` should contain only few files that represent what happens on a high-level.
 - Let's keep name conventions: "Route Strings" (instead of "Static Routes"), "Route Functions" (instead of "Functional Routes"), name files should be camelCased and if they contain only one export they should be named after the export.

brillout:
I suggest the following next steps:
 1. Let's move all Vue Router Integration code to a temporary package `@vite-plugin-ssr/vue-router` (I will add you to the npm org) that lives in this repository in a `vue-router/` directory (=> `package.json#workspaces.includes('vue-router/')`). We can then move it to a separate repository (or not, as we wish).
 2. Revert non-essential changes that have been made to `src/`. E.g. the refactoring of the code.
 3. Let's make the test suite go green. If you run `npm run test:setup && npm run test` you'll see that the test suite fails.

> I see the following options for how we expose configuration of the matcher

The Vite plugin is for building which we don't need to change. Instead, AFAICT, the following should work.

```js
// _default.client.js

import { useVueRouter } from '@vite-plugin-ssr/vue-router/client';

// ...
```
```js
// server/index.js

import { createPageRender } from 'vite-plugin-ssr'
import { useVueRouter } from '@vite-plugin-ssr/vue-router';

const renderPage = createPageRender({ viteDevServer, isProduction, root, customRouting: useVueRrouter });

// ...
```

`src/` would expose functions to override default behavior that `@vite-plugin-ssr/vue-router` uses. For example:

```js
export { changeRouteStringResolver }

let customResolver
function changeRouteStringResolver(customResolver_) {
  customResolver = customResolver_
}

function resolveRouteString(...args) {
   if( customResolver ) {
      return customResolver(...args)
   }
   // Default behavior
   // ...
}
```

Looking forward to have Vue Router deeply intergrated!

examples/vue-router/App.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+  <div>
+    <nav>
+      <router-link class="nav-route" :to="{ name: '/pages/about' }">
```
brillout, line 4: Does `<router-link to="/about">About</router-link>` also work?

src/client/getPage.client.ts
```diff
@@ -40,17 +41,20 @@ async function getPageById(pageId: string): Promise<any> {
 function getPageInfo(): {
   pageId: string
   pageProps: Record<string, unknown>
+  routes: PageRoute[]
 } {
   const pageId = window.__vite_plugin_ssr.pageId
   const pageProps = window.__vite_plugin_ssr.pageProps
-  return { pageId, pageProps }
+  const routes = window.__vite_plugin_ssr.routes
```
brillout, line 48: I don't think we need this; `loadPageRoutes()` provides all the routing information of all pages.

examples/vue-router/App.vue
gryphonmyers: yes

src/client/getPage.client.ts
gryphonmyers: Route functions will be returned as-is from loadPageRoutes. This is unusable in the client because the functions depend on contextProps, which are not available in client code.

gryphonmyers: It also means that the routes can vary on a page-by-page basis. I would much prefer if we could simply load one set of routes for the entire application, but unfortunately that is not possible with the features you've exposed.

gryphonmyers:
> Couple of notes:
> 
> * I agree with some of your refactoring but let's focus this PR on vue integration. Including refactoring in a PR: makes it harder for me to comprehend what the PR changes, and makes it harder to resolve merge conflicts (which will actually occur after I merge my latest to changes to master).
> * We can create a seperate refactor PR before or after this PR, whatever works best for you. In general, refactor PRs should be merged very quickly to minimize merge conflicts.
> * I agree with splitting `route.shared.ts` into smaller files. But there are too many files; many files have only ~10 LOC functions. Instead we can create a `src/routing/utils/` that contains self-contained-low-level functions (similarly to `src/utils/`). The logic of each file in `src/routing/utils/` should be self contained. Having many files in `src/routing/utils/` is fine but `src/routing/` should contain only few files that represent what happens on a high-level.
> * Let's keep name conventions: "Route Strings" (instead of "Static Routes"), "Route Functions" (instead of "Functional Routes"), name files should be camelCased and if they contain only one export they should be named after the export.

The thing is, there are parts of this refactor that are required for the integration to work. If you'd like me to recombine some of these files into single-use functions within their respective files, I can do that. I can also help navigate any merge conflicts and fix tests.

The real problem that the refactor seeks to address in order to make the integration possible, is that you have baked the matching logic into the route function. For a Vue Integration to work, we need to rely on their matching logic. This is because I can't parse a Vue route and get back `{ matchValue: boolean|number, routeParams: Object }` in order to do our own matching on the parsed result, nor, I would argue, should I. What we should be doing, and what this refactor accommodates, is allowing us to pass the routes into vue router, then let it use its own matching logic in order to tell us what route we are now on. 

For these reasons, your suggestion of exposing `changeRouteStringResolver` will not suffice. This is why I have been speaking in terms of a `matchRoutes` function. We need to expose a function that takes all the routes and gives us back a single match.

gryphonmyers:
> I suggest the following next steps:
> 
> 1. Let's move all Vue Router Integration code to a temporary package `@vite-plugin-ssr/vue-router` (I will add you to the npm org) that lives in this repository in a `vue-router/` directory (=> `package.json#workspaces.includes('vue-router/')`). We can then move it to a separate repository (or not, as we wish).
> 2. Revert non-essential changes that have been made to `src/`. E.g. the refactoring of the code.
> 3. Let's make the test suite go green. If you run `npm run test:setup && npm run test` you'll see that the test suite fails.
> 
> > I see the following options for how we expose configuration of the matcher
> 
> The Vite plugin is for building which we don't need to change. Instead, AFAICT, the following should work.
> 
> ```js
> // _default.client.js
> 
> import { useVueRouter } from '@vite-plugin-ssr/vue-router/client';
> 
> // ...
> ```
> 
> ```js
> // server/index.js
> 
> import { createPageRender } from 'vite-plugin-ssr'
> import { useVueRouter } from '@vite-plugin-ssr/vue-router';
> 
> const renderPage = createPageRender({ viteDevServer, isProduction, root, customRouting: useVueRrouter });
> 
> // ...
> ```
> 
> `src/` would expose functions to override default behavior that `@vite-plugin-ssr/vue-router` uses. For example:
> 
> ```js
> export { changeRouteStringResolver }
> 
> let customResolver
> function changeRouteStringResolver(customResolver_) {
>   customResolver = customResolver_
> }
> 
> function resolveRouteString(...args) {
>    if( customResolver ) {
>       return customResolver(...args)
>    }
>    // Default behavior
>    // ...
> }
> ```
> 
> Looking forward to have Vue Router deeply intergrated!

1. Sounds good. I'd suggest keeping everything in this repository as a monorepo.
2. I'd like to get more clarity between on what is considered "non-essential." Please take a look at my previous comment. If you'd like to discuss further perhaps we could take the conversation to Discord?
3. Can do- would like to get point 2 resolved first so we can get consensus on the solution before I fix the tests.

The code structure you're proposing should work, with the caveat that as I said in my previous comment, `changeRouteStringResolver` will not suffice as the override. I think we're overriding `matchRoutes` and possibly `sortRoutes`.

src/client/getPage.client.ts
brillout: How about forbidding Route Functions when using Vue Router? I'd rather have a simple MVP first then we can see whether we want more features.

If we remove Route Function support, we can get rid of `window.__vite_plugin_ssr.routes` and use `loadPageRoutes()` instead right?

brillout:
> For these reasons, your suggestion of exposing changeRouteStringResolver will not suffice. This is why I have been speaking in terms of a matchRoutes function. We need to expose a function that takes all the routes and gives us back a single match.

`changeRouteStringResolver` was just an example of how we can make `@vite-plugin-ssr/vue-router` override default behavior; you're right, we need to override more things.

> I'd like to get more clarity between us on what is considered "non-essential."

Mainly the moving of functions into their own file. IIRC reverting this should do it. If there is anything else I can let you know. Just ping me when there are changes on the PRs.

Let me know if you need anything else.

brillout:
> Sounds good. I'd suggest keeping everything in this repository as a monorepo.

👍. I'll add you to the repo.

gryphonmyers:
Ok! I'll only split out the functions that seem like they're begging for it and let you know when that's complete.

brillout:
Perfect. I merged my changes to master; as expected there are merge conflicts. See the CHANGELOG.md to see why the changes have been made. As usual, let me know if you have any question.

src/client/getPage.client.ts
gryphonmyers: I simplified this bit after merging in your contextProps changes. Instead of separately handling routes, I just made it a built-in part of the context (similar to url) and pushed it into the `passToClient` array so the client will always have it in context. 

> If we remove Route Function support, we can get rid of window.__vite_plugin_ssr.routes and use loadPageRoutes() instead right?

I already removed `window.__vite_plugin_ssr.routes` as mentioned above. We do still need to keep a copy of the routes in the context for all rendered paths, because as I mentioned before there is a possibility the user is using page-specific context in the creation of that route. If you wanted to avoid this overhead, removing the route functions feature would allow us to use a single set of routes for the entire application. We would still need to get that set of routes to the client somehow, so if we're removing the routes from the contextProps, then we'd need another solution.

brillout: > If you wanted to avoid this overhead, removing the route functions

Yes, that's what I have in mind. I'm thinking of doing some simplifications to Route Functions so I'd suggest to add the Route Function feature to Vue Router at a later point.

>  We would still need to get that set of routes to the client somehow

I believe `loadPageRoutes()` is enough. It loads all the route information we need (inlcuding all route strings and all route functions) on the client.

gryphonmyers: > I believe loadPageRoutes() is enough. It loads all the route information we need (inlcuding all route strings and all route functions) on the client.

But currently the route functions need to be compiled on the server as they expect a full contextProps object. Are you proposing changing it so that only client-available context props are available to route functions? 

There is still the problem of the return value of these functions being something Vue Router doesn't know what to do with. Even if I am able to compile the route function in the client, there's not a whole lot I can do with `{ matchValue: number, routeParams: Object }` since Vue Router is the matcher. If the functions were to instead return dynamically generated paths, then we could recompute all the functions on navigation and update the routes. That is what I am currently supporting in the vue router plugin. If the route function gets a non-string, it will throw an error.

gryphonmyers:
@brillout I've recombined some of the routing functions (I looked for anything that was single use and added it back the file where it was being used). I also merged from master and resolved the merge conflicts, did some various clean up / refactor... and also created a package for vue router as you described. Please have a look.

I suppose the last thing to figure out is this stuff about the route functions, whether we're ending support of some part of that feature to make client routing integration easier.

src/client/getPage.client.ts
brillout: What I'm suggesting is to completely forbid the usage of Route Functions when using the Vue Router plugin. I suggest to add this functionality later.

`loadPageRoutes()` should be sufficient then.

> only client-available context props are available to route functions?

FYI this is actually already the case. (That's actually a problem for some use cases and I've plans to solve this.)

src/route.shared.ts
```diff
@@ -82,131 +71,52 @@ function getErrorPageId(allPageIds: string[]): string | null {
...
-  const commonPath = slice(pathsPart, 0, -1).join('/') + '/'
-  return commonPath
+  return compiledRoutes
+    .sort(sortRoutes)
 }
 
-/**
-  Returns the ID of all pages including `_error.page.*` but excluding `_default.page.*`.
-*/
-async function getPageIds(): Promise<PageId[]> {
```
brillout, line 168: Did something chagne here? Can we keep it the function in the file for now? (Because right now I've to compare each line to see if there is a diff.)

src/route.shared.ts
```diff
@@ -82,131 +71,52 @@ function getErrorPageId(allPageIds: string[]): string | null {
...
-  const pageViewFiles = await getPageFiles('.page')
-  let pageViewFilePaths = pageViewFiles.map(({ filePath }) => filePath)
-  pageViewFilePaths = pageViewFilePaths.filter((filePath) => !isDefaultPageFile(filePath))
-
-  let allPageIds = pageViewFilePaths.map(computePageId)
-  return allPageIds
-}
-function computePageId(filePath: string): string {
```
brillout, line 176: See above

src/route.shared.ts
```diff
@@ -82,131 +71,52 @@ function getErrorPageId(allPageIds: string[]): string | null {
...
-  const pageSuffix = '.page.'
-  const pageId = slice(filePath.split(pageSuffix), 0, -1).join(pageSuffix)
-  return pageId
-}
-
-function isDefaultPageFile(filePath: string): boolean {
```
brillout, line 182: Same as above

src/route.shared.ts
```diff
@@ -235,39 +145,7 @@ function resolveRouteFunction(
   }
 }
 
-type PageRoute = {
-  pageRouteFile: string
-  pageRoute: string | Function
-}
-async function loadPageRoutes(): Promise<Record<PageId, PageRoute>> {
```
brillout, line 242: Did something change here or did you just move the code? That's why such code move should live in a separate PR. Not only is it time consuming to compare each line but, more importantly, it's error prone as I may miss some subtle change.

brillout (changes requested):

brillout:
Could we maybe keep all changes in `route.shared.ts` and remove the `src/routing/` directory for now? We can move code later.

brillout:
> created a package for vue router

Neat

src/route.shared.ts
gryphonmyers: Ok I understand punting the change for the sake of readability, so I'll go ahead and do that, but these refactors will need to occur for performance reasons. All client dependencies need to be kept lean. You shouldn't be shoving all functions into one file.

src/client/getPage.client.ts
gryphonmyers: Ok but we will still need a strategy for getting the routes to the client. Currently they're in every contextProps payload. If we say we're dropping support for route functions when using the vue-router integration, we only have to get the routes once, but we still need to get them once. If we're removing the routes member from the contextProps, where would you like to store them? You want to call `loadPageRoutes` in the client? IMO this is unnecessary network overhead. Additionally, `loadPageRoutes` is not actually a comprehensive set of routes - it does not include the implicit ones that do not have a corresponding route file.

brillout: > You want to call loadPageRoutes in the client?

This is already the case: the client loads `route.shared.ts` which calls `loadPageRoutes()`.

>  it does not include the implicit ones that do not have a corresponding route file.

The `route.shared.ts` also knows about all `.page.js` paths.

Also `loadPageRoutes()` works for Route Functions whereas  your serialization approach doesn't. (When using `vite-plugin-ssr` without Vue Router.)

brillout:
(The PR is a *lot* more readable now, looking forward to have a closer look at the `src/` changes. I'll move `src/` to `packages/vite-plugin-ssr/` after the PR is merged.)

src/client/getPage.client.ts
gryphonmyers: > This is already the case: the client loads route.shared.ts which calls loadPageRoutes().

It does? Which client? You mean the built-in client router? Remember, I'm working on an alternative to that code. In the implementations I've done I have no XHR traffic on load, and the only XHR traffic on navigation is for `index.pageProps.json` resources - exactly as it should be, IMO. Any network overhead on load, or additional requests being made on navigation would be wasteful.

brillout: In production these requests are preloaded.

gryphonmyers: It's still another round trip. With Core Web Vitals rollout coming soon, we need to be absolutely meticulous about client performance. We cannot introduce possible bottlenecks. This is a minimal amount of information that could easily be sent to the user as part of another payload.  I see no benefit to loading it as a separate resource in the client.

brillout: One way to think about this is that route files are just like any other source code and are loaded the same way than code loaded with `<script>` tags. (Code that is preloaded with a preload tag and then used with `import.meta.glob` is, from a performance perspective, equivalent to a `<script>` tag.)

I'm all for performance improvements, but tackling the performance problems of `<script>` tags is out of scope for `vite-plugin-ssr`.

Injecting the route strings into HTML (this is basically the only thing you achieve with your approach; the info for filesystem routing is already injected in source code by Vite) solves the `<script>`-tag-delay problem for only 0.01% (probably even less) of source code.

brillout:
I will have a closer look at the PR once we resolved our route info loading disagreement.

src/client/getPage.client.ts
gryphonmyers: Maybe I'm not understanding your solution. I was under the impression that the route files existed outside the context of the bundle and thus were being loaded via AJAX (like the contextProps.json files). If that's not the case and you're telling me these files are are part of the Vite bundle, then I'm not as concerned. Thanks for clarifying.

So just to clarify on the solution you're proposing then:

- In the client Vue Router plugin, load route strings and functions via `loadPageRoutes` 
- For now, throw if we encounter a route function in the loaded routes.
- merge these routes with implicit routes (for pages with no route files). (I'm still not 100% clear on how you're saying we get these in the client. IMO it would be nice if the implicit routes were simply included with the explicit ones.)

brillout: Sounds good.

> merge these routes with implicit routes (for pages with no route files). (I'm still not 100% clear on how you're saying we get these in the client. IMO it would be nice if the implicit routes were simply included with the explicit ones.)

The trick here is that we can determine the filesystem routes by using the list of page ids (the id of a page is the base path of its page files and we can use that to determe its filesystem route), see how `routeWith_filesystem` works.

>  part of the Vite bundle

Yes, check out `dist/server/pageFiles.node.js` to see how it works. You'll also see that the `.page.route.js` files are also in there.

gryphonmyers:
> I will have a closer look at the PR once we resolved our route info loading disagreement.

Might as well wait until I act on your feedback. Won't get to it today

brillout:
Ok

src/client/getPage.client.ts
gryphonmyers: I'm having a problem implementing this solution. Your proposed method for overriding the matching behavior (passing it into the `PageRender` as a customRouting option) involves setting the overrides to the `ssrEnv`. The `ssrEnv` code seems to assume it is executed in a Node environment, but this makes `route.shared.ts` incompatible with browser environments if I reference anything from the `ssrEnv`. This manifests in the form of a `global is not defined` error when trying to call any function that references my `matchRoutes` and `sortRoutes` override functions.

How would you like to solve this? My solution before was to pre-compile the routes on the server and cache them for the client to pick up, but you're pushing back on that so I'm kind of out of ideas here.

gryphonmyers: If you want to take a look at the problem, I've pushed the code in its broken state. You can try checking out my branch, npm installing at root, then starting the vue-router example. BTW I added the vue-router example as a package in the workspace so that it will get all the symlinks from the other packages.

brillout: I wouldn't use `ssrEnv`, but instead implement a custom mechanism that, contrary to `ssrEnv`, is isomorphic.

For example, something along these lines.

```js
// Browser, user-land

// /pages/_default.page.client.js

import { useVueRouter } from '@vite-plugin-ssr/vue-router/client'

useVueRouter()
```

```js
// Browser, `@vite-plugin-ssr/vue-router/client`-land

import { setCustomRouter } from 'vite-plugin-ssr/client/router'

export { useVueRouter }

function useVueRouter() {
  setCustomRouter(/*...*/)
}
```

```js
// Browser, `vite-plugin-ssr/client/router`-land

export { setCustomRouter }

function setCustomRouter(customRouter) {
  window.customRouter = customRouter
}
```

```js
// Isomorphic, `route.shared.ts`-land

function getCustomRouter() {
    if( typeof window !== "undefined") {
      // Browser
      return window.customRouter
    } else {
      // Node.js; `global.customRouter` has been set somewhere during the `createPageRender()` call.
      return global.customRouter
    }
}
```

The Node.js story starting with `createPageRender({customRouter})` would be analogous.

Let me know if something's not clear.

gryphonmyers: OK I'll give this a go

gryphonmyers: This almost works, but I'm getting a strange error when trying to call `loadPageRoutes` on the server (seems to work OK in the client). 

```
Error: [vite-plugin-ssr@0.1.0-beta.34][Internal Failure] You stumbled upon an bug in `vite-plugin-ssr`'s source code (an internal assertion failed). This should definitely not be happening, and you should create a new issue at https://github.com/brillout/vite-plugin-ssr/issues/new that includes this error stack (the error stack is enough to debug this). Or reach out on Discord. A fix will be written promptly.
    at Module.getPageFiles (/REDACTED/code/git-repos/vite-plugin-ssr/src/page-files/getPageFiles.shared.ts:37:9)
    at Module.getPageIds (/REDACTED/code/git-repos/vite-plugin-ssr/src/route.shared.ts:172:43)
    at Module.getRoutes (/REDACTED/code/git-repos/vite-plugin-ssr/packages/vue-router/isomorphic/get-routes.ts:5:68)
    at initRoutes (/@fs/REDACTED/code/git-repos/vite-plugin-ssr/packages/vue-router/server/plugin.ts:22:52)
    at async eval (/REDACTED/code/git-repos/vite-plugin-ssr/packages/vue-router/server/plugin.ts:51:17)
```
Console logging indicates there are some cases where `allPageFiles` is not set when `getPageFiles` is called, I think causing the `assert(hasProp(allPageFiles, '.page'))` assertion to fail. I don't know enough about how this works to troubleshoot why this would ever be the case.

My latest changes are pushed up to this branch if you want to try debugging there.

brillout: Add `@vite-plugin-ssr/vue-router` to Vite's `config.ssr.external`. You can add it directly at `src/plugin/dev.ts` and `src/plugin/build.ts` if you want.

gryphonmyers: OK, doing that sort of worked but it sent me down a rabbit hole of debugging module resolution issues. My current understanding is that these files are now getting resolved here: https://github.com/vitejs/vite/blob/eb66b4350c635fb4f2ef2e8a9eb50958cde73743/packages/vite/src/node/ssr/ssrModuleLoader.ts#L93

However, as you can see, this is a direct call into the `resolve` package, which as indicated here, does not support the node package `exports` field: https://github.com/browserify/resolve/issues/222. More context here: https://github.com/vitejs/vite/issues/1418 it would appear there is SOME mitigation in place in Vite for this shortcoming of resolve, but perhaps something about these packages being marked external is bypassing Vite's handling of the `exports` field? 

All I know is I can get this working if I hack the compiled files of the vue-router package into the root of the package dir (so `resolve` finds the files it expects by virtue of the fact that they literally exist at the imported path). I've pushed it up in this state. It works, kinda, but it sucks. Not sure how you want to work around this.

The good news is that despite this module resolution nonsense, the routing functionality is working!

brillout: For Node.js, we don't need Vite for any of our souce code: `vite-plugin-ssr` only makes user files (i.e. `**/*.page.*` files and their user-land dependencies) be processed by Vite, which is good thing as it simplifies things. (From Vite's perspective the Node.js entry point is `src/page-files/pageFiles.node.ts`.)

Yes, I had problems this as well and I'm not entirely sure how Vite loads dependencies, but what I know is that we don't need something as drastic as `cp -r ./dist/* ./`: instead, we can create some `.js` files at the right positions to do something like `src/client/router/index.js` does. When you think about it: once Vite does `require()`, then we are in Node.js land and Vite is out of the loop and we don't need to hack at this point anymore; we only need to hack the entry points between the user files and our source code.

gryphonmyers: Got it, yeah that's a bit less crappy of a workaround isn't it haha. Went ahead and applied that to the vue-router package. I'm going to mark this thread resolved now

brillout: 🎉 

> yeah that's a bit less crappy of a workaround isn't it haha

:-)

gryphonmyers:
@brillout ready for review

brillout:
Are the tests passing? You can simply rebase/merge master and GitHub should run the tests for you.

gryphonmyers:
Your check is failing on npm install because it's trying to install the not-yet-published `@vite-plugin-ssr/vue-router`. I have been relying on npm workspaces to get this resolving locally, but I see you have some special scripts you're running for installing examples. That kind of seems like unnecessary complication (and it also doesn't work in this case) - why not just add all the examples to the workspace and run a top level npm install?

brillout:
> why not just add all the examples to the workspace and run a top level npm install?

Because, 1. the tests currently run against not only the non-published linked `src/` but also against the published npm `vite-plugin-ssr` pacakge; npm nor Yarn allow to toggle the linkage, and 2. because the user should be able to install a specific example without install all the dependencies of the entire repo.

I will add `/packages/**` to the `npm run link` script, I'll let you know once I've done it.

src/plugin/dev.ts
```diff
@@ -7,7 +7,10 @@ function dev(): Plugin {
     name: 'vite-plugin-ssr:dev',
     apply: 'serve',
     config: () => ({
-      ssr: { external: ['vite-plugin-ssr'] },
+      resolve: {
+        extensions: [ '.js', '.json', '.node', '.ts' ],
```
brillout, line 11: Do we need this?

src/prerender.ts
```diff
@@ -148,6 +148,10 @@ async function prerender({
             return
           }
         }
+        const routeResult = await route(url, allPageIds, contextProps)
+
+        Object.assign(contextProps, { url }, routeResult?.contextPropsAddendum)
+
```
brillout, line 154: I believe we don't need this anymore: there is now `contextProps.urlFull`, `contextProps.urlPathname`, and `cotnextProps.urlParsed`.

src/route.shared.ts
```diff
@@ -263,10 +261,10 @@ async function loadPageRoutes(): Promise<Record<PageId, PageRoute>> {
         `The default export of ${filePath} should be a string or a function.`
       )
       const pageRoute = fileExports.default
-      const pageId = computePageId(filePath)
+      const id = computePageId(filePath)
       const pageRouteFile = filePath
 
-      pageRoutes[pageId] = { pageRoute, pageRouteFile }
+      pageRoutes[id] = { pageRoute, pageRouteFile, id }
```
brillout, line 267: Let's keep it to `pageId`; having consistent naming makes reading source code considerably easier.

src/ssrEnv.node.ts
```diff
@@ -9,13 +10,15 @@ type SsrEnv =
       isProduction: false
       viteDevServer: ViteDevServer
       root: string
-      baseUrl: string
+      baseUrl: string,
+      customRouting?: RoutingHandler
```
brillout, line 14: We don't need this anymore

src/route.shared.ts
```diff
@@ -281,3 +279,112 @@ function isErrorPage(pageId: string): boolean {
...
+  }
+}
+// @TODO eventually the pathToRegexp implementation should be removed from this file entirely so as not to bloat the bundle when it is unused
+function setCustomRouter(customRouter:RoutingHandler = { matchRoutes: matchPathToRegexpRoutes, sortRoutes: defaultSortRoutes }) {
+  if( typeof window !== "undefined") {
+    // Browser
+    return window.__vite_ssr_plugin_custom_router = customRouter
+  } else {
+    // Node.js; `global.customRouter` has been set somewhere during the `createPageRender()` call.
+    return global.__vite_ssr_plugin_custom_router = customRouter
```
brillout, line 388: Let's use a Symbol.

```js
const customRouterKey = Symbol('customRouterKey')
window[customRouterKey]
global[customRouterKey]
```

src/route.shared.ts
```diff
@@ -281,3 +279,112 @@ function isErrorPage(pageId: string): boolean {
...
+function getCustomRouter() {
+  if( typeof window !== "undefined") {
+    // Browser
+    return window.__vite_ssr_plugin_custom_router
+  } else {
+    // Node.js; `global.customRouter` has been set somewhere during the `createPageRender()` call.
+    return global.__vite_ssr_plugin_custom_router
+  }
+}
+// @TODO eventually the pathToRegexp implementation should be removed from this file entirely so as not to bloat the bundle when it is unused
```
brillout, line 381: > so as not to bloat the bundle when it is unused

FYI, I'm afraid this is not possible as things currently are: bundling for the browser is done statically whereas replacing pathToRegexp with a custom router is done at runtime. (But a solution would be to not implicity include pathToRegexp: the user would have to always set `customRouter` (and we would rename `customRouter` to `router`).)

But yes we should still remove it from this file for code readability once the PR is merged.

src/package.json
```diff
@@ -39,6 +39,10 @@
     },
     "./plugin": {
       "require": "./dist/plugin/index.js"
+    },
+    "./route.shared": {
```
brillout, line 43: We can use a new file that controls and makes explicit which epxorts are public.

```ts
// src/route.ts

export { RouteMatch, PageRoute, setCustomRouter, /*...*/ } from './route.shared'
```

brillout:
> I will add /packages/** to the npm run link script, I'll let you know once I've done it.

Actually we can do this instead:
 - Set `examples/vue-router/package.json#dependencies['@vite-plugin-ssr/router']` to `link:../../packages/vue-router/`.
 - In `packages/vue-router/package.json` move `vite-plugin-ssr` to `peerDependencies`.

I believe that should do the trick.

I've added windows to the GitHub actions in `master`.

brillout:
(Btw. all in all the PR looks good, looking forward to merge!)

brillout:
@gryphonmyers Let me know if you are busy and I'll finish up the PR myself. Although note that my comments are all about minor things.

gryphonmyers:
> @gryphonmyers Let me know if you are busy and I'll finish up the PR myself. Although note that my comments are all about minor things.

Won't be at my computer until Monday. I haven't looked at your previous comments yet, but it you've got ideas about changes you'd like to make, feel free!

brillout:
👍 I'll wait :-)

gryphonmyers:
Sorry, pretty slammed with other stuff at the moment. I haven't forgotten about this and will try to get to it ASAP.

brillout:
No rush, take all the time you need :-)

brillout:
@gryphonmyers Any updates on this? Let me know if anything changed on your side, e.g. if you stopped being interested in using Vue Router with `vite-plugin-ssr`.

gryphonmyers:
@brillout nope I just have to deliver a big project with a deadline so have been focusing all my dev time on that. Sorry for the delay, will get back to it when I can.

brillout:
No worries. Do you have an ETA for that big project? I'm just asking to plan my work on vite-plugin-ssr.

gryphonmyers:
> No worries. Do you have an ETA for that big project? I'm just asking to plan my work on vite-plugin-ssr.

About a month. Since we were so close on this I might try to squeeze in some time before then to wrap it up,  maybe on a weekend.

brillout:
No don't worry, I'll do the wrap up.

Thank you for your wonderful work. I'll let you know if I have questions.

brillout:
I didn't manage to run the vue-router example, let's wait until you have more time to finish up the PR.

brillout:
Let me know if you need help merging/rebasing on master. There will be a couple of merge conflicts.

gryphonmyers:
> I didn't manage to run the vue-router example, let's wait until you have more time to finish up the PR.

Ok sorry for the delay - never thought the PR would be open this long

brillout:
No worries :-)

gryphonmyers:
OK I just merged from master and I THINK I resolved all the merge conflicts. Still need to run some code and get the tests working.

gryphonmyers:
@brillout I'm having some issues with the tests - they seem to cause my vue-router package to use an out-of-date (not symlinked) version of vite-plugin-ssr, even though I have it in the package.json as a local path. Any idea what might be causing this? I can confirm an npm install into that directory produces a correct symlink, but running the test command, it somehow gets the bad, non-symlinked copy.

brillout:
Yes that's because `/scripts/link.ts` is not aware of your newly created pacakge.

You can add it to `getDependers` / `scripts/helpers/locations.ts`.

brillout:
Btw. have you thought abour our original conversation? I was arguing that using Vue Router with `vite-plugin-ssr` doesn't make *that much* sense.

While I agree that `vite-plugin-ssr` should be flexible, I only had a handful of users asking for a deep Vue Router integration.

Thoughts?

gryphonmyers:
I mean I've personally seen several people asking how to use it with vue-router and the current answer is you kind of can, but it's complicated to set up and not at all ideal. Nuxt has tight coupling to vue-router (TOO tight if you ask me), but many refugees from that project will expect the integrations it provides. I've already made my points about the advantages of client routing, and your homebrew routing is fine for simple use cases but cannot replace vue-router as it does not integrate with the rendering framework.

gryphonmyers:
I think the direction we have going here is solid: an opt-in package to integrate a popular rendering framework's companion router painlessly with the plugin's structural conventions. Way more versatile than Nuxt and will be just as powerful if we can ever get this pr merged! I do have a few ideas about further improvements but let's save that for later.

src/createPageRender.node.ts
```diff
@@ -5,6 +5,7 @@ import { assert, assertUsage } from './utils/assert'
 import { normalize as pathNormalize } from 'path'
 import { ViteDevServer } from 'vite'
 import { assertBaseUrl } from './baseUrlHandling'
+import { RoutingHandler, setCustomRouter } from './route.shared';
```
brillout, line 8: The import `RoutingHandler` is not used; we can remove it.

PR #44, opened 2021-04-19
Set `as` properly on preload tags.

chrisvariety:
Per https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content

Using `as` to specify the type of content to be preloaded allows the browser to:

- Prioritize resource loading more accurately.
- Store in the cache for future requests, reusing the resource if appropriate.
- Apply the correct content security policy to the resource.
- Set the correct Accept request headers for it.

It should also resolve this error in Safari (at least for the stylesheet, not so much the `svg` - open to ideas there):

Screenshot 2021-04-19 at 18 07 13@2x

### DIFF
--- a/src/getPreloadTags.node.ts
+++ b/src/getPreloadTags.node.ts
@@ -93,10 +93,10 @@ function getPreloadTag(href: string): string {
   assert(href.startsWith('/'))
   assert(!href.startsWith('//'))
   if (href.endsWith('.css')) {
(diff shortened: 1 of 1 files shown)


brillout:
👍 As for other `as` values we'd have to guess them from file extension. Ideally a library would do this, but we could also do it ourselves for the most common file extensions such as `.svg`.

PR #61, opened 2021-05-17
Update boilerplates for lighthouse score optimization

yoroshikun:
Score Locally is now perfect 100s

image


brillout:
Neat new commit :-).

One last little thing: maybe we can also use a class `navitem` for the React boilerplates and define it in `PageLayout.css`?

brillout:
Feel free to add commits instead of push force (so we can see the test go green). Whatever works best for you.

brillout:
Wonderful ❤️ .

Welcome to the contributor list.

brillout:
I reduced the padding to `3px` which is enough to make Lighthouse 100% happy.

brillout:
(Even with Lighthouse for mobile.)

PR #99, opened 2021-06-14
fix: avoid duplicate error logs

aleclarson:
Vite already logs errors that occur during `ssrLoadModule` calls:
https://github.com/vitejs/vite/blob/main/packages/vite/src/node/ssr/ssrModuleLoader.ts#L145-L152

So this plugin should avoid logging them again.


src/page-files/setup.node.ts
```diff
@@ -26,9 +26,9 @@ async function setPageFiles(): Promise<unknown> {
     let moduleExports: any
     try {
       moduleExports = await ssrEnv.viteDevServer.ssrLoadModule(modulePath)
-    } catch (err) {
-      ssrEnv.viteDevServer.ssrFixStacktrace(err)
-      throw err
+    } catch {
+      // Vite logs the error, so just return empty array.
+      return []
```
brillout, line 31: How about

```js
err._alreadyLoggedByVite = true
throw err
```

src/renderPage.node.ts
```diff
@@ -684,7 +684,10 @@ async function render500Page(
   | { nothingRendered: true; renderResult: undefined; statusCode: undefined }
   | { nothingRendered: false; renderResult: string | unknown; statusCode: 500 }
 > {
-  handleErr(err)
+  // When `ssrLoadModule` fails, Vite will log the error itself.
+  if (!(err instanceof Error) || !err.stack?.includes('instantiateModule')) {
```
brillout, line 688: Then here `if( err?._alreadyLoggedByVite )`

brillout:
Hi Alec, that makes sense & thanks for the info

src/page-files/setup.node.ts
aleclarson: Sounds good. We'll need a try-catch around this call too then:
https://github.com/brillout/vite-plugin-ssr/blob/16d6b56bf2c1860c62c36888bb769f1cc57a2ec2/src/renderPage.node.ts#L361

brillout:
Good catch... 👌 

Two things:
 - If I'm not mistaken, Vite only logs in dev. In production it doesn't. (Check out `dist/server/pageFiles.node.js` after `yarn create vite-plugin-ssr && cd vite-ssr-project && yarn && yarn build`.)
 - We should wrap `loadFile` at the source (`allPageFiles[fileType]` are results of `import.meta.glob()` calls).
   https://github.com/brillout/vite-plugin-ssr/blob/16d6b56bf2c1860c62c36888bb769f1cc57a2ec2/src/page-files/getPageFiles.shared.ts#L49-L51

The whole function:

https://github.com/brillout/vite-plugin-ssr/blob/16d6b56bf2c1860c62c36888bb769f1cc57a2ec2/src/page-files/getPageFiles.shared.ts#L35-L54

It is always `getPageFiles()` that provides the page files.

brillout:
Actually, the more I think about it the more I consider it a problem on Vite's side.

It's just unnatural and not expected that Vite logs server-side errors. This is not idiomatic.

On the other hand, if we want to have beautiful dev error handling, then it makes sense that Vite also takes care of SSR errors.

I'm thinking maybe we should actually bake this `_alreadyLoggedByVite` flag into Vite itself.

As we can see in this very PR, it's error prone and not always easy to decide when `vite-plugin-ssr` should handle the error or not. Such flag is AFAICT a one-liner in Vite's src while making plugin authors' life much easier.

Thoughts?

aleclarson:
> I'm thinking maybe we should actually bake this `_alreadyLoggedByVite` flag into Vite itself.

https://github.com/vitejs/vite/pull/3957

brillout:
Neat cross-upstream PR 👌

src/renderPage.node.ts
```diff
@@ -677,7 +677,7 @@ function isFileRequest(urlNormalized: string) {
 }
 
 async function render500Page(
-  err: unknown,
+  err: Error,
```
brillout, line 680: It actually is `unknown` because the user may have `throw 'some-error'` instead of `throw new Error('some-error')`

brillout:
I'll merge as soon as https://github.com/vitejs/vite/pull/3957 is released.

I'll then update the peer dep and release.

PR #114, opened 2021-07-10
feat: Support specifying custom `outDir`

zaxbux:
This gives the developer the option to specify a custom "outDir" (like `build.outDir` in `vite.config.*`) rather than relying on the hard-coded 'dist'.


brillout:
What is your use case for wanting to change the output directory?

brillout:
Closing this, let me know if and why you need to change the output dirs.

dymart:
I would like to use bazel with vite-plugin-ssr.

Great glad to hear you are open to a PR on this!

PR #115, opened 2021-07-11
Fix getting scroll position from history

shwao:
Navigating back with the client router works only once or not all. You can try that in the **react-full** example. Run `npm run dev`, go to `http://localhost:3000`, click on a link in the menu, click the browser back button and watch the console. Either this already failed or if not go forward again. This error message appears:

```
Uncaught TypeError: can't access property "scrollPosition", obj is undefined
    hasProp hasProp.ts:22
    getScrollPositionFromHistory useClientRouter.client.ts:243
    onBrowserHistoryNavigation useClientRouter.client.ts:199
```

The problem seems to be that you set the history state to undefined in the changeUrl function:

https://github.com/brillout/vite-plugin-ssr/blob/191b073ea2c4232dfa2d7626a2f1768cd13ef3a9/src/client/router/useClientRouter.client.ts#L208

And try to get the scroll position from this now undefined object here:

https://github.com/brillout/vite-plugin-ssr/blob/191b073ea2c4232dfa2d7626a2f1768cd13ef3a9/src/client/router/useClientRouter.client.ts#L243

I just added a check if historyState exists to this line:

https://github.com/shwao/vite-plugin-ssr/blob/master/src/client/router/useClientRouter.client.ts#L243

### DIFF
--- a/src/client/router/useClientRouter.client.ts
+++ b/src/client/router/useClientRouter.client.ts
@@ -240,7 +240,7 @@ function setScrollPosition(scrollTarget: ScrollTarget): void {
 }
 
 function getScrollPositionFromHistory(historyState: unknown = window.history.state) {
-  return hasProp(historyState, 'scrollPosition') ? (historyState.scrollPosition as ScrollPosition) : null
+  return historyState && hasProp(historyState, 'scrollPosition') ? (historyState.scrollPosition as ScrollPosition) : null
 }
 
 function autoSaveScrollPosition() {


brillout:
I don't think your PR makes a difference, see:

https://github.com/brillout/vite-plugin-ssr/blob/191b073ea2c4232dfa2d7626a2f1768cd13ef3a9/src/utils/hasProp.ts#L20-L33

So `hasProp()` will return `false` if `obj` is `undefined`.

But let me try to reproduce your bug.

brillout:
I can reproduce, and, actually you are right the next line `const propValue = (obj as Record<any,unknown>)[prop]` will fail. Let me dig further.

shwao:
Exactly, i thought this was intentional in `hasProps`.

brillout:
I found the root cause, I'm writing a fix which I'll publish in a moment.

brillout:
Fix released in `v0.1.5`.

shwao:
Yes it works and you also fixed the / route problem! Thank you!

brillout:
Sorry about the bug; it was a regression in the latest minor release. Was an edge case that slipped through the cracks that the test suite didn't catch.

Let me know if there is anything else.

PR #132, opened 2021-08-14
Add SolidJS example

wobsoriano:
This adds a minimal example for SolidJS with client-side routing


brillout:
👌 

cc @ryansolid FYI

ryansolid:
Looks good. Only semi-unfortunate thing is because Solid has no VDOM a sort of redraw everything from the top of the page approach prevents shared layouts but short of having a nested router(like React Router 6) there isn't much for that. Not sure if there is any possibility of bringing Solid's router in specifically for Solid. It might complicate things since the way Solid's hydration matches up on components the same router would need to be on both sides.

brillout:
@ryansolid Ok

PR #165, opened 2021-10-07
Prefetching route files

Axxxx0n:
This is rough implementation for feature discussion

- exposed prefetchUrl method for manually prefetching route files
- added two strategies for prefetching <a> links: `inViewport` (similar to nextjs) which prefetches when links are in viewport and `onHover` (similar to svelteKit)  which does the same on link hover
- automatic prefetching is only done in production


brillout:
> You want this option for overriding `prefetchLinks` option, or something else?

My idea is this: imagine there is a link that has high probability of being clicked. E.g. a checkout process: the "next" button should be aggressively prefetched. So the global option could be `prefetchLinks: 'onHover'` while `<a href="/checkout/step/2" data-prefetch="inViewport">Next</a>`.

So yea basically overriding, or let me know if you have something else in mind.

brillout:
Maybe we can add the code in a new file `prefetch.ts`.

brillout:
> I was testing how much time will prefetch have when user clicks link on mobile before actual navigation. After consideration I will revert back as there is ~200ms time in which we can prefetch without visual impact on navigation. This is more than enough on normal network speed.

What happens if there are several concurrent `loadPageFiles()` calls for the same `_pageId`? I'm not sure what Vite nor the browser does.

Axxxx0n:
> Btw any objections for also calling getPageContext() on hover/touchstart?

Sorry, I'm not sure what you mean by this

brillout:
So on a `prefetch()` call we currently only call `loadPageFiles()`. We could also call [`getPageContext()`](https://github.com/brillout/vite-plugin-ssr/blob/master/vite-plugin-ssr/client/router/getPageContext.ts) which basically fetches `pageContext` from the server.

Axxxx0n:
Originally was like this, but I had couple of problems:
- we need to fetch page files and context in parallel because we don't want to block fetching static files with fetching api endpoints
- when we hover on link and prefetch page files and context, we could actually click this link at the later time (for example after 5s or 10s). Then we have stale page context data.

Axxxx0n:
> But yea let's focus on loadPageFiles() for now.

Agree. :)

> imagine instant page navigation for stateful pages

Isn't that going to be covered in: https://github.com/brillout/vite-plugin-ssr/issues/95? When you have `onBeforeRoute` in `.page.js`, navigation would be instant.

brillout:
No more conflict, nice. Let me have a final look at your changes.

brillout:
💯

brillout:
>   //TODO enable only in production: if(import.meta.env.DEV) return

I removed that line because it's actually a good thing to have dev to behave similarly to prod. Let me know if you see a problem about that.

brillout:
Docs: https://github.com/brillout/vite-plugin-ssr/commit/9d93b46d5d7c5e22901c97f2d81b6c5b8255477a

Axxxx0n:
> > //TODO enable only in production: if(import.meta.env.DEV) return
> 
> I removed that line because it's actually a good thing to have dev to behave similarly to prod. Let me know if you see a problem about that.

We need to check what happens when you preload page bundle, than change that page code, then navigate to that page. I am not sure if bundle would then be stale or refetched on navigation.

PR #189, opened 2021-10-29
adding miniflare and tests to cloudflare workers examples

rslonik:


rslonik:
I went to check what @lukeed was using (he always build a lot of cool stuff) and he wrote one test runner 😅 https://github.com/lukeed/uvu

brillout:
Yea I know about `uvu` but it still doesn't quite nail it. But yea I'll re-consider uvu, let's see if it changed since the last time I checked it out.

brillout:
Got it working, but `wrangler` doesn't work in GitHub actions because:

> Please run `wrangler login`, `wrangler config`

Is there a way to use `wrangler` in GitHub actions?

rslonik:
Yes, I have a project that uses this to upload static generated site to workers. The step in my action is this:

```
- uses: cloudflare/wrangler-action@1.3.0
      with:
        apiToken: ${{ secrets.CF_API_TOKEN }}
        environment: 'development'
        wranglerVersion: '1.18.0'
```

brillout:
And `${{ secrets.CF_API_TOKEN }}` can be hidden even for public GitHub repos?

rslonik:
Secrets are set in repo's settings > secrets. I think it's safe.

https://docs.github.com/en/actions/security-guides/encrypted-secrets

brillout:
Ah, I actually just stumbled about this page :-) Thanks

brillout:
Both `miniflare` and `wrangler` are now included in the test suite.

PR #198, opened 2021-11-09
Inject link tags at the end of head or custom point

sopium:
Currently vite-plugin-ssr injects link tags (stylesheets, preload links, etc.) at the very start of head, before any application defined head tags.

It makes it rather difficult to use `<meta http-equiv="X-UA-Compatible" content="IE=Edge">`, which must be the first.

Also it is a bit more robust, idiomatic and performant to put `<meta charset>`, `<title>` and `<meta name="viewport">` (and other critical meta tags or resources) near the start, so that charset detection, title and layout are not blocked or delayed.

This PR changes link tags to be injected at the end of head by default, or at the verbatim string `<!-- vite-plugin-ssr-link-tags -->` (which is replaced).

### DIFF
--- a/examples/react-full/.testPages.ts
+++ b/examples/react-full/.testPages.ts
@@ -20,6 +20,11 @@ function testPages(
     expect(html).toContain("<h1>Welcome to <code>vite-plugin-ssr</code></h1>");
   });
 
+  test("link tags are correctly injected", async () => {
+    const html = await fetchHtml("/");
+    expect(html).toMatch(/<head>.*<title>.*<link.*<\/head>/s);
+  });
+
   test("page is rendered to the DOM and interactive", async () => {
     await page.goto(urlBase + "/");
     expect(await page.textContent("h1")).toBe("Welcome to vite-plugin-ssr");
--- a/vite-plugin-ssr/node/html/injectAssets.ts
+++ b/vite-plugin-ssr/node/html/injectAssets.ts
@@ -232,10 +232,11 @@ function injectScript(htmlString: string, script: PageAsset): string {
   return injectEnd(htmlString, injection)
 }
 
+const headClose = '</head>'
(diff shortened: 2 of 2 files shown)


brillout:
> `<meta http-equiv="X-UA-Compatible" content="IE=Edge">`

Is this still used? Isn't that for older IE versions that are nowadays deprecated anyways?

> This PR changes link tags to be injected at the end of head by default

👍 

>  or at the verbatim string  (which is replaced).

I don't think we need this option? Let's try to keep the number of options low.

brillout:
You can rebase on master to make the test go green.

sopium:
> Is this still used? Isn't that for older IE versions that are nowadays deprecated anyways?

Well it's mostly to make the point that it's safer to inject at the end than at the start. To be honest if I need to support IE I probably wouldn't choose a vite based stack.

> I don't think we need this option? Let's try to keep the number of options low.

OK, removed.

examples/react-full/.testPages.ts
```diff
@@ -20,6 +20,11 @@ function testPages(
     expect(html).toContain("<h1>Welcome to <code>vite-plugin-ssr</code></h1>");
   });
 
+  test("link tags are correctly injected", async () => {
+    const html = await fetchHtml("/");
+    expect(html).toMatch(/<head>.*<title>.*<link.*<\/head>/s);
```
brillout, line 25: Let's use `partRegex`, it's slightly more readable.

sopium: `partRegex` does not seem to support flags? (The `/s` flag is necessary to match newlines.)

brillout: You're right

brillout:
Neat PR 👌 

I'll take care of the flaky tests.

brillout:
Let me know if you need this sooner rather than later, then I'll release today.

Issue #151, opened 2021-09-27
detect client-server version mismatch


brillout:
> We add a pageContext._vitePluginSsrVersion
> The client then checks if its version matches

brillout:
Closing for lack of interest.

PR #199, opened 2021-11-09
feat: add disableClientRouter and restoreClientRouter

sopium:
After calling `disableClientRouter`, clicking on links or calling navigate will use server routing, even if you have called `useClientRouter`.

This is useful when e.g. you have detected that a new version of your application have been deployed, and you want the next page load to start fresh (related #151).

And `restoreClientRouter` to the reverse effect.

### DIFF
--- a/vite-plugin-ssr/client/router/index.ts
+++ b/vite-plugin-ssr/client/router/index.ts
@@ -1,4 +1,4 @@
 import '../page-files/setup'
-export { useClientRouter, navigate } from './useClientRouter'
+export { useClientRouter, navigate, disableClientRouter, restoreClientRouter } from './useClientRouter'
 export { prefetch } from './prefetch'
 export type { PageContextBuiltInClient } from './types'
--- a/vite-plugin-ssr/client/router/useClientRouter.ts
+++ b/vite-plugin-ssr/client/router/useClientRouter.ts
@@ -19,8 +19,35 @@ import { skipLink } from './utils/skipLink'
 
 export { useClientRouter }
 export { navigate }
+export { disableClientRouter }
+export { restoreClientRouter }
 
 setupNativeScrollRestoration()
+
+let clientSideNavigationDisabled = false
+
+/**
+ * Disable client router.
+ *
+ * After calling this, clicking on links or calling navigate will use server
+ * routing, even if you have called useClientRouter.
+ *
+ * This is useful when e.g. you have detected that a new version of your
+ * application have been deployed, and you want the next page load to start
+ * fresh.
+ */
+function disableClientRouter() {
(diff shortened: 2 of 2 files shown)


brillout:
👍 

It's actually something I wanted to work on https://github.com/brillout/vite-plugin-ssr/issues/151.

AFAICT we could even make `vite-plugin-ssr` do this automatically, without the user having to implement its own version mismatch detection.

Should be possible by generating a build UUID upon `$ vite build`. It essentially ensures that the client-side code and the server-side code were generated by the same build run. 

Thoughts?

brillout:
You can rebase on master to make the test go green.

sopium:
I wasn't sure about the versioning scheme to use, so I haven't done that part. I'm using `hash(readFile("dist/client/manifest.json"))` for my app.

I was thinking that after you implement the automatic detection, `disableClientRouter` would be one of options to handle the mismatch.

brillout:
> `disableClientRouter` would be one of options to handle the mismatch.

I don't think this will be needed.

> I was thinking that after you implement the automatic detection,

Up for trying to implement #151? We can add a new field `buildId` to `dist/client/vite-plugin-ssr.json` (generated by https://github.com/brillout/vite-plugin-ssr/blob/master/vite-plugin-ssr/node/plugin/manifest.ts)

brillout:
> Up for trying to implement #151? We can add a new field `buildId` to `dist/client/vite-plugin-ssr.json` (generated by https://github.com/brillout/vite-plugin-ssr/blob/master/vite-plugin-ssr/node/plugin/manifest.ts)

Actually, even simpler would be to use a new Vite transformer to inject the `buildId` to a new file `shared/buildId.ts`.

Both client-side code and server-side code then both import that file to have the `buildId` at their disposal.

We then set `pageContext._buildId` and always pass it to the client. The client then compares both numbers.

brillout:
(You can rebase on master again for a more robust test suite.)

sopium:
> > Up for trying to implement #151? We can add a new field `buildId` to `dist/client/vite-plugin-ssr.json` (generated by https://github.com/brillout/vite-plugin-ssr/blob/master/vite-plugin-ssr/node/plugin/manifest.ts)
> 
> Actually, even simpler would be to use a new Vite transformer to inject the `buildId` to a new file `shared/buildId.ts`.
> 
> Both client-side code and server-side code then both import that file to have the `buildId` at their disposal.
> 
> We then set `pageContext._buildId` and always pass it to the client. The client then compares both numbers.

I'm not very familiar with vite internals. Free free to close this and implement it your way.

brillout:
I won't implement this anytime soon though as I'm working on higher-prio things.

Issue #196, opened 2021-11-03
<link rel=preload> must have a valid as value (chrome and firefox) when use less styles


txus2202:
i installed less in my project but i see an error in console of generated preload links by my less styled components.

error: Preload of http://localhost:3000/pages/index/index.page.vue?vue&type=style&index=0&scoped=true&lang.less was ignored due to unknown “as” or “type” values, or non-matching “media” attribute.

i think is necessary add as="style"

How can I introduce this "as" somehow in my preload links?

brillout:
What exactly is the preload tag? Do yo get this in prod or dev?

txus2202:
in pro and dev.
In my components i use this style label to add the styles:
<style scoped lang="less">
...
</style>

brillout:
Seems like a problem with Vue, not with vite-plugin-ssr.

But give me a repro and I'll have a look.

txus2202:
unfortunately I can't provide you with the repository, but to reproduce it, you just need to add it to the package "less": "^4.1.2" as a dependency, install and use <style scoped lang = "less"> ... </style> in any component

brillout:
Do it yourself and publish the repo and I'll have a look.

Closing in the meantime.

brillout:
Fix released in `0.3.28`.

PR #206, opened 2021-11-15
Add scss/sass/less to inferred types

hellboywar:
When using scss/sass/less fixes console warning (#196)

Снимок экрана 2021-11-15 в 14 58 47

and makes SSR css to load with html in dev mode which removes page flickering and layout shifting after client css is loaded


brillout:
I will release in a moment.

brillout:
Released in `0.3.28`.

PR #210, opened 2021-11-19
fix: fix client routing deadlock when ensureHydration is set

yxonic:
The problems I've fixed in this pull request:

1. When `ensureHydration` is set, the first call of `fetchAndRender` should never abort. Otherwise, follow-up calls will all stuck at `await hydrationPromise`, if user clicks before the first call finishes.
2. Transition should start as soon as user clicks. If transition starts after `await hydrationPromise`, users with slow network will have to wait for initial hydration without any visual indication.
3. I also fixed a typo here.

### DIFF
--- a/vite-plugin-ssr/client/router/useClientRouter.ts
+++ b/vite-plugin-ssr/client/router/useClientRouter.ts
@@ -79,12 +79,7 @@ function useClientRouter({
     const callNumber = ++callCount
     assert(callNumber >= 1)
 
-    if (ensureHydration) {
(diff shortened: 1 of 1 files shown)


brillout:
Makes sense.

brillout:
FYI https://github.com/brillout/vite-plugin-ssr/commit/2e6f74565b8f7a797966b11193f3cef96819cdfb

brillout:
Released in `0.3.31`

PR #218, opened 2021-11-28
Update Vercel deployment instructions & example

patryk-smc:
Few notes:
- I renamed `api` folder to `vercel`, because Vercel treats `api` folder specially -[it is meant for simple Serverless Functions](https://vercel.com/docs/concepts/functions/serverless-functions#deploying-serverless-functions), and now we are using [File System API](https://vercel.com/docs/file-system-api) to deploy the entire app. 

- sh script requires setting chmod to work. The one in repo has correct chmod, but if someone copies the code from repo it will throw an error. 

- I tested sh script on Mac, I'm not certain if it will work on Windows or Unix

- New File System API is in beta, so settings env variable is required to work


examples/vercel/vercel.json
```diff
@@ -1,8 +1,9 @@
...
+  "redirects": [
     {
-      "source": "/((?!assets/).*)",
-      "destination": "/api/render.js"
+      "source": "/view-source",
+      "destination": "https://github.com/vercel/vercel",
+      "permanent": false
     }
   ]
 }
```
brillout, line 9: I'd remove `vercel.json` since the integration doesn't need it anymore.

docs/pages/vercel.page.mdx
```diff
@@ -2,24 +2,20 @@ import { RepoLink } from 'libframe-docs/components/RepoLink'
...
+ 3. In Vercel's web UI, set the env variable `ENABLE_FILE_SYSTEM_API` of your Vercel app to `1`.
 
 Example:
  - <RepoLink path='/examples/vercel/' />
 
 More infos:
  - Insightful conversation: [github.com/vercel/vercel/discussions/6393](https://github.com/vercel/vercel/discussions/6393#discussioncomment-1008607).
  - `vercel.json`'s config `includeFiles` doesn't work for files that are not committed to your Git repository: [github.com/vercel/vercel/issues/6478](https://github.com/vercel/vercel/issues/6478).
-
+ - You can use `vercel.json` to handle redirects. Example: <RepoLink path='/examples/vercel/vercel.json' />
```
brillout, line 21: Not sure if should add this note. I mean how Vercel supports redirects has not much to do with vite-plugin-ssr.

docs/pages/vercel.page.mdx
```diff
@@ -2,24 +2,20 @@ import { RepoLink } from 'libframe-docs/components/RepoLink'
 
 To deploy a `vite-plugin-ssr` app to Vercel:
  1. Load `dist/server/importBuild.js` in your Vercel serverless function.
-    Example: <RepoLink path='/examples/vercel/api/render.js' />.
+    Example: <RepoLink path='/examples/vercel/vercel/render.js' />.
     > Behind the scenes, Vercel bundles your serverless function and therefore needs to know its entire dependency tree;
     > [`dist/server/importBuild.js` enables](/importBuild.js) Vercel to discover the entire dependency tree of your serverless function.
 
- 2. Define your build script in `package.json#scripts['vercel-build']` instead of `package.json#scripts.build`.
+ 2. Define deploy.sh script in `package.json#scripts['vercel-build']` along with `package.json#scripts.build`.
```
brillout, line 9:
```diff
- deploy.sh
+ `deploy.sh`
```

examples/vercel/vercel/render.js
```diff
@@ -1,18 +1,25 @@
 import { createPageRenderer } from 'vite-plugin-ssr'
-// `importBuild.js` enables Vercel to bundle our serverless functions, see https://vite-plugin-ssr.com/vercel and https://vite-plugin-ssr.com/importBuild.js
```
brillout, line 2: Any reason for removing this?

brillout (changes requested):
👍 Awesome.

Does Vercel has a local preview thing? So that we can add some tests for the Vercel integration. Would be good especially since we are using a beta API.

examples/vercel/vercel/render.js
patryk-smc: Nope, I just missed it. Added.

patryk-smc:
> 👍 Awesome.
> 
> Does Vercel has a local preview thing? So that we can add some tests for the Vercel integration. Would be good especially since we are using a beta API.

They have this: https://vercel.com/docs/cli#commands/dev - but I'm not confident if that would be useful for testing. Vercel has an option to select root of the app, so by pointing it to the example folder, it would build on each PR and prompt each time if it succeeded or not. Not sure if that works for you, though.

That Vercel CLI gave me an idea to ditch express completely from this example and just relay on Vercel CLI. It works just fine. What do you think?

brillout:
> That Vercel CLI gave me an idea to ditch express completely from this example and just relay on Vercel CLI. It works just fine. What do you think?

Do you mean using Vercel's dev server?

If the app doesn't depend on Vercel features to work, then I'd argue it's a good idea to dev with Express and only use Vecel for deploy.

But we can have both, the Express.js dev server as well as the Vercel dev server.

```json
{
  "scripts": {
    "dev:vercel": "...",
    "// For increased dev speed we can use an Express.js dev server": "",
    "dev:express": "node ./dev-server"
  }
}
```

The user can then remove the dev server she's not interested in.

> They have this: https://vercel.com/docs/cli#commands/dev - but I'm not confident if that would be useful for testing. Vercel has an option to select root of the app, so by pointing it to the example folder, it would build on each PR and prompt each time if it succeeded or not. Not sure if that works for you, though.

A prompt is no good for CI - there is no option to remove the prompt?

patryk-smc:
> If the app doesn't depend on Vercel features to work, then I'd argue it's a good idea to dev with Express and only use Vecel for deploy.
> 
> But we can have both, the Express.js dev server as well as the Vercel dev server.

Agree. For now, let's keep just express. We will use Vercel CLI for some time, and if it works fine, I'll make another PR.

 



> A prompt is no good for CI - there is no option to remove the prompt?

There are options such as: https://vercel.com/docs/cli#commands/overview/unique-options/confirm and https://vercel.com/docs/cli#options/global-options/token

brillout:
👍 Should work then, e.g. that's what we already do for Cloudflare https://github.com/brillout/vite-plugin-ssr/blob/master/.github/workflows/ci.yml#L37-L38

Issue #176, opened 2021-10-19
[TypeScript] Type check accross files


brillout:
Currently:

```ts
// server/index.js

app.get("*", async (req) => {
  const pageContextInit = {
    url: req.url,
    someInitProp: 'hello :-)'
  }
  const pageContext = await renderPage(pageContextInit)
  // TS complains that `someNewProp` is not defined on `pageContext`
  console.log(pageContext.someNewProp) // prints `42`
});
```

```ts
// some.page.server.js

import type { PageContextBuiltIn } from "vite-plugin-ssr";

epxort async function onBeforeRender(pageContext: PageContextBuiltIn) {
  // TS complains that `someInitProp` is not defined on `pageContext`
  console.log(pageContext.someInitProp) // prints `hello :-)`
  return {
    pageContext: {
      someNewProp: 42
    },
  };
}
```

How can we improve the situation?

AFAICT, something like [Next.js' utility types](https://nextjs.org/docs/basic-features/typescript#static-generation-and-server-side-rendering) don't add any value compared to `vite-plugin-ssr`'s [`PageContextBuiltIn`](https://vite-plugin-ssr.com/pageContext#typescript).

Contribution welcome.

CC @stouffi

magne4000:
[redux-toolkit](https://redux-toolkit.js.org/tutorials/typescript) has some really nice patterns regarding Typescript usage, it kinda look like [this solution you proposed](#issuecomment-950314783).

I would even go for something like this:
```ts
export  { renderPage, onBeforeRender };
const { renderPage, onBeforeRender } = createDefine<PageContextInit, PageContextAfterOnBeforeRender>();
```

I also like your [first proposed solution](#issuecomment-950296223), as it makes all of this less error prone I guess.

brillout:
@magne4000 What would be neat is a little TS playground to experiment our ideas. Something simple with only `renderPage()` and `onBeforeRender()`; that should be enough to showcase cross-file type checking.
 
> ```ts
> export { renderPage, onBeforeRender };
> const { renderPage, onBeforeRender } = createDefine<PageContextInit, PageContextAfterOnBeforeRender>();
> ```

What exactly do you mean? Note that we cannot use a `createHooks()` function to define all hooks at once. Because one hook may be defined in `.page.server.js` (i.e. only exectued in Node.js) while another in `.page.js` or `.page.route.js` (i.e. also executed in the browser).

magne4000:
[...]

PR #219, opened 2021-11-28
Improve types for createPageRenderer and onBeforeRender

magne4000:
Following the proposals on #176, here is a base to continue the discussion on how to improve current typings.

### What's in there?
- Type proposal for `createPageRenderer`
- Type utils for `PageContext` and `OnBeforeRender`

See comments below for details.


vite-plugin-ssr/node/renderPage.ts
```diff
@@ -72,12 +72,18 @@ export { throwPrerenderError }
...
+    pageContextInit: PageContextInit,
+  ): Promise<
+    PageContextInit &
+      (({ httpResponse: HttpResponse } & PageContextAdded) | ({ httpResponse: null } & Partial<PageContextAdded>))
+  >
+}
+
+const renderPage: RenderPage<{}> = async function renderPage (
+  pageContextInit,
+) {
```
magne4000, line 86: `PageContextAdded` needs to be specified, but `PageContextInit` can always be infered. I tested it that way:

```ts
const renderPage = createPageRenderer<RenderPageCustomFields>({
  isProduction: true,
});

...
const pageContextInit = { url: req.url, body: req.body, cookies: req.cookies };
const pageContext = await renderPage(pageContextInit);
// pageContext typing contains Partial<RenderPageCustomFields> and type infered from pageContextInit
```

vite-plugin-ssr/node/types.ts
```diff
@@ -11,3 +11,9 @@ export type PageContextBuiltIn = {
     hash: null | string
   }
 }
+
+export interface DefinePageContext<PageContext extends {}, U extends {}> {
+  PageContext: PageContextBuiltIn & PageContext;
+  OnBeforeHook: (arg: PageContextBuiltIn & PageContext) =>
+    { pageContext?: { pageProps?: Record<string, unknown> } } & U;
+}
```
magne4000, line 19: usage:
```ts
type PageContextStates = DefinePageContext<PageContextHere, RenderPageCustomFieldsRequired>;

type OnBeforeHook = PageContextStates['OnBeforeHook'];
type PageContext = PageContextStates['PageContext'];

export { OnBeforeHook, PageContext };
```

magne4000:
Also, how and where should this appear in the documentation/boilerplates/examples?

vite-plugin-ssr/node/renderPage.ts
brillout: 👍 That's a neat idea and neat trick with `interface`.

brillout:
Just created https://vite-plugin-ssr.com/typescript

> here is a base to continue the discussion on how to improve current typings.

How about we set this PR as `draft`?

brillout:
I'm wondering, regardless of how to achieve it with TypeScript, how far can we push type safety in the context of vite-plugin-ssr?

magne4000:
Pretty far I would say if we follow the right patterns. With the current state + this PR, I think we are already pretty good.
After that it's just a matter of improving exiting typing to represent more precisely what is going on (like find a better solution than using `Partial<PageContextAdded>` for `RenderPage` typing).

The goal of this plugin is to have a pretty good TS definition (current state), or consider TS as a feature for this lib? The second option probably require specific pattern like the ones used by redux-toolkit (dummy fumction overrides with typings and such). TS as a feature means absolute perfect TS definitions, and it could become hard to maintain (look at the type definitions of [rxjs operators](https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/groupBy.ts), this can become time consuming).
However, the public API of this plugin is quite small compared to the libraries I mentioned, so I think it could be manageable to ensure really precise typings for those.

NB: I also tried to leverage the `declare module *.page.server.tsx ...` syntax but without success, it doesn't enforce type for the file itself, only for the files that import such files.

brillout:
I think we can consider it a built-in feature. The extra type safety is opt-in; if it breaks that's ok.

I think we'd cover already 90% if we:
 1. Automatically provide `PageContextInit` to all hooks.
 2. Enable different `pageContext.pageProps` per-page. So basically typesafty for these files:
    ```js
    // pages/product/index.page.route.ts

    export default '/product/:productId'
    ```
    ```js
    // pages/product/index.page.ts

    // How can we connect the `pageProps` here with the
    // `pageProps` of the `onBeforeRender()` hook below?
    export function Page(pageProps) {
     // ...
    }
    ```
    ```js
    // pages/product/index.page.server.ts

    export async function onBeforeRender(pageContext) {
      const { productId } = pageContext.routeParams
      const productData = await retrieveProductData(productId)
      return {
         pageContext: {
           pageProps: {
             productData
           }
         }
      }
    }
    ```

magne4000:
I think I managed to find an elegant solution for typing `onBeforeRender`, `Page` and `PageProps` (and possibly some more without too much difficulties), that requires overriding only what's necessary in a declared namespace: [TSPlayground](https://www.typescriptlang.org/play?#code/PQKhCgAIUhRAPAlgZwC6IHYHNKoJ4AOApslCMOEfAQPYBOquhRkACgIZZEDCNGqVVACEArogA2qAJIZIAXkgBvKG05EAXJHYY8KgmoS0GyTQCUiAY3oATADxo6mLABpIIjAGsMNAO4YAfCp0NCICHHTsALYmkOZWdHYOTq5J2IGQbnTimqlYKiJZHKgAFhhRGpC5+VkAcvSR7OKIAF5E1jmojtjV4uHIbZrKGRn6JWWRFVXDlUTsdBbFmhgi4uKQAD6xljb2ncmVe2kqGcXsyIuQy6sbB115GQC+4E-goBDQkDXlyPoWLCXsRg+FgWbSQfQMRCNcR4SA0ABuRDojmsRDIFFRFnEcxY4xIvxYADVEGFxCIsJgAMrIOhKFTAYCVYohcTWSAAIxYCKRKKIGBUVCMjEwAjoADN2H9INwRGgaJEOFxePxBEoXhkGaolXwBPBGKh2B4nNLZah5YqeDrVSKaFoLFZ3KhXCUUJAfOw8K4fMU+XDEcjENZjTK5Qq1Mrdah6YyXcgmMQ3RI1o1kLbOW4CNZAW0BdR6MKVeLJSwLRHVYI+dY4wARSzYuhEUtWvW2ENmsPalV6-xq8DRpkstnp7kB1H8jKC-OQEVIiVSgDyGCERDF9CIAAkaDQPOZUAUMG35QAxRBEVlxxTqyCa1MTb2A6eFsprfDEXNCx+iucsADifKRiAWE2XaMEMwz6J2kYAPyDMc4FqKwwQEMgMFaDocFPI8fYaoyi7LquDabtukANnudCyK+LCCBEFiMARJGzFm7ISCSsLaGyDayuwzFNPgkDvlOlGQHhK5rkRO5EGRGC2AAGpAFYYFWkB-hgAFAeGzaoD2CgABThOgjS2CJBEbluElSYekQnmeVY9gAZJAMkAJQ3Ih8ooEQtj6VC4hGUuomEWZu77pZ1nnvZjn+AA3Nh164f5JnifGaITnmDCfrOxbCQlYlmbJPZgRkOlWCBmjAZGTmaMZuXbsF5H5SoV6aqEEi4LaszIBIsIjryWqWiBAnpUJtZYji5WCLYAAq8l6pWF4PNpfVlnqogSNIsgOZNMX9i1L7tWcXWQBYDbZlokBiu4tGIHwjTSvKtCqfwg0Fl+WUWl5SHIAVcE6QQn1lZ9lVoboWEvG8ZCQNWIiRJEsKTZSboRAQxC0ma8kYMgBT-D6500KsvhOOoEMZAAtJAzLEUJrrxA2tHE5AZMUx4DFSclR1gumsptG1j5iki4IIZ9kDFdoADkjDptYfAsD4JLFGzfAwpUeMtXwyBOei4AXRgV18G4-TVQ25iKUiskzQIilxsSpLklSNIAHQqWp43djpTOaNbjZkhSGDUnQ9uG6Z275S5YGkfu5NmTFYNgBDACq6AvswcZo9RkqMBabnIXCtJNpED18vqyfCw6rIciwkTcmyNqXN8BIa9AFBCTIfN0Jnn0yKKz5Tebc2QJ7rDe3b-tO446mQYIi3TQpSlgRB-XQbB0zz1nKGaJgreQHHjXPJAUFb5AgwPDFzcYK37c0MhPcz1bJJe7bvsO4H4m2NoeD+FPveW-3d+Dw-fsBxyoFYOG9+Zxx7PvFuSIL7IU7kibu4DD5qhPswSAUC25qGvrNb+A8h6P39s-PKb8P7yEgNPbBSlcH-yfkAoOHhbCgNpIg-eVCfYAPeugmByA4HkUMuAnsR9tp9mEX2cGHwIZ1B8JHKRUs3QsC5rgHGTR2SPjQjQEo-MGy0CghDTWwixEwHnP6XqrDh61wmD8YsmtMT1lxHXLKpj8F0hwvrfmFhTTynBELAEQIkzly0PCdgEhuLiBYJgOEtDjaolpCVSMKgZxFilJZF2oE4K3kksUQmkBRaBLJEQUWO8dr9BiR4yIXjL4p1OIwPg+E1xRLcWCcO5FnoZUSSwQhtVJIhVKWFWecErDblPDEOIOxciuHcF4XwAQd4vBEQY-iHwQD22Kf6e2897aoFIOIxuojGQNEUoCegsIFE0DFIo+RBtImVn5j4JGKNpxnOBG6bQRcEzIHYHzPiABqXmSJuYr28cwcAJU0ARNqUba5tIFBc0DvUugOkzh4B1iXTSLk5DfQyIgM5xVNLLPlBkpwoc4pHVxekl02AZooE2VoOMNicTV1kMkjSA1phNNkIVeCE89RL2mJyxs-1nG8umFYGGfBNCi3YKLZwcEhVksydgcVuSRD5JlcMTCvKHjSqFQMo0JAj7EqpnGBYlgPDczOJAOlDYhywg6eZbpoZelbLVTFLCQQunkUFSMZlkYeXL0FhU31vKRWVwwOK9kUrVUTnEP0TQAAWAATKqzVGFGrRycttTUo9TpcLZmKYIZSakBSIHC8AQkc0KE4f6q+lFTlgqLXC6KrxGT0EgFwVSEQBBLXutLfgub811pMiWstGl849sYBWs+0DME1rOYWwdkLG1NrIT6WQk50r3iBLiIgOY12gT6qvVwecC78FcHOupkLIDHxEaI2O2yYBLKxKefgay1AbK2TATWmpECjqGqgxQ+7PqHpHcexgDxzr9tFvbUAyykSrPWZs0W21d3nUuugPWFpfpVpiFw0OegsP22DXwaOuyc59hBRnNQ8ayrAbHaQ7WutZAWnjZhrgq9cNetY59Aj8oQ3EaAA) (details in the commented code)

PS: before creating this solution I tried to find a solution where the user could use type utils to declare the overriden types on his/her repo (like the current state of this PR), but I couldn't find a way to have an override of `useOnBeforeRender` with proper typing witthout the compiler complaining.

brillout:
Many neat and clever tricks, looks promising indeed.

I guess we can use this same technique for `PageContextInit`?

```ts
// User code

declare namespace VitePluginSsr {
  interface PageContextInit {
    user: string
  }
}
```

```ts
// vite-plugin-ssr code

declare namespace VitePluginSsr {
  interface PageContextInit {
    url: string
  }
}
```

```ts
// User code

const pageContextInit = {
  url: req.url,
  user: 'alice',
  bla: 42
}

// TS is OK with `pageContextInit.url` and `pageContextInit.user` but
// complains about unkown `bla`.
renderPage(pageContextInit)
```

Maybe we could simply keep names.

```ts
declare namespace VitePluginSsr {
  // Instead of `CustomPageContext`
  interface PageContext {
    cookies: Record<string, string>
  }
}
```

Makes it slighlty clearer that interfaces are being merged.

I guess we can then also add `PageContextInit` to `useOnBeforeRender()`?

Also thinking maybe we should name it `defineOnBeforeRender()`; I think it fits Vite's ecosystem slightly better (e.g. matching Vite's `defineConfig()`).

Thoughts / objections?

Really like that `PageProps` is being inferred, that's super convenient.

magne4000:
Just to be clear:
- `PageContextInit`: what we feed to `renderPage`
- `OnBeforeHookReturnCustomFields`: what possible fields are returned by `onBeforeRender` hooks (and thus available as return of `renderPage`)

The `PageContext` I declared in the playground was in fact `PageContextInit` I guess. Now regarding `OnBeforeHookReturnCustomFields`, I do not like the name, should we rename this one to `PageContext`, or something else?

I think we only need to override those 2 types, "what type do I give to `renderPage` for its page context at init (`PageContextInit`)", and "what type do I get back from it (the other type)". All intermediary types are internals and should be infered from those 2. Did I miss something here?

And OK for me to rename `useOnBeforeRender` to `defineOnBeforeRender`, I'm not really familiar with Vite yet.

brillout:
I'm thinking, do we really need 2 types?

How about having only one?

```ts
// User code

declare namespace VitePluginSsr {
  interface PageContext {
    // PageContextInit
    url: string /* (although this would be superfluous since `vite-plugin-ssr`'s internal
VitePluginSsr.PageContext interface would already includes all PageContextBuiltIn) */
    user: string
    // PageContext added by onBeforeRender() hooks
    redirect?: string
    cookies?: Record<string, string>
  }
}
```

So all init props are required while other props are all optional. Which makes sense.

This with `defineOnBeforeRender()` and `InferPageProps` would be the entire interface from the user perspective right?

Also maybe we can rename `InferPageProps` to `GetPageProps`; I'd say it fits slightly better the name of TypeScript's utility types https://www.typescriptlang.org/docs/handbook/utility-types.html.

magne4000:
:thinking: let me try that :+1:

magne4000:
Are `onBeforeRender` hooks allowed to modify properties defined by `PageContextBuiltIn`?

brillout:
Mostly not, but in some (rare) situations yes. But I think it's ok if we are not type safe about this (we can always complain at run-time).

magne4000:
It would look [something like this](https://www.typescriptlang.org/play?#code/PQKhCgAIUhRAPAlgZwC6IHYHNKoJ4AOApslCMOIgLYED2ATqpAN6QDG9RAhqkQApcsRAEpEMAEyKd6kAL6QAZvVpVIAcgB0wDt14ChoiVKlrw4IvDqNchIpH1EAwrQy94qAEIBXRABtUAJIYkAC8LFD2gkQAXJBcGHgRBFEIVqjIsaJsDOIAPGj0mFgANJBeGADWGLQA7hgAfBHKXnpc9FxUGZBZOfmohdilBUWNkGX0vrHD2BFeEwKoABYYHTGQ01izEwByDFRcvogAXkTiU-1FW74C9MinscwRY8lLK1RrG0-r3PRsi7EYLy+XyQAA+3SI2XoeQ2Qwu2FGY0giy4yH+kEBwLB63hmzGsnABPAoAg0Eg21WyGSbDsSx4kBqdjY8UgyUYiAOvjwkFoADcpIVJGQKBY0pBJGxfG07G8SNS7AA1RB6XxeLCYADKyBkjyRkGAwHWi1oQPEkAARnY+QLEJIMBFRQwmJhePQFFwaZEhM5XBYmMwiWMDesVEQaijnb76CsQfhiA7LE7IC6pO7PQBxMRSRBsBw+tz+r7Jb0uAsAfgeX2eUT4ygIyArcQSXwJ+LMY0d1hTbo9djzpb9QWV4T1c0mOIGeLk7f1hoA8hgPEQFAwiAAJWi0CqQTioObBON2P3tNhMFcyThccRcc1+ZXc+JmzheZA3u-4BNiw+QBdL8-rzcKlEPdo1yAANSA-TEcRkEgTMMGzXMonzP16lCL4AAobnQA5cn7X13DQgAySAwIASmxWsVBQIg8LaHDfDoksCNQYjSPqABuGdg1-ZdV0MSQZGLJwB3cZEuH5EN3mTKMYy5dgVAIFpTlZOtYLpJhGTicRxE-JNuzTOxeP-ASpHwgtIPcaDYPMv1vD8QIMFKWyxIDbj50XPjOA3LcbHjDtEy7KNDJ-Tz-x8ipwLQ3UkQw7IWNiYz+OgszkNE1AyMSsLVwi4D9yiiJA1nMp0FjWhINRPxuR0Hg7C4RRylPRAXAOSBnBoFwxFQPSgtdEKHDwtToswgg1NiKj60yptEjbIkSTISAABEvCoKhuQAFQ1Bl2gIYgZFQcqxGQOZaUWOwV2BWoimiBaxgAWmRQC-LsFAFPoThT1uyAHuNXzd33Z72BZS0yjuM0DpkhQpFZGs1MgOL4jUJgQfETqGWVRZAZceTkFoVV0BcZAyOFcAFEagngkkBRMCIJLOFM+hwMs3gJFgpUVTVTVtQ0eDEJc1iMN+ipYnZ-hVXVDAtXoDQ6YArcooomL-ujR6ty4uawAWgBVUrnvUw73BPJgHAm2CGC9ESOoQ1w9fh7JTQtOwqGtM0XXK2UqV7YnoAob9M1QE21KCV0Y1ydbmesyBRb4cWuel3nCiQ5iCzQsJw6g1mRxh5O-UbGLqyEU3G0wKGZC1wrCUgMtIC1yAHlkLi-aIAPYdoesw4jzPo9jyXudliLcniPB6lTyB06srvlTFzne+l-vANyEvoa1tCywif3A7b5Bg6kUOV7rlgG-AJuW6EDuM5gqOp5jmepZl7LvIXoeR9CMfO8v7vb77h+5cipey9XhET+Es74DQ3q3esO9oy4RXmheuXEzCIMQfNaAC1dg1Eehg1GDI7AvlOnYQ45oZJNloEsaGnA6Br1QT7JBKCYBzn5O9W0ipr49ylhiSk8oSadiYBKKUnAOHvE9p6YBccs7BlEAARx8JwM0o0t5xAEfsCQPAGDcnDGIIGwIig7hSvQBwEQeIEApq1eR9ZFFWmMc1GMDIzrBGVhgHRZDVBLGaFgTGLg-zJSMPQCIBlewWxQq5L4uN3hLCKEOVA5xJzFQALKPlUfQdRditGHGwLonxBi9TZC3IgEgjYejQj6JOUo5Qqi1AaMVRwwM7AOJUpgHkP8GYVyJLQzW1CYAgA0HcegjCNDCQ0OkMgJNiSGmUdeA6STQZWgULgM60zZYM22lwXa0NECzK0jUeITBvyvihvgSAABqSGxg5EQPUrYcA8U0CNK8fTPRr8qY00WXojCqI8AYDYHbdKFEQjDTGOs75LFumhnCdgSJitirxQLCCsJiwImOKYAcLZeBYIWBQEMvUDis56mEkEqJOK9TZ34GNQlRKFKrRcLENQXA1DFCrES3kBwvBrDUEy1URBTDkrbOS2Q9LyU5IqHkroAZirflen8SEFQVKonFJCfhKlzTcn5i2LibYmjNwBvnYl+LKzkuEqbPV3LsiUowNS80dKGVjHiKQs69AFTMrWAAFgAEwMr5S2L4wZqiQAAAaCuFb65Exg4iwWVGbKxLVfCFXVmRBBwYE61QtqbQGSgVC3K8iIPRx9bDJrhmEcBhc1K5EPLQWZnjM0M04qMnkMghAIXaLwQJilOo22-Gm1QFaTLZu-P2K2XVX6FtoqW8tTS9HVpretFJPDbH0i0ghU45hAr+jzVvZyaV+2uFKF27xgk5AILaaSI90ANCSjya4fpURBmkA6RQYM1Avy5tYJves66SybqYPIDt6gtBdJ6X0gZ6Q1AIJnWTT5FMLYYQNaSl9RMs7Qa3qelQzsMDqzMMGBgZhrnGyiC68aG66DWyYGEMDTUXAWxdVB85iskjnKQ6atD6H2nHqkSym5KIJC+CkKULgChXRlAIBMnR3YbFxiceVLjjCohXw5iA7Ugi5S9mGTQq5hMmCcEydJsINVWgGD0VIDCAZibYYUixV+GnBIOEM+Mccag6VSWbvC8FiLqVqDkHGrD6UkO5JIGK3NhSYS4lKZUaodQ0LgnKE8hC4ggA).

The only remark I have concerning `PageContext` as a unique type is that you must really be aware of what the prop being optional means. My personal feeling is that I would, as a user, understand thing easier if I separated what I give to `renderPage`, and what `onBeforeRender` can return. Having what defines them in one type could be confusing. But I might be wrong, and I do not have a big opinion on this.

brillout:
Good point. I'll think about it.

brillout:
I think you are right. How about this then:


```ts
declare namespace VitePluginSsr {
  interface PageContextInit {
    user: string
  }

  interface PageContextOnBeforeRender {
    // Optional prop
    redirect?: string
    // Required prop: always available in `render()` hook and
    // if `renderPage()` is succesfull also always available there.
    pageExports: {
      documentProps: {
        title: string
        description: string
      }
    }
  }
}
```

Also, how about `typescript('onBeforeRender', () => {})` instead of `defineOnBeforeRender(() => {})`?

magne4000:
> I think you are right. How about this then:
> 
> ```ts
> declare namespace VitePluginSsr {
>   interface PageContextInit {
>     user: string
>   }
> 
>   interface PageContextOnBeforeRender {
>     // Optional prop
>     redirect?: string
>     // Required prop: always available in `render()` hook and
>     // if `renderPage()` is succesfull also always available there.
>     pageExports: {
>       documentProps: {
>         title: string
>         description: string
>       }
>     }
>   }
> }
> ```

I like this solution :+1: 

> Also, how about `typescript('onBeforeRender', () => {})` instead of `defineOnBeforeRender(() => {})`?

That's unusual, but it's reusable and understandable that it should be used only when using Typescript, so why not.

brillout:
Or `withTypescript()`. Makes it slightly clearer that `withTypescript()` is a dummy function.

magne4000:
Regarding:
```ts
...
    // if `renderPage()` is succesfull also always available there.
    pageExports: {
      documentProps: {
        title: string
        description: string
      }
    }
...
```

_succesful_ means that `httpResponse !== null`?

brillout:
Yes

brillout:
It's `null` <=> there is no `_.error.page.js` or it couldn't be rendered.

Actually I'm just realizing that `// if renderPage() is succesfull also always available there.` is wrong.

If a hook throws an error and there is a `_error.page.js` then `pageContext` will miss all required props even though it's not `null`.

magne4000:
```ts
/**
 * Override VitePluginSsr namespace
 */
export declare namespace VitePluginSsr {
  export interface PageContextInit {
    somethingInit: string // Mandatory when calling renderPage
  }

  export interface PageContextOnBeforeRender {
    cookies: Record<string, unknown> // Can be returned in onBeforeRender
  }
}

/**
 *
 * ---------- *.server.page.ts
 *
 */

// mandatory use of the useOnBeforeRender wrapper if we want type safety + infered pageProps type
const onBeforeRender = withTypescript('onBeforeRender', async (context) => {
  if (context.somethingInit) { // context.somethingInit always exists
    return {
      pageContext: {
        pageProps: {
          common: 'a',
          value: 'value'
        }
      },
      cookies: {} // type is checked as declared by PageContext
    };
  }

  return {
    pageContext: {
      pageProps: {
        common: 'b',
        anotherValue: 42
      },
    }
    // no `cookies` here as its optional
  };
});

// Generate PageProps type from onBeforeRender
type PageProps = GetPageProps<typeof onBeforeRender>;
// or generate PageComponent type from onBeforeRender
type PageComponent = GetPage<typeof onBeforeRender>;

// Then export what we need
export { PageProps, PageComponent, onBeforeRender };

/**
 *
 * ---------- *.client.page.ts
 *
 */
// import type { PageProps, PageComponent } from './*.server.page.ts';

export function Page(pageProps: PageProps) {
  pageProps.common;
}

// or

const Page2: PageComponent = function Page2(pageProps) {
  pageProps.common;
}


/**
 *
 * ---------- renderers
 *
 */

async function test() {
  const renderPage = createPageRenderer({})
  const context = await renderPage({ url: '', somethingInit: '' });

  context.pageExports // optional
  context.cookies // optional

  if (context.httpResponse) {
    context.pageExports // not optional
    context.cookies // optional
  }
}

withTypescript('render', function render(pageContext) {
  pageContext.pageExports // not optional
  pageContext.urlParsed // not optional
  pageContext.somethingInit // not optional
  pageContext.cookies // optional
});
```
There are still things that I'm not sure about. There is a `PageContextBuiltIn` and a `PageContextPublic` type, are both in use? Could we merge them into just one type one move it to `node/types.ts`?

magne4000:
:tada: I successfully managed to override the types from another repo like so:
```ts
import 'vite-plugin-ssr';

declare module "vite-plugin-ssr" {
  export namespace VitePluginSsr {
    export interface PageContextInit {
      initProp: string
    }

    export interface PageContextOnBeforeRender {
      renderProp: string
    }
  }
}
```

magne4000:
> If a hook throws an error and there is a `_error.page.js` then `pageContext` will miss all required props even though it's not `null`.

Does this mean we can just return `Partial<everything>` in `renderPage`?

brillout:
Will review tentatively tomorow.

brillout:
Tomorrow (super excited about this; just been busy with high prio things).

examples/react-full/pages/star-wars/index/index.page.server.ts
```diff
@@ -1,11 +1,16 @@
...
 import type { Movie, MovieDetails } from '../types'
 
 export { onBeforeRender }
 export { prerender }
 
-async function onBeforeRender() {
+const onBeforeRender = withTypescript('onBeforeRender', async (pageContext) => {
+
+  // `documentProps?: { title: string }` is missing
+  pageContext.pageExports.documentProps?.title
```
brillout, line 12: Doesn't seem to work? TS complains that `pageContext.pageExports.documentProps` is of type `unknown`.

examples/react-full/pages/star-wars/index/index.page.server.ts
```diff
@@ -1,11 +1,16 @@
 import fetch from 'node-fetch'
+import { withTypescript } from 'vite-plugin-ssr'
 import { filterMovieData } from '../filterMovieData'
 import type { Movie, MovieDetails } from '../types'
 
 export { onBeforeRender }
 export { prerender }
 
-async function onBeforeRender() {
+const onBeforeRender = withTypescript('onBeforeRender', async (pageContext) => {
```
brillout, line 9: Seems like something is broken; TS complains that the function doesn't match what `withTypescript` expects.

examples/react-full/renderer/pageContext.d.ts
```diff
@@ -0,0 +1,24 @@
...
+  namespace VitePluginSsr {
+    interface PageContextInit {
+      pageExports: {
+        documentProps?: {
+          title: string
+        }
+      }
+    }
+
+    // Is it possible to also support `pageContext` on the browser-side by doing something like the following?
```
brillout, line 13: Thoughts? So that we can also provide `PageContextInit` to [`getPage()`](https://vite-plugin-ssr.com/getPage) and [`useClientRouter()`](https://vite-plugin-ssr.com/useClientRouter).

brillout (changes requested):
Super looking forward to this.

I've more time capacity now; I should be able to be more responsive from now on.

examples/react-full/pages/star-wars/index/index.page.server.ts
magne4000: There seems to be some conflict between default and extended types. I can fix that by naming Internal and External typing different and merging them into a third type

examples/react-full/pages/star-wars/index/index.page.server.ts
magne4000: IDEM

examples/react-full/renderer/pageContext.d.ts
magne4000: It would be great, I'll need to take a look at it

examples/react-full/renderer/pageContext.d.ts
```diff
@@ -0,0 +1,24 @@
+import 'vite-plugin-ssr';
+
+declare module "vite-plugin-ssr" {
+  namespace VitePluginSsr {
+    interface PageContextInit {
+      pageExports: {
+        documentProps?: {
+          title: string
+        }
+      }
```
magne4000, line 10: Where does `pageExports` comes from? That the init context and `renderPage` in `server.index.ts` complains that it is not there. I'm guessing that is something that is infered somewhere?

magne4000: Or it perhaps should be optional

magne4000: :thinking: I think there should be specific native handling for `pageExports` values. Currently they are defined in both `PageContextBuiltIn` and `OnBeforeRenderPageContextInternal`, and I think that is where they should be either merged or used differently.

EDIT: check my proposal in the PR where I defined a `PageExportsInternal` type

examples/react-full/renderer/pageContext.d.ts
```diff
@@ -2,22 +2,18 @@ import 'vite-plugin-ssr';
...
 
     // Is it possible to also support `pageContext` on the browser-side by doing something like the following?
     interface PassToClient {
       value: 'pageProps' | 'documentProps'
     }
 
     interface PageContextOnBeforeRender {
-      documentProps?: {
-        title: string
+      pageContext?: {
```
magne4000, line 13: I added `pageContext` here so that it matches what I defined in the types.ts files, but we could also auto wrap `PageContextOnBeforeRender` into a `{ pageContext: PageContextOnBeforeRender }` in types.ts. It just a matter of letting this responsibility to the user or not

vite-plugin-ssr/node/types.ts
```diff
@@ -1,9 +1,16 @@
...
+  documentProps: {
+    title: string
+    description: string
+  }
+}
+
 export type PageContextBuiltIn = {
   Page: any
-  pageExports: Record<string, unknown>
+  pageExports: Record<string, unknown> & Partial<PageExportsInternal>
```
magne4000, line 13: A little more accurate by default

examples/react-full/renderer/pageContext.d.ts
magne4000: Let me check that once this PR is merged or stable enough to be merged

brillout: Ok

examples/react-full/renderer/pageContext.d.ts
brillout: > Where does `pageExports` comes from?

It's a `pageContext` built-int, see https://vite-plugin-ssr.com/pageContext

brillout: So `pageContext.pageExports` is set by and comes from `vite-plugin-ssr`.

> check my proposal

Does it means I would need to do the following?

```ts
interface PageExports {
  documentProps?: {
    title: string
  }
}
```

Would it be possible to do what I tried instead?

```ts
interface PageContextInit {
  pageExports: {
    documentProps?: {
      title: string
    }
  }
}
```

Or do you just find it less user friendly?

I'm wondering, from the typings perspective it only makes a difference in the sense that `pageExports` is not required in `renderPage()`, am I right? Maybe a `Omit<PageContextInit, 'pageExports'>` for `renderPage()` would do the trick here.

examples/react-full/renderer/pageContext.d.ts
brillout: If it's possible then yes; other than `pageContext` I don't forsee any other properties.

So I guess we can straight for:

```ts
interface PageContextOnBeforeRender {
  redirectUrl?: string
}
```

brillout: Although, I'm thinking, how would the user assign everything at once?

```ts
interface PageContextOnBeforeRender {
  // Is this possible if we don't have the `pageContext` indirection? 
  pageContext: PageContextTypeDefinedByUser
}
```

Or maybe we do this:

```ts
// This naming seems nice as well
interface OnBeforeRender {
  pageContext: PageContextTypeDefinedByUser
}
```

brillout:
MacOS tests will likely go green if we rebase on master.

examples/react-full/renderer/pageContext.d.ts
magne4000: Currently with last commits you don't even need to define a `PageExports` or `PageContextInit['pageExports']`, as `PageContextBuiltIn['pageExports']` is now more specific.

examples/react-full/renderer/pageContext.d.ts
magne4000: > ```ts
> // This naming seems nice as well
> interface OnBeforeRender {
>   pageContext: PageContextTypeDefinedByUser
> }
> ```

That's a good idea, the interface name not containing `PageContext` and having a `pageContext` is clearer I think, and it's future proof if we want to return something other then `pageContext`

vite-plugin-ssr/node/types.ts
magne4000: That what makes unecessary for the user to redefine:
```ts
interface PageContextInit {
  pageExports: {
    documentProps?: {
      title: string
    }
  }
}
```
as it's by default now

examples/react-full/renderer/pageContext.d.ts
magne4000: renamed in 1166de1

brillout:
I pushed a commit that removes `documentProps` from the built-in because `documentProps` is defined only on the app-side; `vite-plugin-ssr`'s source code doesn't know anything about `documentProps`. The only thing the plugin knows is that `pageContext.pageExports` is `Record<string, unknown>`.

I also pushed a commit to rename `PageContextInit` to `OnInit` but I'm not sure why it doesn't seem to work. (I'm guessing overriding `OnInit` doesn't override `OnInit['pageContext']`.) Not sure how to fix this, maybe you have an idea. Also let me know if you object the renaming.

Otherwise it works, super exciting. Also added a commit to try `GetPageProps` which works like a charm.

magne4000:
> I pushed a commit that removes `documentProps` from the built-in because `documentProps` is defined only on the app-side; `vite-plugin-ssr`'s source code doesn't know anything about `documentProps`. The only thing the plugin knows is that `pageContext.pageExports` is `Record<string, unknown>`.

👌 good to know

> I also pushed a commit to rename `PageContextInit` to `OnInit` but I'm not sure why it doesn't seem to work. (I'm guessing overriding `OnInit` doesn't override `OnInit['pageContext']`.) Not sure how to fix this, maybe you have an idea. Also let me know if you object the renaming.

You guessed right, extending an interface doesn't merge it's props like `type C = A & B` would, so creating `interface AInternal {}`, `interface A {}` and `type AMerged = AInternal & A` is a good way to circumvent this issue. We now also have the opportunity to merge internal and user defined types as we see fit.

brillout:
Makes sense and thanks to the fix it works now 👌

Is there a way to forbid non-existing props, e.g. https://github.com/brillout/vite-plugin-ssr/commit/3cbbec11d13bd2802c39f7c6b3ea1da75de0422c?

magne4000:
For returned values I don't know if it's possible, especially considering that it's a nested type and we are infering some types from it, I do not see a way to do that.

brillout:
Yea I'm also not sure if it's even possible. It's quite sad though because it kind of defeats the whole purpose of defining `VitePluginSsr.OnBeforeRender.pageContext`.

magne4000:
It at least provides autocompletion and type checking on those nested properties

brillout:
True. Although I'm thinking, if it's worth it then...

I'm thinking that defining the `pageContext`'s type directly and using that defined type throughout the app is a simpler approach and is only a negligible DX deterioration in comparaison to what we are offering here.

Thoughts?

Maybe we can release it as an experimental feature.

Btw tests are green now :-).

magne4000:
That should indeed be enough. This means that we should just get rid of `withTypescript` helper and `OnBeforeRender` type. We keep `OnInit` and utils like `GetPageProps`. Is that what you were thinking?

magne4000:
btw, I think I found a way to forbid unknown nested props, but I think that it's not easy to debug:
image

Also it could generate falso positive as its very hackish (it leverages [ts-toolbelt `Exact`](https://millsp.github.io/ts-toolbelt/modules/function_exact.html) helper)

magne4000:
So, back to simpler typings and utils:
- No more interface override, no more namespace. Instead types are declared locally
- Most types helpers are gone, we just keep the ones that are necessary
- `GetPageProps` and `GetPage` util are kept
- `GetPageContext` replaces interface overriding. See usage
- Types that can be returned by `onBeforeRender` for `renderPage` caller to use are typed when calling `createPageRenderer`. Example:
```ts
const renderPage = createPageRenderer<{ cookies: any }>({ viteDevServer, isProduction, root })

...
const pageContextInit = {
  url: '',
  something: 42
}
const pageContext = await renderPage(pageContextInit)

pageContext.something // infered, always exists
pageContext.cookies // can exist because defined when calling createPageRenderer
```

brillout:
Exciting, will have a look :-)

brillout:
> I found a way to forbid unknown nested props

That's actually super neat and it would make our namespace solution totally worth it, I think.

> it could generate false positive

How so?

> its very hackish (it leverages ts-toolbelt Exact helper)

ts-toolbet seems quite popular, so I guess if they couldn't find a simpler solution then there is not much chance we can find something simpler 😕.

I really do like setting types globally with namespaces and I think the only thing missing is forbidding unknown props.

magne4000:
>> it could generate false positive

> How so?

Typing definitions are more complex for this use case, and having them not quite doing the right thing is probable. But I think that can fixed with TS unit test. Also I do not want this to be hard to maintain. So I guess we'll have to see.

I'm going to create another branch that will be a mix of the code that I stashed and this PR (tomorrow I think) so you can see all that by yourself, so we can make a decision on which solution we want to continue with (so it will be another PR).

brillout:
Sounds good.

Note that we can make the whole / some TypeScript support a beta feature. (E.g. see how https://vite-plugin-ssr.com/stream is marked as beta.)

magne4000:
> Note that we can make the whole / some TypeScript support a beta feature. (E.g. see how https://vite-plugin-ssr.com/stream is marked as beta.)

I think that would be a good idea in all cases :)

I'm not gonna have the other branch ready today, harder to merge than I thought. I'll take the time to do that properly on next monday

brillout:
No rush :-)

magne4000:
I think it's doable, but I'm having many false negative typings for the moment, so it's not usable. You can check [work in progress](https://github.com/brillout/vite-plugin-ssr/compare/master...magne4000:advanced-types-bis?diff=split#diff-5cf8eed05f630748310f28b6f3f0b5a53bb2a595a19a094d78859fdd35a0ec17), but I'm not sure I'll give it more time soon.

brillout:
> I think it's doable, but I'm having many false negative typings for the moment, so it's not usable. You can check [work in progress](https://github.com/brillout/vite-plugin-ssr/compare/master...magne4000:advanced-types-bis?diff=split#diff-5cf8eed05f630748310f28b6f3f0b5a53bb2a595a19a094d78859fdd35a0ec17), but I'm not sure I'll give it more time soon.

👍 Let me know whether you want to further try this or not.

magne4000:
> 👍 Let me know whether you want to further try this or not.

I don't think so. The branch will stay there if anyone want to take a look. We should focus on the current simpler PR right now so that we can merge it

brillout:
Still no luck on forbidding unkown props?

brillout:
Actually, I'm wondering: the following works right?

```ts
function onBeforeRender(): Partial<VitePluginSsr.OnBeforeRender.PageContext> {
}
```

So I'm thinking we should be able to use this for `withTypescript` as well, or am I missing something here?

magne4000:
You will not be able to extract pageProps by hard coding the return type, that's the whole difficulty. Hard coding the type we can easily (I think) restrict props, but if we want to be able to extract pageProps, this type can't be hard coded

magne4000:
I did some more tests with simple cases. I either succeed to only allow specified fields in `onBeforeRender` return type, or infer `pageProps`, but never both successfully.

We can:
- Keep `PageProps<>` extraction. `onBeforeRender` return type is not validated (current state of this PR)
- Only allow valid fields in `onBeforeRender` return type, with user specified `PageProps`

brillout:
Damn, so frustrating to be so close to perfection.

Which one would you choose?

magne4000:
Hard to tell, I think giving the choice to both could be good (you just can't use them at the same time), but it's hard to explain which one should be used.
But perhaps I talked too soon :slightly_smiling_face:, I thought of another approach to solve the issue. I'll push the test with some explanations in a sec

magne4000:
[Here you go, all infos in the comments](https://github.com/brillout/vite-plugin-ssr/pull/219/commits/fad11c0c1c80801f9164b1de4c71f949bcdc63bb)

magne4000:
In my project I use utils like that:
```ts
// ---  utils

type Context<P> = OnBeforeRenderPageContext & { pageProps: P };

// equivalent of `wrapResponse`
export function ctx<T>(...ctx: Context<T>[]): { pageContext: MergeAll<{}, typeof ctx> } {
  return {
    pageContext: Object.assign({}, ...ctx),
  };
}

export function redirect<T>(target: Context<T>['redirectTo']): Pick<Context<T>, 'redirectTo'> {
  return {
    redirectTo: target,
  };
}

export function props<T extends object>(p: T): { pageProps: T } {
  return {
    pageProps: p,
  };
}



// --- usage

async function onBeforeRender({ urlParsed }: PageContext) {
  return ctx(
    props({ a: 1 })
  );
  // or combine multiple utils
  return ctx(
    props({ b: 2 }),
    redirect('/home'),
  );
}
```

brillout:
Nice.. but I guess there is no way to achieve the same with a single function `withTypescript()`?

magne4000:
For the moment I'm unable to find a way to make it work with `withTypescript`. The last code examples are what I'm currently using on my project, and I think this is the best that we can have right now. It's easy-to-maintain TS, it handles props validations ans forbid unknown props, and the reason I created those in the first place was also to have a modular way to return multiple typed properties, thus avoiding typos and such

brillout:
I can see that's it's quite perfect for you, but I'm a bit worried about the DX. A single `withTypescript()` would be quite a pleasant DX but, yea, I guess we can't achieve that. Actually, I'm assuming that TS always needs one side of a type assignement to be the "source of truth" and we are essentially trying to have both side to be the "source of truth".

Btw., our discussions have been super insightful and I'm actually reusing some of this PR's work for Telefunc.

magne4000:
I'm closing it for now then

brillout:
It's really a bummer, I would have loved `VitePluginSsr.OnBeforeRender.PageContext` / `VitePluginSsr.OnRenderStart.PageContext` / ...  + `withTypescript()` + `GetPageProps`. Would have been perfect.

Btw. `Telefunc.Context` works really well for Telefunc.

Thanks a lot for our disucssions and the PR. Greatly appreciate it.

Rom :-)

magne4000:
We could certainly add `GetPageProps` on its own, I'm using it, and it jsut requires the following lines (could even be just in the doc):
```ts
type GetPagePropsInternal<T> = T extends {
  pageContext?: {
    pageProps?: infer U;
  };
}
  ? U
  : never;
type AwaitableReturnType<T extends (...args: any) => any> = ReturnType<T> extends Promise<infer U> ? U : ReturnType<T>;
export type GetPageProps<T extends (...args: any) => any> = GetPagePropsInternal<AwaitableReturnType<T>>;
```

brillout:
I was actually thinking of what parts of the PR we can merge.

Also the namespace thing actually, for being able to add `user` to `pageContextInit` and have it available automatically for each `onBeforeRender()`.

But I'm not sure about the naming.

magne4000:
> But I'm not sure about the naming.

Story of my life :smiling_face_with_tear: 

With the namespace comes the small refacto around `renderPage` typing. As soon as we find a good naming I can write another PR for that.

brillout:
I'll be thinking about naming again after I release Vike's beta.

brillout:
(Branch `advanced-types-with-namespace` moved to https://github.com/brillout/vps-advanced-typescript-support. @magne4000 You can ignore this, nothing new here.)

PR #220, opened 2021-11-29
[Fix] Bundle app before deploying to Vercel

patryk-smc:
I made some assumptions when updating javascript example, based on my typescript project. It turned out that it didn't work. Theoretically, Vercel should bundle javascript serverless functions on their side, but they do not do that. This PR adds another step in deployment process, which uses [@vercel/ncc](https://github.com/vercel/ncc) tool which bundles entire app to a single, lambda-friendly file.

### DIFF
--- a/examples/vercel/package.json
+++ b/examples/vercel/package.json
@@ -6,6 +6,7 @@
     "dev": "node ./dev-server"
   },
   "dependencies": {
+    "@vercel/ncc": "0.32.0",
     "@vitejs/plugin-vue": "^1.2.3",
     "@vue/compiler-sfc": "^3.0.11",
     "@vue/server-renderer": "^3.0.11",
--- a/examples/vercel/vercel/deploy.sh
+++ b/examples/vercel/vercel/deploy.sh
@@ -15,10 +15,10 @@ mkdir .output
 mkdir -p .output/static
 cp -a dist/client/. .output/static
 
-# Step 5: Copy render function and rename it to index.js
-# If you are using typescript, you should transpile it to javascript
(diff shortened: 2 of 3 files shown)


brillout:
So Vercel bundles when the serverless function is TypeScript but doesn't when it's JavaScript?

patryk-smc:
No. I had this bundling step previously in my TS project, but thought that it is not necessary for JS, as their docs mention that they do bundling on their side.

brillout:
Thus we shall have a test at some point :-)

brillout:
Especially considering Vercel's past of breaking changes.

patryk-smc:
@brillout Have you seen this? https://vercel.com/docs/file-system-api#presets/framework-detection

brillout:
I did but it doesn't add that much value?

patryk-smc:
I'd say the highest value here is reducing friction for new developers to get started with vite-plugin-ssr. I am not sure how hard it would be to maintain this, but it would for sure get more people interested in this stack :) Vercel itself is promoting frameworks with direct integrations. https://twitter.com/leeerob/status/1465414189829304326?s=12

brillout:
> reducing friction

What steps would it remove from https://vite-plugin-ssr.com/vercel ?

brillout:
I believe it only sets the right settings; I believe we still need to apply the steps ourselfs (but correct me if I'm wrong).

What we can do is to create a `vite-plugin-vercel` that works with any Vite app. (Regardless of whether vite-plugin-ssr is used or not since `vite-plugin-ssr` is just a Node.js server middleware.) Up to (co)-work on this?

PR #260, opened 2022-02-11
Add example code for SST (Serverless Stack) + Vue

ffxsam:
Some initial comments:

1. This is a non-working state, and I need some help with this.
2. There are two vite.config.js files, mostly because I wasn't sure where it belonged (feel free to remove the incorrect one).


brillout:
Feel free to hit me up when you're done with it.

ffxsam:
@brillout I'm afraid I think I've gone as far as I can without help.

Run `yarn build:vite && yarn start` (note that this will create two CloudFormation stacks), and open the API URL shown by the CLI. For some reason, `httpResponse` is undefined, and I can't figure out why. `src/entry.ts` is the Lambda handler.

brillout:
Let's move this to an extra repo and add an entry to https://github.com/brillout/vite-plugin-ssr/tree/master/examples#community-examples once it works.

I'm busy with high prio things right now, I may have some time to have a look at it after I release Vike's MVP.

ffxsam:
@brillout Done! https://github.com/ffxsam/vite-ssr-sst

I also added you as a collaborator.

brillout:
👍 

Definitely hit me up when you get this working.

What is the problem btw.?

brillout:
> watch your console and you'll see that httpResponse is undefined for some reason.

That doens't seem to be the root cause. By bisecting inside vps's code, you should be able to get down to it.

ffxsam:
Yeah, I'm still digging into it. Just takes a bit more time since I'm unfamiliar with the code.

brillout:
It used to not be a 404

ffxsam:
I abandoned the project after spending too much time stuck.

I also discovered that Nuxt 3 now uses Vite as its core, and a new server engine called Nitro (which makes it super easy to use with AWS Lambda). So I switched to that instead and put this together:

https://github.com/ffxsam/nuxt-sst

brillout:
@ffxsam Thanks for the heads up 👍

PR #282, opened 2022-03-17
Fixed broken link to workflow example

jrson83:
The `vite-plugin-ssr.com`  link to workflow example was pointing to a 404 Page.

https://github.com/brillout/vite-plugin-ssr/blob/master/.github/workflows/docs-website.yml

Instead of:

https://github.com/brillout/vite-plugin-ssr/blob/master/.github/workflows/docs.yml


brillout:
Good catch, thanks!

PR #286, opened 2022-03-22
Added passive EventListener

### LINKED ISSUE #285: Prefetch EventListener touchstart (passive)
If you run an example with `npm run prod` and check the rating with Google Lighthouse, it has almost perfect (99%), just [suggestes to add a passive EventListener](https://web.dev/uses-passive-event-listeners/?utm_source=lighthouse&utm_medium=devtools#how-to-make-event-listeners-passive-to-improve-scrolling-performance) to the `touchstart` event, which is happening in the file `dist/esm/client/router/prefetch.js`.

2022-03-22_11h51_35

I checked and added it to the code:

```js
linkTag.addEventListener('touchstart', () => prefetch(url), {passive: true});
```

When I run the test again, it has a rating of 100%. I would do a PR for this?

jrson83:
## Reference Issues/PRs

resolve #285

## What does this implement/fix?

- Added passive EventListener to touchstart event

### DIFF
--- a/vite-plugin-ssr/client/router/prefetch.ts
+++ b/vite-plugin-ssr/client/router/prefetch.ts
@@ -68,7 +68,7 @@ function addLinkPrefetchHandlers(prefetchOption: boolean, currentUrl: string) {
       observer.observe(linkTag)
     } else {
       linkTag.addEventListener('mouseover', () => prefetch(url))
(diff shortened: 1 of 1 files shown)


brillout:
Good catch 👌

PR #291, opened 2022-03-30
Fix duplicate decodeURI logic

ptb:
The current code uses `decodeURI` on `url` then asserts that it `isParsable` which does a second `decodeURI`. If the initial decoded URI contains a `%` character, the second `decodeURI` in `isParsable` will fail. Asserting `isParsable` before the `decodeURI` in `parseUrl` allows a `url` containing a decoded `%` character to be parsed successfully.


brillout:
Good catch 👌. Released in `0.3.61`.

PR #296, opened 2022-04-04
pinia example added

luke951:
Added example of using `vite-plugin-ssr` with Vue and Pinia.


brillout:
@luke951 Looks good 👌. One thing: let's align the example with the boilerplate, so that we can run the boierlplate tests, see https://github.com/brillout/vite-plugin-ssr/pull/296/commits/dd89816ae7c6e405348153bf761209d32e1c3e06.

luke951:
@brillout OK, I updated the example with the boilerplate.

brillout:
Good idea to create a custom test instead of re-using the boilerplate test suite.

Let's push it forward:
 - Let's reduce the test to the bare minimum (only testing that the counter button is working)
 - Let's reduce the example code to the bare minimum. No logo, no layout, only one page, only a counter.

Let me know if you object.

Looking forward to merge!

luke951:
@brillout  I updated .testRun let me know if that works 'cos I cannot run this test locally for some reason.

brillout:
The tests are failing. Seems like you modified the `libframe/` submodule?

>  I cannot run this test locally for some reason.

What's the error?

luke951:
@brillout As far as I know if you want to test pinia state management you need to use @pinia/testing library so I added this package. I don't think unit tests are that important when it comes to the counter. I just wanted to add an example of using pinia similarly to an example with vuex.

brillout:
I didn't mean unit testing. Same test you had earlier but stripped to down to test only one page.

brillout:
Closing. But feel free to re-work on this and I'll gladly re-open.

PR #303, opened 2022-04-09
Improve types boilerplate vue ts

louwers:
Composables can be typed using `InjectionKey`.


brillout:
👍 Would it also make sense to update the JS boilerplate-vue?

louwers:
@brillout Not really, since it's a type. I could add the check & throw, but that is only done so `usePageContext` always returns `PageContext` and not `PageContext | undefined`. ;)

brillout:
I see makes sense :-).

PR #304, opened 2022-04-13
Update cloudflare-workers documentation

Immortalin:
Add an updated example with a shorter worker config, streaming and esbuild.

https://github.com/Immortalin/vite-plugin-ssr-cloudflare-pages-demo


brillout:
That's quite a nice example 👌.

PR #313, opened 2022-04-29
fix: isolate framework pageFiles symlink directory

XiNiHa:
> VPS symlinking the framework pageFilesDir to node_modules breaks exports from the framework (since it doesn't copy contents other than pageFilesDir, which  means that files including package.json is not there)

Fixes issue [mentioned here](https://discord.com/channels/815937377888632913/815937377888632916/969417390771290182)


brillout:
👍 Will merge after the branch is green again.

brillout:
Let's see if it causes problems (I don't think so) then I'll release a new beta.

PR #319, opened 2022-05-06
add support for urls to start with . to allow raw html files to impor…

jobarah:
**URL Format Validation Update**

The below changes add support for urls, such as assets imports to start with a . eg: `./assets/file.ext`.

Use case for these changes are when using ssg simply to generate raw html and wanting to serve the html files outside vue's ssr. This will allow the html files to bind the assets correctly rather than relying on ssr.

Currently the generated html will bind paths such as:
Screen Shot 2022-05-06 at 11 18 19 AM

the update would output the paths like:

Screen Shot 2022-05-06 at 11 52 18 AM

### DIFF
--- a/vite-plugin-ssr/node/renderPage.ts
+++ b/vite-plugin-ssr/node/renderPage.ts
@@ -1078,7 +1078,7 @@ function isFileRequest(urlPathname: string) {
 
 function _parseUrl(url: string, baseUrl: string): ReturnType<typeof parseUrl> & { isPageContextRequest: boolean } {
   assert(url.startsWith('/') || url.startsWith('http'))
-  assert(baseUrl.startsWith('/'))
+  assert(baseUrl.startsWith('/') || baseUrl.startsWith('.'))
   const { urlWithoutPageContextRequestSuffix, isPageContextRequest } = handlePageContextRequestSuffix(url)
   return { ...parseUrl(urlWithoutPageContextRequestSuffix, baseUrl), isPageContextRequest }
 }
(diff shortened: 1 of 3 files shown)


brillout:
👍

I'd like to know the full extent of your motivation.

Are you integrating your static website at several Base URLs? Is that why you don't have a single static Base URL? What's your use case that needs several Base URLs?

> I wonder, should there be like a file of constants that would contain the valid characters like in a dictionary? not that these chars are to change much in the future but it's repeated by hand in multiple files (if you want me to, I can raise another PR for that, minor refactor)

Yes I agree. A function would be even better, I think.

jobarah:
> 👍
> 
> 
> 
> I'd like to know the full extent of your motivation.
> 
> 
> 
> Are you integrating your static website at several Base URLs? Is that why you don't have a single static Base URL? What's your use case that needs several Base URLs?
> 
> 
> 
> > I wonder, should there be like a file of constants that would contain the valid characters like in a dictionary? not that these chars are to change much in the future but it's repeated by hand in multiple files (if you want me to, I can raise another PR for that, minor refactor)
> 
> 
> 
> Yes I agree. A function would be even better, I think.

Ahhh, hmm so the particular scenario is: We have this site we built using vue. This will render a user profile. The plan is to use the vue app to generate a profile page per every user in our dump (around 9k users). We already have an express server running and we would like this server to be where we have our html files. Given the files will be raw html, the path for the imports must be relative, that on character missing breaks our static profiles, the patch would allow us to stick to this path in the dev process. This would basically unblock our project.

brillout:
Your PR doesn't work. I reverted it.

jobarah:
You are right, I tested this in the index file, which is the one I needed to generate, but I did another addition and I see it breaks if any other file than the index one, I'm working on it and will provide and update, this time around with more testing and attachments and a working sandbox before requesting to merge. Sorry for the hiccup

brillout:
No problems.

Looking forward to your changes 👍.

PR #329, opened 2022-05-29
Update data-fetching.page.mdx

lemredd:
minor grammar and spelling fixes.
Not sure about line 162 (adding the word "use")


brillout:
👍 It should be `then we have the option to define` which I'll commit now.

PR #343, opened 2022-06-19
Improve SolidJS Example

XiNiHa:
- Fix SolidJS example navigation rerender issue
- Adapt streaming SSR

I have tested the changes in local.


brillout:
Neat, makes sense & thanks for the PR.

PR #344, opened 2022-06-19
add pinia example with client navigation

4350pChris:


examples/vue-pinia/pages/todos/todo.page.vue
```diff
@@ -0,0 +1,22 @@
...
+
+<script setup>
+import { onServerPrefetch, computed } from 'vue'
+import { useTodos } from '../../stores/useTodos'
+
+const props = defineProps(['todoId'])
+
+onServerPrefetch(async () => {
+  const todosStore = useTodos()
+  if (todosStore.todoList.length === 0) {
```
brillout, line 15: Why is that needed?

4350pChris: You're right, it's not needed. I'll remove it.

brillout:
Just seeing you commit 👌.

Also, I wonder: I believe you mentioned that `onBeforeRender()` didn't work? I don't see why it wouldn't. But I like the `onServerPrefetch()` way. That's a neat Vue feature.

Neat PR overall.

4350pChris:
Thanks for your feedback. However, I think it is still lacking as of now.
1. There seems to be a hydration mismatch for the todo detail page.
2. A major thing that I need for the app I'm working on is access to route props (e.g. fetching only the todo that we're currently viewing) - I'll push some changes shortly that showcase that.

The reason I did not use `onBeforeRender` is that I just couldn't figure out how to access the store properly. I tried creating a new one, then passing that down via page context and updating the actual client's store but couldn't get it to work right. It also seemed rather complex, plus I'd have to create the app multiple times during rendering and I'm not sure about the performance impact that entails when having a proper web app with hundreds of components as opposed to a small showcase.

This is what I'd need:
```
<script>
export async function onBeforeRender(pageContext) {
  const todoStore = useTodoStore(store) // this is where I need the original store instance
  // logic comes here
}
</script>
```

This is kinda what I've tried:
```
<script>
export async function onBeforeRender(pageContext) {
  const { app, store } = createApp(pageContext)
  const todoStore = useTodoStore(store)
  // logic comes here
  
  // and now I have a problem - I can't just pass the app down via context
  // so I'd need to render it to e.g. a stream for SSR and pass that down
  // and for the client side I'd have to export the store's state to pass down and patch the app's state
}
</script>
```

brillout:
`1.` I've seen that and yes it's something we should fix.

See my latest commits that adds a `console.log`.

I don't know why Vue is not picking up the value even though the browser prints `todo.text Buy milk`.

`2.` I'll show you after we fix 1.

brillout:
> Found the reason

So what happened is that Vue was picking up the `pageProps` from the server which was missing `pageProps.todoId`.

brillout:
The merge was a `gh` CLI typo. I reverted the merge.

brillout:
Superseded by #345.

PR #356, opened 2022-07-08
fix apollo cache in .page.server (vue apollo example)

Amfeon:
Hello. In graphql-view example console.log outputs an error, because the server side passes an empty cache object.
This pull request fix it.


brillout:
Thanks!

PR #361, opened 2022-07-12
Docs typo: removed extra "the"

whotookzakum:
"We cannot use pre-rendering because the pre-rendered **the** HTML ..."


brillout:
Thanks!

PR #365, opened 2022-07-15
Added Vercel Build Output API example

AaronBeaudoin:


brillout:
Thanks!

I rebased, and changed the doc a little. The code looks good.

One last thing: how about we move the Vercel example code into its own repository?

Because there is no way (but correct me if I'm wrong) to have an automatic test for the Vercel deploy integration (AFAIK there is no way to preview Vercel deploys locally?).

Each example living in the vps repo is part of vite-plugin-ssr's CI.

brillout:
I externalized the example. Let me know if you object, and we can also move the example under your GitHub name if you want.

Thanks for the PR! The Vercel docs are quite in a good state now, I think. 👌

Looking forward to further PRs and bug reports (I'll have a look at your current bug reports tentatively today).

PR #368, opened 2022-07-20
Export helpers needed by vite-plugin-vercel vps integration

magne4000:
- `page-files/setup`: For other helpers to work
- `shared/route`: Need `loadPageRoutes` and `route` functions
- `shared/getPageFiles`: Need `getPageFilesAllServerSide`

Will be mergeable when `vite-plugin-vercel` compiles correctly and without hacks.

### DIFF
--- a/vite-plugin-ssr/__internal.ts
+++ b/vite-plugin-ssr/__internal.ts
@@ -1,0 +1,34 @@
+// Internal functions of vps needed by other plugins are exported via this file
+
+import { loadPageRoutes, PageRoutes, route } from './shared/route'
+import { getPageFilesAllServerSide, PageFile } from './shared/getPageFiles'
+import { getGlobalContext } from './node/globalContext'
+import { setProductionEnvVar } from './shared/setProduction'
+
+export { route, getPagesAndRoutes }
+export type { PageRoutes, PageFile }
+
+/**
+ * Used by {@link https://github.com/magne4000/vite-plugin-vercel|vite-plugin-vercel}
+ * to compute some rewrite rules and extract { isr } configs.
+ * Needs `import 'vite-plugin-ssr/__internal/setup'`
+ * @param config
+ */
+async function getPagesAndRoutes() {
+  setProductionEnvVar()
+  await getGlobalContext(true)
+
+  const { pageFilesAll, allPageIds } =
+    await getPageFilesAllServerSide(true)
+
+  const { pageRoutes } = await loadPageRoutes({
+    _pageFilesAll: pageFilesAll,
+    _allPageIds: allPageIds,
+  })
+
+  return {
+    pageRoutes,
+    pageFilesAll,
+    allPageIds,
+  }
+}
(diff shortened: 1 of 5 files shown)


magne4000:
I also need to check if I can [remove those lines](https://github.com/magne4000/vite-plugin-vercel/blob/d277bb784d56dcc0f88a8167aaff0afb562670b2/packages/vite-plugin-ssr/vite-plugin-ssr.ts#L137-L139) now or if we need to adapt `addComputedUrlProps` (I'm waiting for vite `3.0.3` to continue my tests)

brillout:
That's also an option. But then I'd export everything under a single export path `const { getPageRoutes, ... } from 'vite-plugin-ssr/__internal'`.

Instead of exporting these low-level functions, would there be a way to export a single higher-level function that is used by `prerender()` and `vite-plugin-vercel`? That'd be cleaner and easier to maintain down the road.

magne4000:
> That's also an option +1. How about we export everything under a single export path `const { getPageRoutes, ... } from 'vite-plugin-ssr/__internal'`? In order to disincentivize users from using them.

I'll do that when I have something that works :+1: 

> Instead of exporting these low-level functions, would there be a way to export a single higher-level function that is used by `prerender()` and `vite-plugin-vercel`? That'd be cleaner and easier to maintain down the road.

I need `route` function in prerender itself. All other helpers are not related to prerender but to ISR and some rewrite rules. Once I have figured out how to make everything work together, I'll propose some higher-level functions.

brillout (approved):

brillout:
Looks good 👌. Shall I merge & release?

> Once I have figured out how to make everything work together, I'll propose some higher-level functions.

👍

brillout:
> I still need those 2 lines, not sure of the proper way to fix that.

I think the high-level function will solve this.

> . I'm not sure if this function is robust enough, and I don't really know how it's suppose to work with frameworks with template files like Vue or Svelte.

Yes it should work. (The file extensions are abstracted away early; 99.99% of the vps code doesn't need to take care of that.)

> I also defined a [custom PageContext type](https://github.com/magne4000/vite-plugin-vercel/blob/671abb99f86a79acd5404981bc6b09d3f2499e1e/packages/vite-plugin-ssr/vite-plugin-ssr.ts#L39-L52). There probably is an existing one that I could re-export, but which one?

I think it's fine for now and the high-level function will tentatively solve that.

I think the current PR is good to go and let's see later what high-level function we can implement.

Let me know if you want me to take a closer look at something.

magne4000:
> Looks good 👌. Shall I merge & release?
> 
> > Once I have figured out how to make everything work together, I'll propose some higher-level functions.
> 
> 👍

Yep !

brillout:
Released in `0.4.11`. Let me know if something's missing, happy to re-release. (I don't mind high minor numbers.)

magne4000:
Not sure to totally understand how `exports` works yet...
image
(Same error when compiling)

When I had it linked it worked, so this is weird.
I have the feeling I need to add new `__internal.js` and `__internal.d.ts`, like you have for `client`, or `cli`, etc. but I don't understand why.

PR #369, opened 2022-07-20
chore(deps): update dependencies to latest to demo Apollo Client with Vite 3

alessbell:
Updating dependencies to `latest` to demo `@apollo/client` usage with `vite` v3. After updating, the only deprecation warning was the following:

CleanShot 2022-07-20 at 11 53 15@2x

Making that prescribed change to the options passed to `vite.createServer` eliminated the warning. The demo app runs as expected and behaves as it did when previously built with `vite` v2:

CleanShot 2022-07-20 at 11 52 59@2x

### DIFF
--- a/examples/graphql-apollo-react/package.json
+++ b/examples/graphql-apollo-react/package.json
@@ -3,14 +3,14 @@
     "dev": "node ./server"
   },
   "dependencies": {
-    "@apollo/client": "3.5.9",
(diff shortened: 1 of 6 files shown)


alessbell:
Closed and re-opened the PR to kick off the tests again, sorry for the noise :)

alessbell:
In my final commit I also updated the vue example + test so it doesn't skip on v3. Looks like we're all green \o/

PR #371, opened 2022-07-22
Some changes to make vite-ecosystem-ci pass

sapphi-red:
This PR applies some change to make vite-ecosystem-ci pass.

## bump pnpm version (https://github.com/brillout/vite-plugin-ssr/commit/719d1d3fba063c82ef1c729cab11fe0e2bce0487)
*This is not related to vite-ecosystem-ci.*

I think this `packageManager` field is outdated. The lockfile uses lockfileVersion 5.4 but [this is supported from pnpm 7.0.0](https://github.com/pnpm/pnpm/releases/tag/v7.0.0).
https://github.com/brillout/vite-plugin-ssr/blob/fe4ff6af483d03346eeef34904c6f3fca9dec161/pnpm-lock.yaml#L1

So I upgraded to 7.6.0.

## set react-dom as peerDeps of @apollo/client (https://github.com/brillout/vite-plugin-ssr/commit/bc9591ddf36b893f0c8a462b1c4ef243b5a83290)

`examples/graphql-apollo-react` was failing.
https://github.com/vitejs/vite-ecosystem-ci/runs/7462644762?check_suite_focus=true#step:7:478

This was because `require('react-dom/server')` inside `@apollo/client` was resolved to react 18.1.0 but react 18.2.0 was used by Vite.
I added `react-dom` to peerDep of `@apollo/client` by `pnpm.packageExtensions`. Maybe this should be fixed on package side? I'm not sure.

## fix type error in examples/layouts-react/.testRun.ts (https://github.com/brillout/vite-plugin-ssr/commit/f7257e3beec85722478d0b2f326410446ffbcfc1)

`examples/layouts-react` had a type error.
https://github.com/vitejs/vite-ecosystem-ci/runs/7462644762?check_suite_focus=true#step:7:591

I simply fixed it.
BTW is there any `tsconfig.json` which covers files under examples directory? IIUC if there isn't any, [the tsconfig in vite-ecosystem-ci](https://github.com/vitejs/vite-ecosystem-ci/blob/main/tsconfig.json) will be applied.
Maybe the tsconfig in ecosystem-ci should exclude files in `workspace` directory so that it won't be applied to files which are cloned?

## set ssr.noExternal to graphql-apollo-vue (https://github.com/brillout/vite-plugin-ssr/commit/e7238e19994437ac31dda009e830c6197c00a90e)

`examples/graphql-apollo-vue` was failing.
https://github.com/vitejs/vite-ecosystem-ci/runs/7462644762?check_suite_focus=true#step:7:751

When building client, `@apollo/client`, `@vue/apollo-composable` both resolves to `module` field.
When building SSR, these packages are externalized. But these packages does not have ESM entries, and these CJS entries are not compatible with [node.js's automatic named export support](https://nodejs.org/api/esm.html#interoperability-with-commonjs:~:text=For%20better%20compatibility%20with%20existing%20usage%20in%20the%20JS%20ecosystem%2C%20Node.js%20in%20addition%20attempts%20to%20determine%20the%20CommonJS%20named%20exports%20of%20every%20imported%20CommonJS%20module%20to%20provide%20them%20as%20separate%20ES%20module%20exports%20using%20a%20static%20analysis%20process.).
So these packages needs to be bundled. I added them to `ssr.noExternal`.

Also I seems something strange is happening around package overrides.
When you run `pnpm build` in `examples/graphql-apollo-vue`, the output is:
```text
vite v3.0.2 building for production...
// omit
vite v2.9.14 building SSR bundle for production...
// omit
```
So Vite 3 is used for client build and Vite 2 is used for SSR build. This maybe the reason why the CI is passing in this repository but not passing in ecosystem-ci.

## (**Not included in this PR**) preact
Preact now supports Vite 3, so this code might be able to be removed after upgrading `@preact/preset-vite` to `2.3.0`.
https://github.com/brillout/vite-plugin-ssr/blob/fe4ff6af483d03346eeef34904c6f3fca9dec161/boilerplates/.testRun.ts#L38-L44

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,7 @@ jobs:
       - name: Run tests
         run: ${{ matrix.testCmd }}
         env:
+          GIT_REPOSITORY: ${{ github.repository }}
           GIT_BRANCH: ${{ github.ref_name }}
           CF_ACCOUNT_ID: ${{ secrets.CF_ACCOUNT_ID }}
           CF_API_TOKEN: ${{ secrets.CF_API_TOKEN }}
--- a/boilerplates/.testRun.ts
+++ b/boilerplates/.testRun.ts
@@ -12,7 +12,6 @@ import {
   editFileRevert,
 } from '../libframe/test/setup'
 import assert from 'assert'
-const viteVersion = '3.?.?'
 
 function testRun(
   cmd: 'npm run dev' | 'npm run prod' | 'npm run preview',
@@ -34,14 +33,6 @@ function testRun(
 
   const isProd = cmd === 'npm run prod' || cmd === 'npm run preview'
   const isDev = !isProd
-
-  if (uiFramewok === 'preact' && isProd && viteVersion.startsWith('3')) {
-    // https://github.com/preactjs/preact/issues/3558
-    const msg = 'SKIPPED preact prod until it supports Vite 3.'
-    console.log(msg)
-    test(msg, () => {})
-    return
-  }
 
   test('page content is rendered to HTML', async () => {
     const html = await fetchHtml('/')
--- a/examples/cloudflare-workers-react-full/.testRun.ts
+++ b/examples/cloudflare-workers-react-full/.testRun.ts
@@ -10,25 +10,26 @@ import { page, run, autoRetry, fetchHtml, isGithubAction, urlBase } from '../../
 function testRun(cmd: 'npm run dev' | 'npm run preview', { hasStarWarsPage }: { hasStarWarsPage: boolean }) {
   const isWrangler = cmd === 'npm run preview'
 
-  if (isWrangler) {
-    if (isGithubAction() && process.env['GIT_BRANCH'] !== 'main') {
-      // Cloudflare tokens not available in Pull Requests:
-      //  - GitHub Actions doesn't make secrets available to Pull Requests.
-      //    - https://github.community/t/feature-request-allow-secrets-in-approved-external-pull-requests/18071/4
-      const envVars = Object.keys(process.env)
-      expect(envVars).not.toContain('CF_ACCOUNT_ID')
-      expect(envVars).not.toContain('CF_API_TOKEN')
-      const msg = 'SKIPPED: wrangler tests cannot be run in Pull Requests.'
-      console.log(msg)
-      test(msg, () => {})
-      return
+  if (isGithubAction()) {
+    const repository = process.env['GIT_REPOSITORY']
+    expect(repository).toBeTruthy()
+    // GitHub Actions doesn't make secrets available to Pull Requests.
+    // - https://github.community/t/feature-request-allow-secrets-in-approved-external-pull-requests/18071/4
+    if (!process.env['CF_ACCOUNT_ID']) {
+      expect(repository).not.toBe('brillout/vite-plugin-ssr')
+      expect(process.env['CF_ACCOUNT_ID']).toBeFalsy()
+      expect(process.env['CF_API_TOKEN']).toBeFalsy()
+      if (isWrangler) {
+        const msg = 'SKIPPED: wrangler tests cannot be run in Pull Requests.'
+        console.log(msg)
+        test(msg, () => {})
+        return
+      }
+    } else {
+      expect(repository).toBe('brillout/vite-plugin-ssr')
+      expect(process.env['CF_ACCOUNT_ID']).toBeTruthy()
+      expect(process.env['CF_API_TOKEN']).toBeTruthy()
     }
-  }
-
-  if (isWrangler) {
-    const envVars = Object.keys(process.env)
-    expect(envVars).toContain('CF_ACCOUNT_ID')
-    expect(envVars).toContain('CF_API_TOKEN')
   }
 
   {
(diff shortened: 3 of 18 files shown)


brillout:
I'm seeing that main is now failing, probably a minor thing I'll fix tomorrow. I'll reply to the rest tmrw as well. In the meantime: thanks for the PR!

brillout:
Alright, both the branch `main` (mostly Vite 2) and the [branch `vite3`](https://github.com/brillout/vite-plugin-ssr/tree/vite3) (always Vite 3 with `pnpm.overrides`) are green.


> I added `react-dom` to peerDep of `@apollo/client` by `pnpm.packageExtensions`. Maybe this should be fixed on package side? I'm not sure.

I think so too: [github:apollographql/apollo-client - PR #9933 - file changes](https://github.com/apollographql/apollo-client/pull/9933/files).

> `@apollo/client`, `@vue/apollo-composable` [...] needs to be bundled. I added them to `ssr.noExternal`.

I can confirm that `ssr.noExternal` is needed. I tried to find a way to circumvent using `ssr.noExternal` but I failed and I don't think it's possible.

In care you are curious:
 - Reproduction: https://github.com/brillout/apollo-vite3-bug
 - Bug Report: https://github.com/apollographql/apollo-client/issues/9938

Hopefully, Apollo fixes these issues soon. I don't think that using `ssr.noExternal` is something the user should have to do.

> Also I seems something strange is happening around package overrides. When you run `pnpm build` in `examples/graphql-apollo-vue`, the output is:

That's probably because vite-plugin-ssr automatically runs `$ vite build --ssr` on behalf of the user: https://github.com/brillout/vite-plugin-ssr/blob/55eb285f1b19baa39abe089687d0d9b90461701b/vite-plugin-ssr/node/plugin/plugins/autoFullBuild.ts#L39

(So that the user can simply run `$ vite build` instead of having to run `$ vite bulid && vite build --ssr`.)

This means that the first `vite` version comes from user land whereas the second `vite` version comes from this repository's vite-plugin-ssr's `vite` `devDependency`.

I'm surprised though because I believe that vite-ecocystem-ci uses `pnpm.overrides.vite`, so I'm not sure why there can be two different installed versions then.

> Preact now supports Vite 3

Neat. I re-enabled Preact's test suite. I also re-enabled urql's test suite which now also works for Vite 3. Everything is green and no test suites are being skipped anymore. :tada:

## Why did `vite-ecocystem-ci` failed when `vite3` was green?

I wonder why that is since I was under the impression that the only thing vite-ecocystem-ci does is to use `pnpm.overrides`, which is exactly what the `vite3` branch does. There are several instances where `vite3` was green whereas vite-ecocystem-ci failed.

So it seems that vite-ecocystem-ci does some other things, which I'm not sure it's something we want since it will lead to other PRs like this one. (I did regularly rebased `vite3` on top of `main` to ensure that vite-ecocystem-ci stays green.)

sapphi-red:
> I'm surprised though because I believe that vite-ecocystem-ci uses `pnpm.overrides.vite`, so I'm not sure why there can be two different installed versions then.

I meant this repository uses two different versions and the vite-ecosystem-ci uses the same version. I now understand it's because of the `devDependency` of vite-plugins-ssr.

> I re-enabled Preact's test suite. I also re-enabled urql's test suite which now also works for Vite 3. Everything is green and no test suites are being skipped anymore. 🎉

:tada:

> ## Why did vite-ecocystem-ci failed when vite3 was green?

I found that vite-ecosystem-ci uses [`ni`](https://github.com/antfu/ni) to detect the package manager. [`ni` uses `packageManager` field](https://github.com/antfu/ni/blob/fe03357ab31636489ba6b44f0bfb6a6ab971e5e2/src/detect.ts#L27-L44).
So ecosystem-ci used pnpm@6 previously.
But this repository [explicitly installs pnpm@7](https://github.com/brillout/vite-plugin-ssr/blob/367a5b2a836cf43d1187074f4c1900c0abcd0c0e/.github/workflows/ci.yml#L43-L46). Since this repository has `packageManager` field, it might be better to remove `with.version` here, because [`packageManager` field will be used if it is omitted](https://github.com/pnpm/action-setup#version).

BTW I think this needs to be `GITHUB_REPOSITORY` instead of `GIT_REPOSITORY`. (I'm not sure why it isn't failing in this repo.)
https://github.com/sapphi-red/vite-ecosystem-ci/runs/7487536712?check_suite_focus=true#step:7:443

brillout:
> it might be better to remove `with.version` here

👍 Done: https://github.com/brillout/vite-plugin-ssr/commit/b8060ac5222368fe085e16db942015efe2dcc3c2.


> BTW I think this needs to be `GITHUB_REPOSITORY` instead of `GIT_REPOSITORY`. (I'm not sure why it isn't failing in this repo.) https://github.com/sapphi-red/vite-ecosystem-ci/runs/7487536712?check_suite_focus=true#step:7:443

It's set here: https://github.com/brillout/vite-plugin-ssr/blob/367a5b2a836cf43d1187074f4c1900c0abcd0c0e/.github/workflows/ci.yml#L68

Is there a way to detect whether the test is being run in vite-ecosystem-ci? I can then change the code accordingly.

sapphi-red:
> It's set here:
> Is there a way to detect whether the test is being run in vite-ecosystem-ci? I can then change the code accordingly.

I see. I think just replacing `GIT_REPOSITORY` with `GITHUB_REPOSITORY` will work. `GITHUB_REPOSITORY` is automatically defined by GitHub Actions.
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

brillout:
Done https://github.com/brillout/vite-plugin-ssr/commit/9e3ad05975c693ce555a55aef71618cf84d5bf02. I also think it should work now 👍.

PR #384, opened 2022-07-28
chore: add sourcemap for debug

hemengke1997:
I'm learning the source code for VPS, and I found local debugging inconvenient, so I did these things:
1. add typescript sourcemap, and ignore `.map` from `npm publish`
2. replace `rimraf` for `rm -rf` as windows do not support it.

Sorry for my bad english. If I did something wrong, please tell me!


brillout:
The source map thing is a good idea!

brillout:
FYI I had to revert this because of https://github.com/brillout/vite-plugin-ssr/issues/183.

PR #387, opened 2022-07-28
fix: make onPagePrerender pageContext concurrent safe

magne4000:
While working on `vite-plugin-vercel` I noticed that prerender was only generating `.pageContext.json` files, twice.

Same `pageContext` is used for a page `.html` and `.pageContext.json`, resulting in `_prerenderResult` being overwritten in parallel runs, before `onPagePrerender` finishes its execution.

`objectAssign` can also now take any number of args, like `Object.assign` would.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,6 @@ env:
 on:
   push:
   pull_request:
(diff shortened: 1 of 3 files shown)


brillout:
Thanks!

PR #395, opened 2022-07-30
add support for server-side rendered preprocessed stylesheets

justablob:
This PR adds support for importing preprocessed stylesheets (Sass and similar) from server-only pages.
The regex used for checking these is taken directly from Vite's source code.


brillout:
Neat!

brillout:
I'll release this after my bug fixing sprint, ETA being today.

brillout:
FYI https://github.com/brillout/vite-plugin-ssr/commit/84ea94087ee0b744ee880a0f9d37777d172766f7 but I believe it should still work.

PR #410, opened 2022-08-05
Fix minor documentProps error

arthurgailes:
[Line 24](https://github.com/brillout/vite-plugin-ssr/blob/158059610a9a76611b5f43576dcee6b1e8ab0f92/boilerplates/boilerplate-react-ts/renderer/_default.page.server.tsx#L24): 

```
const { documentProps } = pageContext;
const title = (documentProps && documentProps.title) || 'Vite SSR app';
```

I believe this should be `const { documentProps } = pageContext.exports;`

Which also needs to be resolved downstream in ./types.ts


brillout:
Indeed.

Thanks for the PR!

PR #430, opened 2022-08-28
add solid js spa example with tailwind and shoelace

samuelstroschein:
Took me some time to get solidjs in spa mode working. I thought it might benefit other people.


brillout:
That does indeed benefit the community 👌.

How about we publish this example as a repository under your name and I link it to the docs and to https://github.com/brillout/vite-plugin-ssr/tree/main/examples#community-examples? I think it's better if we keep the `examples/` directory focused on single integrations.

samuelstroschein:
Oh, I didn't see the community examples. 

We can either: 

a) Host it under my name 
b) Take out Tailwind and/or Shoelace and put it in this repo. 

The reasoning for b) is simpler renderer files. The [Solid example](https://github.com/brillout/vite-plugin-ssr/blob/main/examples/solid/renderer/_default.page.server.tsx) is a bit overwhelming and requires reading through the documentation. For example, what is `pipe`? For beginners, hydrate in the client renderer might also open questions. 

This SPA example is straightforward. The server sends HTML with a root div `<div id="root">` and the client renders the app. It shows the strength of vite-plugin-ssr. Developers can control what is happening. That said, routing is not implemented yet.

brillout:
That's a good point.

How about both a) and b)? Under your name we keep a version with TailwindCSS & Shoelace, and under `examples/` we keep a simple version.

samuelstroschein:
Hm. Considering that I haven't looked for vite-plugin-ssr plugins outside of this repository and didn't even see the community packages at the bottom, I don't think a) is of much value. 

I refactored the example and removed tailwind and shoelace.

brillout:
> Hm. Considering that I haven't looked for vite-plugin-ssr plugins outside of this repository and didn't even see the community packages at the bottom, I don't think a) is of much value.

I'd also link it from the docs. As you wish.

examples/solid-spa-tailwind-shoelace/renderer/types.ts
```diff
@@ -1,12 +1,9 @@
+import type { PageContextBuiltInClient } from "vite-plugin-ssr/client";
 import type { PageContextBuiltIn } from "vite-plugin-ssr";
-import { Component } from "solid-js";
-
-/**
- * PageContext used by the renderer.
- *
- * vite-plugin-ssr is framework agnostic. Hence, `Page` is defined as `any`.
```
samuelstroschein: Interesting. Why did you take out that comment? 

Using `Omit` is difficult to understand without this comment, don't you think?

samuelstroschein:
> I'd also link it from the docs. As you wish. 

I prefer to have the example in this repository.

examples/solid-spa-tailwind-shoelace/renderer/types.ts
brillout: I agree, see https://github.com/brillout/vite-plugin-ssr/commit/4e3ac790f165c73c6189630cef08b04129f4905a and https://github.com/brillout/vite-plugin-ssr/commit/ce0db176f99864e99697198ef1ecaf6da7308c2d.

samuelstroschein: It might be better to define Page as `unknown` instead of `any` to increase type safety, see https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type. 

However, the docs should then probably clarify that Page is to be set by the developer.

brillout: I agree from a type safety POV.

But, from a DX POV, I prefer `pageContext.Page: any`: I don't want to force users to have to define `Page` while `Page` type safety isn't really worth it.

PR #440, opened 2022-09-04
fix: make streaming SSR work on Firefox

XiNiHa:
This PR fixes streaming SSR on Firefox.

Firefox streams the HTML content only if `Content-Type` has `charset`, therefore the previous version didn't do streaming SSR on Firefox since `contentType` was `text/html` which doesn't include `charset`.


brillout:
Wonderful. Released in `0.4.31`.

PR #450, opened 2022-09-14
fix: #449

### LINKED ISSUE #449: vps injects css into the page, even though it's disabled
### Description

vite-plugin-ssr injects css assets marked with ?inline, even though it shouldn't.
https://vitejs.dev/guide/features.html#disabling-css-injection-into-the-page

For context: Angular compiles the stylesheets server-side, makes them component scoped, and inlines them into the html response. I use ?inline in [@nitedani/vite-plugin-angular](https://github.com/nitedani/vite-plugin-angular), in a compilation step.

Given the scss file:
```css
div {
    color: red
}
```
Running in dev mode, the response looks like this:

```html
<html lang="en">
    <head>
        <link rel="stylesheet" type="text/css" href="/pages/about/index.page.scss?inline?direct">
...
    </head>
    <body>
        <div id="page-view">
            <html>
                <head>
                    <style ng-transition="ng-component">
                        div[_ngcontent-sc14] {
                            color: red;
                        }
                    </style>
                </head>
                <body>
                    <ng-component>
                        <div>
                             <div>This should be white</div>
                            <ng-component _nghost-sc14="">
                                <div _ngcontent-sc14="">This should be red</div>
                            </ng-component>
...
```

The result is, that all text turns red on the page, once the stylesheet finished downloading, because the style is applied globally.
I expect css files marked with ?inline to not be injected automatically

### Error Stack

_No response_

nitedani:
Fixes #449

### DIFF
--- a/vite-plugin-ssr/node/renderPage/getPageAssets.ts
+++ b/vite-plugin-ssr/node/renderPage/getPageAssets.ts
@@ -52,6 +52,10 @@ async function getPageAssets(
     const { mediaType = null, preloadType = null } = inferMediaType(src) || {}
     const assetType = mediaType === 'text/css' ? 'style' : 'preload'
     if (isDev && mediaType === 'text/css') {
+      // https://github.com/brillout/vite-plugin-ssr/issues/449
+      if (src.endsWith('?inline')) {
+        return;
+      }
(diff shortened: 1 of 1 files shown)


brillout:
Lovely. Released in `0.4.34`.

PR #458, opened 2022-09-17
an example for vue&router&vuex

atenxyz:
todo: use VuexPersistence in vite-plugin-ssr


brillout:
👍.

In general, the `examples/*` are focused on one integration at a time.

Se we can either change this PR to focus on only one integration or we can move the example to your GitHub repository and add links to it in the docs and in `examples/README.md`.

(P.S. thanks for the donation :-).)

PR #471, opened 2022-10-12
Update getPageContextSerializedInHtml.ts

AdvTechnoKing:
Selector only in scripts.
https://github.com/brillout/vite-plugin-ssr/blob/637cb877ccc2bb64395318f9d99674ed3e8dfed2/vite-plugin-ssr/node/html/injectAssets.ts#L209

Can I Use? Yes!
https://caniuse.com/mdn-api_htmlselectelement_nameditem

Working? Yes!
image


brillout:
I don't think we need this.

PR #473, opened 2022-10-13
Update Vercel documentation.

leerob:
Hey there! 👋 

Thanks again for this amazing library. Wanted to update the Vercel docs to clarify a few things. First, we should probably recommend deploying with `vite-plugin-vercel` to enable support for all Vercel features. This is no longer a fast-moving target (sorry for the changes between v1 and v3) and has now stabilized. The Build Output API is what all of the meta-frameworks on Vercel and moving towards targeting, including SvelteKit, Nuxt, Astro, and more.

Learn more: https://vercel.com/blog/build-output-api

### DIFF
--- a/docs/pages/vercel.page.server.mdx
+++ b/docs/pages/vercel.page.server.mdx
@@ -1,36 +1,16 @@
 import { Link, Note } from '@brillout/docpress'
 
-Using a Vercel API Route is the recommended way for deploying `vite-plugin-ssr` to [Vercel](https://vercel.com/).
-
-There are also alternatives, in particular [vite-plugin-vercel](#vite-plugin-vercel) if we want partial static deploy and/or [ISR](#isr).
-
-
-## API Route
-
-We can deploy our `vite-plugin-ssr` app to Vercel simply by creating a Vercel API Route `api/ssr.js` that server-side renders our app.
-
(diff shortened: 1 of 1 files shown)


leerob:
Also - would be happy to sponsor y'all with a free open-source plan on Vercel. Say the word and I'll help get it set up!

brillout:
Thanks for the PR!

Knowing my users, I think a lot of them actually do like to directly use Vercel APIs instead of using a middle-man such as `vite-plugin-vercel`. Don't take me wrong: `vite-plugin-vercel` is great and I like it a lot. It's just about documenting the full range of options to vite-plugin-ssr users. A lot of them actually directly use the Build Output API &mdash; that's the kind of users vite-plugin-ssr has. So I accommodated the PR for these users. But feel free to let me know if you disagree with anything. Merging in the meantime.

> Also - would be happy to sponsor y'all with a free open-source plan on Vercel. Say the word and I'll help get it set up!

Neat, thanks. The free Vercel plan works for me so far but I'll let you know if I need one 👍.

I'm actually thinking: Vercel is sponsoring projects like Astro and SvelteKit. What does it take to become a Vercel sponsor? I'm thinking maybe @magne4000 being the author of vite-plugin-vercel is eligible.

leerob:
Oh I didn't realize you were using Vercel since https://vite-plugin-ssr.com/ wasn't on Vercel here 😄

PR #474, opened 2022-10-14
Improve getTagClosing performance

bagr001:
Hello @brillout! Thank you for your awesome work on this plugin. I love it's flexibility and ease of use.

I am building quite a large SSR project around it and seeking the best possible performance. I've noticed, there is a non-negligible framework overhead. So I started profiling the costly functions calls. Surprisingly, all the expensive operations are located in `injectHtmlSnippet.ts`. There was very inefficient `getTagClosing` using `String.includes` and lowercasing all html... I refactored this function using `RegExp` as in `tagClosingExists`. The result is the whole `injectAssetsToStream` operation is approx. 3.5x faster.

BEFORE:
default

AFTER:
Screenshot_11

I have also noticed a lot of `assert()` to heavy check results during runtime. However, those runtime "double" or sometimes "tripple" checks where the regexp has to be evaluated on large HTML multiple times is costly.

I can imagine more performance gains when rewriting `injectAssetsToStream` fewer checks, resulting in fewer regex evaluations and faster execution.

Thanks!

### DIFF
--- a/vite-plugin-ssr/node/html/injectAssets/injectHtmlSnippet.ts
+++ b/vite-plugin-ssr/node/html/injectAssets/injectHtmlSnippet.ts
@@ -98,13 +98,17 @@ function injectAtClosingTag(tag: 'head' | 'body' | 'html', htmlString: string, h
   assert(tagClosingExists(tag, htmlString))
 
   const tagClosing = getTagClosing(tag)
-  const htmlParts = htmlString.split(tagClosing)
+  const matches = htmlString.match(tagClosing)
+  assert(matches && matches.length >= 1)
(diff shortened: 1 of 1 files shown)


brillout:
I'm glad you like it :-) & nice PR 👌.

I think we should add some comments so that future changes don't add performance regressions.

> I can imagine more performance gains when rewriting `injectAssetsToStream` fewer checks, resulting in fewer regex evaluations and faster execution.

If you're up to it, a PR would be much welcome.

brillout:
Released in `0.4.41`.

Neat PR 👍. Further PRs like this are much welcome.

Issue #478, opened 2022-10-18
`npm init vite-plugin-ssr@latest` logs "fatal" git errors. 


wilcoxmd:
### Description

When scaffolding a new `vite-plugin-ssr` project with `npm init vite-plugin-ssr@latest`, the following git errors are logged in the command's output: 

```bash
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
error: No such remote 'origin'
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

This doesn't seem to cause any actual issue with git initialization or the first commit.  Rather, this seems to be a normal git behavior when committing for the first time in a fresh repository. The error logs on the first commit but then won't log again on subsequent commits. 

However, this behavior is a little noisy in the scaffolding logs, and made me initially wonder if something went wrong with the operation.

As a solution, we could simply prevent the log from making it to the user by passing `{ stdio: 'ignore' }` in the git-related `execSync` options. As far as I can tell, create-react-app [uses](https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/react-scripts/scripts/init.js#L59-L82) the same approach to prevent these errors from making it to users. 

I'm not sure of another option offhand; there might be something better. If ignoring the output here sounds like an OK path forward to you as well though, I'd be happy to submit a PR with that change. 

### Error Stack

Included in description

brillout:
I don't think this is related to vps.

wilcoxmd:
This error is created when the boilerplate creation script [creates an initial commit](https://github.com/brillout/vite-plugin-ssr/blob/fafed66a1833ec9cb8d7fbb63183b20e7521c5c3/boilerplates/index.js#L185-L195).

This error happens if you're doing bare git operations as well, but other similar tools (create-react-app) just hide the error log for a nicer scaffolding UX.

brillout:
Ok I see. Yes, PR welcome! Good idea.

wilcoxmd:
Great, I'll put something together!

brillout:
Looking forward to it! Apologies for the hasty closing reaction - it was getting late :-).

brillout:
Done.

PR #480, opened 2022-10-19
chore: ignore git logs during new app scaffolding

wilcoxmd:
Previously, running `npm init vite-plugin-ssr@latest` would include a few logs from git that were cluttering the scaffolder's output. This change prevents git from logging to the user's terminal when initializing the new git repository and creating the first commit. 

See #478 for full details.

### DIFF
--- a/boilerplates/index.js
+++ b/boilerplates/index.js
@@ -180,8 +180,8 @@ function initGitRepo(cwd) {
     return
   }
 
-  execSync('git init', { cwd })
-  execSync('git add .', { cwd })
+  execSync('git init', { cwd, stdio: 'ignore' })
+  execSync('git add .', { cwd, stdio: 'ignore' })
   execSync(
     [
       'git',
@@ -191,7 +191,7 @@ function initGitRepo(cwd) {
       '--no-gpg-sign',
       '--message="boilerplate Vite w/ vite-plugin-ssr"'
     ].join(' '),
-    { cwd }
+    { cwd, stdio: 'ignore' }
   )
 }
 function isGitInstalled() {
--- a/boilerplates/index.js
+++ b/boilerplates/index.js
@@ -180,20 +180,35 @@ function initGitRepo(cwd) {
     return
   }
 
-  execSync('git init', { cwd, stdio: 'ignore' })
-  execSync('git add .', { cwd, stdio: 'ignore' })
-  execSync(
-    [
-      'git',
-      '-c user.name="Romuald Brillout"',
-      '-c user.email="vite-plugin-ssr@brillout.com"',
-      'commit',
-      '--no-gpg-sign',
-      '--message="boilerplate Vite w/ vite-plugin-ssr"'
-    ].join(' '),
-    { cwd, stdio: 'ignore' }
-  )
+  let isInitialized = false
+  try {
+    // Attempt to create a git repository.
+    // We hide standard git-related logs from the user, but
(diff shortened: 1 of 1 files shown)


brillout:
1. I think we should add a comment
2. I think we should whitelist the error instead of ignoring all errors 

Or, optionally, we can wrap the whole git command calls in a `try {} catch {}` and swallow the error while showing a warning that git couldn't have been initialized.

wilcoxmd:
> I think we should add a comment

Yep, agreed. 👍  I can add one.

> whitelist the error

I'm not quite sure what you mean by this. Is the suggestion to just ignore the log on the commit command? Or something else?

I'm not actually sure if there is an error to catch on this, or if git is just logging. I can take a look more closely at that, but my hunch is that it is just a log since the commit command successfully creates a commit.

brillout:
Right now, the PR ignores all `stderr` logs. Instead how about we only ignore `fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.`?

wilcoxmd:
I played with this a bit more, and trying to blocklist specific errors from logging ends up feeling a bit hacky. In this particular case, git is logging an error to `stderr`, but the process still succeeds so there is not anything to catch. In order to do the blocklist behavior, there were a few options I found:

1. We could `exec` instead of `execSync` in order to capture the `stderr` output, but that just returns a single string that we need to do some replacing on. Replacing and formatting the errors with newlines and such is a bit awkward.
2. We could use `spawn` to run the task, and process individual errors as they come in a `stderr` callback. That would involve reworking the existing code a bit more, which feels a bit overkill in this case. 

As a middle ground, I've kept all git initialization commands as `execSync` calls , but wrapped them in `try/catch`. In the event that one of those commands does actually fail and exit, `execSync` should still throw and we will catch and log the message. If the git initialization process fails, we can also remove the possibly-bad `.git` directory.

This route feels the most straightforward to me, and also aligns with how other similar tools are handling this type of process. Here's the relevant code from [create-react-app](https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/react-scripts/scripts/init.js#L59-L82) and [create-next-app](https://github.com/vercel/next.js/blob/a5d53155cddb41fec1901955bb660fa6c1d4b4cf/packages/create-next-app/helpers/git.ts#L22-L48))

Let me know if you have any other thoughts!

boilerplates/index.js
```diff
@@ -180,20 +180,35 @@ function initGitRepo(cwd) {
...
+        '-c user.email="vite-plugin-ssr@brillout.com"',
+        'commit',
+        '--no-gpg-sign',
+        '--message="boilerplate Vite w/ vite-plugin-ssr"'
+      ].join(' '),
+      { cwd, stdio: 'ignore' }
+    )
+  } catch (error) {
+    if (isInitialized) {
+      fs.rmSync(path.join(cwd, '.git'), { recursive: true, force: true })
```
wilcoxmd, line 205: I noticed a note above on L168 about Node 12 support, but it doesn't seem like that is covered any more in test cases? Is it safe to use rmSync now?

brillout: Yea it's fine.

brillout:
FYI https://github.com/brillout/vite-plugin-ssr/commit/3e7750bab5e22fc00e4d88ff4291962f753733ea.

PR #488, opened 2022-10-26
Add Dynamic import example

### LINKED ISSUE #487: Error in Dynamic Import example
### Description

When using the example on dynamic loading with a React.Suspense component, I receive the following error: `Uncaught Error: Hydration failed because the initial UI does not match what was rendered on the server.` 

The page still works, but I'm assuming I'm losing some SSR benefit here. I believe this is resulting from the components in node and the browser being different - even when they call the same underlying function. As an example, the code below produces the same error, adapted from the example:
```

// pages/pick-location.page.jsx

import React from 'react'
const Loading = () => <div>Loading...</div>;

export function Page() {
  return <>
    Please pick a location.
    <Map />
  </>
}

function Map() {
  const isBrowser = typeof window !== 'undefined'
  const isNodejs = !isBrowser

  // If I delete this section, the error disappears
  if (isNodejs) {
    return <Loading />
  }

  // Removing the lazy load to illustrate the problem
  return (
    <React.Suspense fallback={<Loading />}>
      <Loading />
    </React.Suspense>
  )
}
```

### Error Stack

```shell
react-dom.development.js:19849 Uncaught Error: There was an error while hydrating. Because the error happened outside of a Suspense boundary, the entire root will switch to client rendering.
    at updateHostRoot (react-dom.development.js:19849:57)
    at beginWork (react-dom.development.js:21615:14)
    at beginWork$1 (react-dom.development.js:27426:14)
    at performUnitOfWork (react-dom.development.js:26557:12)
    at workLoopSync (react-dom.development.js:26466:5)
    at renderRootSync (react-dom.development.js:26434:7)
    at recoverFromConcurrentError (react-dom.development.js:25850:20)
    at performConcurrentWorkOnRoot (react-dom.development.js:25750:22)
    at workLoop (scheduler.development.js:266:34)
    at flushWork (scheduler.development.js:239:14)
```

arthurgailes:
A working example submitted for inclusion here: https://vite-plugin-ssr.com/dynamic-import 

resolves #487

### DIFF
--- a/docs/pages/dynamic-import.page.server.mdx
+++ b/docs/pages/dynamic-import.page.server.mdx
@@ -3,6 +3,7 @@ We can use `import()` to lazily load components.
 ```diff
 - import { SomeComponent } from 'some-component'
 ```
+
 ```js
 // Somewhere in our code
 function someFunction() {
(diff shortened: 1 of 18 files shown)


brillout:
How about we move the example to `github.com/arthurgailes`? Also I'd add a readme then explaining how it works.

arthurgailes:
Understandable, feel free to close.

brillout:
We should still add a link in the docs though. An example is very much valuable (it's actually been on my to-do list).

Thanks for contributing an example!

brillout:
@arthurgailes Did you already move the example to you GitHub handle?

arthurgailes:
I've added it here: https://github.com/arthurgailes/vite-plugin-ssr_dynamic-import-example

brillout:
Wonderful. Thanks.

PR #500, opened 2022-11-06
changed the order of the e.g.  and added a Number

jannesblobel:
To me, the "routing precedence" appears a bit confusing. 

because the order of the examples is not correct.
you mention " Example (4) + (6) + (7):" 
but I think the order of the following examples is incorrect.   

currently the Example with the number "6" is in the first position and "4" in the second. 

I changed the order and added a number to the examples. 

Examples: 

4 =
 ```js
// product/list.page.route.js
export default '/product'
```

6:
```js
// product/item.page.route.js
export default '/product/@productId'
```
7:
/ product/catch-all.page.route.js
.......


brillout:
Makes sense. Thanks!

PR #503, opened 2022-11-06
provide info for redirect

samuelstroschein:
It took me some time to realize that vite-plugin-ssr provides no built-in redirect functionality. Makes sense but should probably be mentioned in the docs.


brillout:
That makes sense indeed. Thanks!

PR #504, opened 2022-11-06
fix: type export

### LINKED ISSUE #501: navigate() wrong declaration file
### Description

The declaration file of `navigate()` is incorrect. TypeScript lints that navigate expects 0 arguments.

image

The declaration file in dist reveals why. Navigate is exported but has no arguments. 

```ts
// dist/cjs/node/client/router.d.ts
export { navigate };
declare function navigate(): void;
```

The implementation takes multiple arguments. 

```ts
// vite-plugin-ssr/client/router/useClientRouter.ts
async function navigate(
  url: string,
  { keepScrollPosition = false, overwriteLastHistoryEntry = false } = {}
): Promise<void>
```

I was not able to figure out why typescript creates a wrong declaration file for `navigate`, and potentially other declarations too.

### Error Stack

```shell
version 0.44.4
```

samuelstroschein:
fixes #501

Read more here https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#packagejson-exports-imports-and-self-referencing

### DIFF
--- a/vite-plugin-ssr/node/client/router.ts
+++ b/vite-plugin-ssr/node/client/router.ts
@@ -1,10 +1,11 @@
 export { navigate }
 
 import { assertWarning } from '../../utils/assert'
+import type { navigate as navigateOriginal } from '../../client/router'
 
-function navigate() {
+const navigate = (() => {
   assertWarning(false, 'Calling navigate() on the server-side has no effect', {
     showStackTrace: true,
     onlyOnce: false
   })
-}
+}) as any as typeof navigateOriginal
--- a/examples/react-full/package.json
+++ b/examples/react-full/package.json
@@ -17,7 +17,7 @@
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
     "react-streaming": "^0.3.5",
-    "typescript": "^4.6.4",
+    "typescript": "^4.8.4",
     "vite": "^3.0.4",
     "vite-plugin-ssr": "0.4.44"
   },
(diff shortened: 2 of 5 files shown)


samuelstroschein:
Note that I only added type exports to client/router as the other exports seem to work and I don't want to break things that are not broken. 

Tested locally in the repo by changing the moduleResolution of one of the examples to `node16`

brillout:
I see. Thanks 💯.

vite-plugin-ssr/node/client/router.ts
```diff
@@ -1,10 +1,11 @@
 export { navigate }
 
 import { assertWarning } from '../../utils/assert'
+import type { navigate as navigateOriginal } from '../../client/router'
```
samuelstroschein, line 4: The path `../../client/router` does not exist in the `dist` directory and is therefore incorrect when consuming vite-plugin-ssr. See https://www.jsdelivr.com/package/npm/vite-plugin-ssr?path=dist%2Fcjs

samuelstroschein: @brillout a lot of problems seem to stem from the fact that you bundle cjs + esm. is there a particular reason why you compile node-related code to cjs?

samuelstroschein: Also, v0.4.45 does not include this fix. The package.json does not define types for the export "./client/router".

brillout: Indeed. Fixed (76f1e75280f115c7ff67813ea0566254958ee3c8). It should now work with `v0.4.46`. Let me know if it doesn't.

brillout: @vchirikov PR or reproduction welcome.

PR #551, opened 2022-12-11
Readme: Add URL to comparison with NextJS

redbar0n:
Since it's useful documentation,  and was requested in https://github.com/brillout/vite-plugin-ssr/issues/158#issuecomment-1345461730

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,9 @@ Examples (official): [examples/](https://github.com/brillout/vite-plugin-ssr/tre
 <br/>
 Examples (community): [examples/README.md](https://github.com/brillout/vite-plugin-ssr/tree/main/examples#community-examples)
 
+Comparison: [vite-plugin-ssr vs. NextJS](https://github.com/brillout/vite-plugin-ssr/issues/158)
+
+
 Version history (& breaking changes): [CHANGELOG.md](/CHANGELOG.md)
 
 Follow: [Twitter > @brillout](https://twitter.com/brillout)


brillout:
While I share the sentiment here, I'm reluctant to do this out for marketing style.

Thanks for the PR nevertheless.

redbar0n:
hmm.. what do you mean by «out for marketing style»?

redbar0n:
See similar pages:

https://tanstack.com/query/v4/docs/comparison

https://formidable.com/open-source/urql/docs/comparison/

brillout:
After [VPS's new design](https://github.com/brillout/vite-plugin-ssr/issues/53) (which includes Nested Layouts) we can rename the section "Full-fledged" on `https://vite-plugin-ssr.com/` to "Feature-complete" and mention complete Next.js feature coverage.

Under "Overview" we can add a "Migrate from Next.js" title.

Since the new design will mark the v1 release, we can then rename the "Rock-solid" section to "Stable".

I foresee the v1 to live quite a long time.

But, most importantly, the big sellers are going to be the frameworks built on top of VPS. Including one I'm currently building. The second, and even bigger seller, will be Stem.

redbar0n:
Yes, but a [comparison table](https://github.com/brillout/vite-plugin-ssr/issues/158#issuecomment-1250068578) detailing every feature, even those that vite-plugin-ssr has and NextJS doesn't have, is still valuable, imho..

brillout:
Ok. But then I'd say let's add it to `https://vite-plugin-ssr.com`. How about a new page under "Overview"?

redbar0n:
Awesome! That sounds like a good idea. I suggest calling it "NextJS comparison".

Created a PR: https://github.com/brillout/vite-plugin-ssr/pull/559

PR #577, opened 2022-12-26
fix: pipe original stream asynchronously

XiNiHa:
This PR fixes a minor bug that the streaming not starting when the writable pipe from `render()` wrote the initial chunk synchronously.

### DIFF
--- a/vite-plugin-ssr/node/html/stream.ts
+++ b/vite-plugin-ssr/node/html/stream.ts
@@ -458,7 +458,7 @@ async function manipulateStream<StreamType extends Stream>({
     assert(typeof writableProxy.flush === 'function')
 
     const pipeOriginal = getStreamPipeNode(streamOriginal)
-    pipeOriginal(writableProxy)
+    setTimeout(() => pipeOriginal(writableProxy))
 
     return { streamWrapper: pipeProxy as typeof streamOriginal, streamOperations: { writeChunk, flushStream } }
   }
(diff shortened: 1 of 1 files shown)


brillout:
The root bug probably lies somewhere else.

Do you maybe have a reproduction at hand? No problem if not: I'll then merge as-is and get a deeper look next time I touch the streaming code.

XiNiHa:
I don't have a reproduction right now ;( However maybe you could easily reproduce it with SolidJS `renderToStream()` that renders a suspendable component (component with `createResource()`) wrapped with a Suspense.

brillout:
👍 Merging although note that there may be a regression in the future since I don't understand what's going on.

brillout:
Released in 0.4.67

XiNiHa:
@brillout the idea is basically deferring the pipe operation to be started after [this line](https://github.com/XiNiHa/vite-plugin-ssr/blob/886456afbc60aefd9c1508bef0aa2157992dd852/vite-plugin-ssr/node/html/stream.ts#L324) runs, so it can pass over [this branch](https://github.com/XiNiHa/vite-plugin-ssr/blob/886456afbc60aefd9c1508bef0aa2157992dd852/vite-plugin-ssr/node/html/stream.ts#LL331C33-L331C33).

brillout:
@XiNiHa I [did a bit of refactoring](https://github.com/brillout/vite-plugin-ssr/compare/v0.4.67...v0.4.68#diff-e3f6550e0e5a01058d7e638e021612e705361c6528eace4814a296de46c54d6c). Released as `0.4.68`. The code should now better handle edge cases such as the one you described. I also removed your workaround, so let's see if your problem pops up again after updating to `0.4.68`. I've added some insightful assertions.

XiNiHa:
Can confirm that the reworked one behaves fine in my case!

brillout:
Neat. I'm glad shooting in the dark did work out :D.

Issue #654, opened 2023-02-15
Prerendering with i18n


vchirikov:
### Description

Hi @brillout, I'm continue working on my homepage and ran into the issues:

I found that prerendering with [i18n as you suggest](https://vite-plugin-ssr.com/i18n#pre-rendering)
doesn't work well. In code above you rewrite `urlOriginal` which is used in filepath generation
in `write` function
https://github.com/brillout/vite-plugin-ssr/blob/9d3cc76de9c966a9c38803553ecfcae740a02860/vite-plugin-ssr/node/prerender/runPrerender.ts#L695

so it produces **double writes** ( N == locale count actually ) on prerendering and last locale in `prerender()` wins. I think it's a bug and write should try to use `_urlPristine` https://github.com/brillout/vite-plugin-ssr/blob/9d3cc76de9c966a9c38803553ecfcae740a02860/vite-plugin-ssr/shared/route/callOnBeforeRouteHook.ts#L76 

if it presented in `pageContext`.

And I tried to make a workaround with `onPagePrerender()` (from [prerender() programmatic](https://vite-plugin-ssr.com/prerender-programmatic) ) and ran into issues, because I want to use typescript (ts-node here) with `module: esm`, but vite-plugin-ssr doesn't export `prerender` in `dist/esm` and only in cjs, but even with `esModuleInterop` I got an import error. Could you also fix this, because for now the workaround to use cjs only :(

brillout:
I'm not understanding the first issue.

As for the second issue, I'm fairly confident it's a problem on your side.

Closing but I'll re-open I can understand the first issue.

brillout:
Let me check.

brillout:
I can't reproduce:
![image](https://user-images.githubusercontent.com/1005638/219350897-15479a62-13cc-482d-81e1-60cf7657b2cd.png)

brillout:
I did run `$ vite build && vite preview`.

brillout:
![image](https://user-images.githubusercontent.com/1005638/219352989-59915115-841c-44be-bc0c-08b6d511cd22.png)

brillout:
Can you try in a posix environment? That's the only potential diff I can see.

brillout:
[...]

PR #657, opened 2023-02-17
fix: add pageContextMatcher for prerendering

vchirikov:
Fix for #654 
I published it & checked in https://github.com/vchirikov/vite-plugin-ssr-i18-issue

https://github.com/vchirikov/vite-plugin-ssr-i18-issue/blob/7e1d52955b7766229380b6a4efaa734d001fa801/prerender.js#L14-L17

Works fine, but feel free to do it more elegant :)


brillout:
👍 Although I'm not sure I understand what exactly the PR is solving. I think the best would be a minimal reproduction with a description of what is happening and what you expect instead.

Issue #638, opened 2023-02-08
Asset injection causes "A charset attribute on a meta element found after the first 1024 bytes." error in Lighthouse


Csszabi98:
### Description

The asset injection strategy causes the following lighthouse error: "A charset attribute on a meta element found after the first 1024 bytes." .

My template looks like this:

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
<!-- ... -->
```

What my output template will be in production:
```
<!DOCTYPE html>
<html lang="en">
  <head>
    <link rel="stylesheet" type="text/css" href="/assets/css/ResponsiveLayout-435854b2.css">
    <link rel="stylesheet" type="text/css" href="/assets/css/ClientOnly-8a52504d.css">
    <link rel="stylesheet" type="text/css" href="/assets/css/default-f8722781.css">
    <link rel="stylesheet" type="text/css" href="/assets/css/_default.page.client-ba829c0f.css">
    <meta charset="UTF-8">
<!-- ... -->
```

As the assets are injected to the beginning of the head the meta charset gets pushed down which causes the lighthouse error.

**Possible solutions:**
Injecting assets to the end of the head tag would solve this issue. (Although I don't know what type of drawbacks it would bring)

brillout:
I don't think this has any real consequence? I'll re-open if there is a real problem here.

> Injecting assets to the end of the head tag would solve this issue. (Although I don't know what type of drawbacks it would bring)

CSS should be loaded ASAP, so I don't think this is an option.

Csszabi98:
@brillout 
It does have real consequences and can cause issues.
Browsers are not defaulting to "utf-8" character encoding, in fact, they are using heuristics to determine which encoding to use, and this process can lead to performance issues or even incorrectly inferred character encoding in some rare cases.

Cited from the WHATWG docs:
> § 13.2.3.2 Determining the character encoding
> ...
> The user agent may wait for more bytes of the resource to be available, either in this step or at any later step in this algorithm. For instance, a user agent might wait 500ms or 1024 bytes, whichever came first. In general preparsing the source to find the encoding improves performance, as it reduces the need to throw away the data structures used when parsing upon finding the encoding information. However, if the user agent delays too long to obtain data to determine the encoding, then the cost of the delay could outweigh any performance improvements from the preparse.

https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding

[...]

PR #669, opened 2023-02-24
Feature/#638 inject assets after meta charset

Csszabi98:


Csszabi98:
@brillout Could you please reopen the related ticket? (#638)
Thanks!

examples/stem-react/pages/_default/Head.tsx
```diff
@@ -8,6 +8,7 @@ const description = 'My first Vite/Stem app'
 function Head() {
   return (
     <>
```
Csszabi98, line 10: The `@brillout/stem-react` package's current implementation will always trigger the warning.
https://github.com/brillout/stem-react/blob/main/renderer/_default.page.server.tsx#L34

brillout:
Neat 💚.

> best practice to always include it in the HTML source. Not doing so could lead to a variety of issues.

Do you know the worst-case consequence of omitting setting the charset? If the consequence is potentially bad/sneaky (hard to find that the root cause is because of a missing charset value), then we should def keep both `assertWarning()`. If the consequences are mild/obvious then I think we can show only the second warning.

brillout:
I'm starting to question the whole idea of using RegExp to inject assets. Not only is it complex and error prone, but it also performs badly (especially with SSR that generates long HTML).

Maybe we can find a better alternative than using RegExp. Thoughts?

Csszabi98:
Yes, I was having the same feelings during the implementation. Usually when manipulating html string source the first thing I use is DOMParser. That is more robust by miles and a much more pleasantly experience to work with as you are just using the document API. For the node environment there are two main choices in this regard, HappyDOM and JsDOM for this purpose. The former is more performant while the latter is more mature.

brillout:
I'm thinking maybe something like this:

```js
import { html } from 'vite-plugin-ssr'

export function render() {
   return html`<html>
     <head>
       <meta charset="en" />
       ${html.ASSETS_BEGIN}
     </head>
     <body>
       ${html.ASSETS_END}
     </body>
   </html>`
}
```

That way we don't need to manipulate strings at all &ndash; we just concatenate strings instead.

Csszabi98:
If injecting assets is the only thing required, then this seems simple enough ➕.
Only thing is that this seems like a breaking change, so I would keep the old behavior while adding this as an extra feature.

brillout:
Since my proposal involves a little bit of architectural changes, I propse I implement this myself. Let me make some progress on the V1 design then I'll implement this on this PR. I'll keep you updated.

> I would keep the old behavior while adding this as an extra feature

Yes, that's the plan.

Csszabi98:
Sounds great, thanks!

brillout:
"Merged" 😀 🎉.

Many thanks for the PR, very useful to realize we should go with another path.

PR #674, opened 2023-02-28
v1 alternative proposal

samuelstroschein:
## Problem 

The v1 proposal leads to a lot of boilerplate. 

## Proposal 

Only have one touchpoint with VPS namely `+config` files. The mental model of understanding VPS become much simpler boosted with additional DX through autocomplete and typesafety. 

1. Have one config schema that defines VPS and can be set in a `+config.ts` file. 
2. Nested routes can overwrite parent `+config` files. 
3. `+config` file are regular JS with a few conventions.

## Explanation 

- A `+config` is defined at the root https://github.com/brillout/vite-plugin-ssr/pull/674/files#diff-356c36b8f4660461f9162c9fbb7d0e3f51c76b60d45bafa398d621b5d98e3379. 
- subroutes can define `+config` files too, see for example https://github.com/brillout/vite-plugin-ssr/pull/674/files#diff-33c2ab2a1b673c6cd4afe87e4a33a54fa0c0aa15437ff531d7ca26e59c67c7d0

## Open questions 

1. Some config properties should only affect one route e.g. `onBeforeRender` and should not be propagated to nested children. How to deal with that? 
2. @brillout had reasons why he did not want to allow JS in the config. I recall scope problems (using global variables) and code splitting. Scope problems are easy: Disallow or lint global variables in the `+config` file. Regarding code splitting, I hope that dynamic imports resolve the problem.


brillout:
> `onBeforeRender: () => import('./onBeforeRender').then((m) => m.default)`

That was actually my original design. But after experimenting with it I found it annoying to have to modify `+config.js` every time I defined a new hook. It felt pesky and without adding enough value.

> ```js
> // /pages/error-page/+config.js
>
> export default {
>   isErrorPage: true
> }
> ```
That's actually already possible and is equivalent to `+isErrorPage.js`.

> // define functions directly in the config

My concern:

```js
// +config.js

export default { onRenderHtml, onRenderClient }

var someVar;

function onRenderHtml() {
   // This variable
   console.log(someVar)
}

function onRenderClient() {
   // And this variable are *not* the same
   console.log(someVar)
}
```

Because `onRenderHtml()` lives on the server-side and `onRenderClient()` lives on the client-side, the variable `someVar` isn't the same.

This breaks the JavaScript semantics.

I'm not categorically against it and I'm actually talking with someone who's working on a framework-agnostic libray to do this kind of "server code splitting". (It isn't a trivial problem and requires a decent amount of edge case handling.)

If the ecosystem at large goes in that direction, then Vike will adopt this. But, so far, I'd rather go with a more conservative approach. Let's see.


> // the pattern of pages to render
>  pagePattern: '**/+Page.tsx'

I'm not sure what a use case for this would be. In general / by default, I'd rather go for "convention over configuration". But if there is a use case with a concrete added value, then I'm open to consider this.

> // no need for so much boilerplate

I definitely share the sentiment here. But because `+` files are loaded in different environments I don't see much of an alternative. FYI it isn't only server/client but, also, some are "config only" and loaded only while building (there aren't loaded in production runtimes). And some are routing configs that have their own separate loading logic. That's why with the current `0.4` design configs can live in 5 different places:
 - `.page.js`
 - `.page.client.js`
 - `.page.server.js`
 - `.page.route.js`
 - `vite.config.js` (note that with the V1 design all configs previously defined in `vite.config.js` are now defined in `+config.js`)

<br/>

Let me know if I missed a change in `git log -p v1-proposal`.

brillout:
> VPS is configured via a root `+config` file with nested `+config` files per route.

I'd love this, I really would. But I'm not sure how/if we can achieve this.


> 2\. But lint global variables in `+config` to avoid the footgun of different variable scopes.

Hooks often need global variables, e.g. https://github.com/brillout/vite-plugin-ssr/blob/a9057be9beb08854034b5cc24572edf4292f1d02/examples/react-full-v1/renderer/+onRenderClient.tsx#L10

So I'm not sure how/if we can implement this. Maybe there is a way, but so far I can't see a solution.

The "closure problem" (global variables being the most problematic instance) is the only problem though (all other issues are fixable). So maybe there is hope.

samuelstroschein:
> Hooks often need global variables, e.g.
> So I'm not sure how/if we can implement this. Maybe there is a way, but so far I can't see a solution.

> The "closure problem" (global variables being the most problematic instance) is the only problem though (all other issues are fixable). So maybe there is hope.Hooks often need global variables, e.g.

That's okay. 

If my `onRenderClient` functions needs a global variable, I need to define it in another file that I import into `+config`: 

```ts
onRenderHtml: () => import('./renderer/onRenderHtml.jsx'),
```

I argue that VPS should only have one enforced rule "no global variable in a +config file". Developers are free to define properties of `+config` like `isErrorPage` or `onBeforeRender` directly in `+config` if no global variable is required. Else, import the function from a file BUT VPS does not enforce the file to be named `+isErrorPage` or `+onBeforeRender` and what not.

brillout:
Alternatively, VPS can forbid *shared* global variables instead. That would
solve the problem.

I’ll think about it.

samuelstroschein:
> Alternatively, VPS can forbid *shared* global variables instead. That would
solve the problem.

If you could pull that off, great DX improvement. I (and everyone really) ran into the issue of shared variables that shouldn't be shared already. Simple quick fix would be disallowing shared variables in the `+config` file.

redbar0n:
> My concern:
> 
> ```js
> // +config.js
> 
> export default { onRenderHtml, onRenderClient }
> 
> var someVar;
> 
> function onRenderHtml() {
>    // This variable
>    console.log(someVar)
> }
> 
> function onRenderClient() {
>    // And this variable are *not* the same
>    console.log(someVar)
> }
> ```
> Because onRenderHtml() lives on the server-side and onRenderClient() lives on the client-side, the variable someVar isn't the same.

Could you make them the same? By sending `someVar` and other global variables to the client in the HTML and then loading them from there on the client side? Could be a feature, idk.

samuelstroschein:
> Could you make them the same? By sending someVar and other global variables to the client in the HTML and then loading them from there when runningonRenderClient? Could be a feature, idk. 

Risk for cross-request state pollution. Two concurrent SSR requests share the same state (on the server) while they should not.

brillout:
> Could you make them the same? By sending `someVar` and other global variables to the client in the HTML and then loading them from there when running`onRenderClient`? Could be a feature, idk.

AFAICT the added value isn't worth the added complexity.

brillout:
Actually, I didn't mention the biggest problem: leaking server secrets and accidentally adding server code to the client-side bundle. Big apps get messy, this *will* happen.

At the end of the day, it's only about creating a couple more files, and, so far, I think it's worth it given the added safety. It's not like each page will have 10 `+` files, but realistically more like 2-4 of them, so I'm thinking it's ok.

But let's see with some real apps whether creating `+` files becomes *that* annoying.

I suggest we wait and see how the ecosystem evolves &ndash; maybe Remix / Solid Start / TanStack Start can pull off a sound implementation with decent protective measures.

FYI the React team tried but ended up deciding to do file based splitting instead. Dan: https://twitter.com/dan_abramov/status/1630573034120650761?s=46&t=0mDYJk_N9_CWJRyO-zZz6A and I know sebmarkbage feels the same.

brillout:
We can still make everything be defined in `+config.js`:

```ts
// +config.ts

import type { Config } from 'vite-plugin-ssr'

export default {
  Page: () => import('./Page'),
  onBeforeRender: () => import('./onBeforeRender')
} satisfies Config
```

> onBeforeRender: () => import('./onBeforeRender').then((m) => m.default)

We can remove `.then((m) => m.default)` as we can set `Config['onBeforeRender']` to `() => Promise<{ default: OnBeforeRender }>`.

Also, we could do this:

```ts
// +config.ts

// The file Page.js isn't actually imported (VPS replaces it with a pointer instead)
import Page from './Page'
import onBeforeRender from './onBeforeRender'
import type { Config } from 'vite-plugin-ssr'

export default { Page, onBeforeRender } satisfies Config
```

It's aesthetically nicer and has slighther better DX, but it adds confusion for those who aren't familiar with VPS (they'll be tempted to define hooks inside `+config.ts` which is forbidden). So far I've a preference for `() => import('./Page')`.

samuelstroschein:
> leaking server secrets and accidentally adding server code to the client-side bundle. Big apps get messy, this will happen. 

Are you sure? If VPS disallows global variables and used dynamic imports in `+config`, leaking server code into the client should not be possible. Nothing is imported into the `+config` files, only on executing the import functions which are split by client/server.

brillout:
I think we also need to forbid closure for functions.

```js
// /renderer/+config.js

export default { onRenderClient }

function onRenderClient() {
  // This is forbidden
  someFn()
}

function someFn() {
}
```
It seems like inline hooks would end up being quite limited, and what I can see hapening then is that most users will end up moving hooks from `+config.js` to an extra file.

Another downside is that it adds friction as we have to explain the user the rules of inline hooks.

It makes me think that I'd rather go for slightly more annoying but simpler, than less annoying but more complicated.

As much as I'd like this to work out, I don't see a clear and simple path for supporting inline hooks (yet?).

I'll give this further thoughts if users complain about having to create too many `+` files. (I think the situation isn't as bad as it seems. We'll see.)

samuelstroschein:
> I think we also need to forbid closure for functions.

You are not solving the problem of leaking server-side information by forbidding closures. The problem would just be moved to dedicated `+` files instead. I don't see the benefit.  

> It seems like inline hooks would end up being quite limited, and what I can see hapening then is that most users will end up moving hooks from +config.js to an extra file. 

Even if that were the case, great!

Users can use dedicated files or not. VPS stays simple "VPS is configured via a root +config file with nested +config files per route. That's it." Import files, don't import files. Your choice.

brillout:
Closing but we can continue the dicussion.

redbar0n:
> Risk for cross-request state pollution. Two concurrent SSR requests share the same state (on the server) while they should not.

How would they, if the execution context is different? Either in a separate Lambda function, or v8 isolate?

If you are running a persistent Node server, then since Node is single-threaded it wouldn't bleed across requests, right? https://stackoverflow.com/a/60858423/380607

brillout:
Update: instead of `() => import('./Page')`, I think VPS should support static imports as shown in the example https://github.com/brillout/vite-plugin-ssr/issues/578#issuecomment-1479297807. For improved TS ergonomics.

I actually can see that, with static imports, your original sentiment and longing will be mostly fulfilled.

PR #694, opened 2023-03-09
Docs: Type mismatch explanation for dynamic imports with TS

charlieforward9:
Originally planned to put the explanation outside of the code, but preferred to leave it as a comment to avoid 'not-seeing the notice' discussions from showing up in the future.


charlieforward9:
Addressing #693

brillout:
Thanks! Also thanks for the sponsoring 💚.

charlieforward9:
Thanks for providing great software and support!

brillout:
I'm glad you like it!

Issue #445, opened 2022-09-09
Server runtime: remove need for node compat layer (=> removing 200k from worker bundle size)


brillout:
### Description

We can make the server runtime completely agnostic to Node.js. Removing the need for Node.js shims on Edge platforms.

brillout:
If someone wants/needs this, PM on discord.

zeyuri:
I would be interested in doing this

brillout:
> I would be interested in doing this

👍 PR welcome.

zeyuri:
Can you help me a little bit giving some guidance? How I have some experience porting some code from node to workers but I will be a little lost trying to find where this parts of the code are.

Also how would we make this work to not be a breaking change? A flag passed to the ssr plugin setup inside the `vite.config.ts` ?

brillout:
I believe we can replace all runtime Node.js dependencies with our own functions (self written or copied from e.g. StackOverflow). I think it's fairly straigthforward (although I may miss something). Last time I checked there were only few trivial Node.js dependencies. Note that it's only about the dependencies of the runtime code, not the entire VPS code.

With that approach we don't need any new VPS flag.

I'd run the Cloudflare example (/examples/cloudflare) without the Node.js compat flag and fix the errors until it works.

brillout:
Yea, I've been thinking about adding `eslint`. Although note that this is about the server runtime. VPS's Vite plugin, for example, will still use Node.js.

But we can use the Cloudflare Workers tests to ensure that we don't mistakenly add back a Node.js dependency.

kochrt:
I'm mildly curious about this so I am fiddling around with it a bit. Not sure how far I'll make it so wanted to document in case I drop off and someone wants to pick up.

> I'd run the Cloudflare example (/examples/cloudflare) without the Node.js compat flag and fix the errors until it works.

Seems the initial errors are all about `path`:

```sh
[ERROR] Could not resolve "path"

    node_modules/@brillout/import/dist/index.js:8:39:
      8 │ const path_1 = __importDefault(require("path"));
        ╵                                        ~~~~~~

  The package "path" wasn't found on the file system but is built into node.
  Add "node_compat = true" to your wrangler.toml file to enable Node compatibility.


✘ [ERROR] Could not resolve "path"

    node_modules/vite-plugin-ssr/dist/cjs/node/runtime/renderPage/getPageAssets.js:10:39:
      10 │ const path_1 = __importDefault(require("path"));
[...]

PR #768, opened 2023-04-08
Replace node path with pathe library

kochrt:
Haven't run tests yet


kochrt:
#445

brillout:
Tests are passing, including the [wrangler ones](https://github.com/brillout/vite-plugin-ssr/commits/tmp-remove-path) 🎉.

Maybe we can use a little custom implementation instead of `pathe`? So far, it seems me so that it's doable with only a couple lines of code. Especially given that we can assume paths to be posix (Vite normalizes the paths).

kochrt:
After pushing a minimal implementation I see you say vite normalizes the path. Does this mean we don't need the `normalize` method? We can just assume it's all normalized? That would help because I was about to say I've pulled in like half of the `pathe` library into this shim

brillout:
Yes, Vite does some normalization already. I didn't take a close look at it but seems like we don't need most of it?

If we do then, yea, we might as well just use `pathe` then 😄.

kochrt:
If it were up to me I would just go with `pathe`, it's small, doesn't have any dependencies of its own, and I trust it more than a shim that I'm picking and choosing to implement. Maybe make a follow up bug to pull out only the bare minimum from `pathe`.

If I knew more about what vite does and doesn't normalize, or wasn't just error-chasing in this PR, I might be more confident in my shim.

brillout:
Released as `0.4.112`.

FYI: the shim ended being a one-liner. We saved ourself an extra dependency 🙂.

Thanks for the PR. It was fun.

PR #779, opened 2023-04-12
Update Features.tsx

rand:
Moving Netlify from **Static Host** to **Deploy Anywhere**, as it offers full support for dynamic applications of this type.

### DIFF
--- a/docs/pages/index/features/Features.tsx
+++ b/docs/pages/index/features/Features.tsx
@@ -142,11 +142,11 @@ function Features() {
           desc: (
             <>
               <p>
-                <b>Deploy anywhere</b> (Netlify Functions, Vercel Serverless/Edge/ISR, Cloudflare Workers, AWS EC2/Lambda, Firebase, Google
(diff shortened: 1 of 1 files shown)


brillout:
👍

How does Netlify call its static hosting offering? We may want to keep it in the list as it's quite a popular static host.

Also I think we should list Netlify Functions/Edge after AWS and before Firebase. (In order of popularity).

rand:
Netlify's static hosting is more a subset of its general front-end application support and caching product. If the build yields no dynamic elements, only static content is published and cached. Would you like me to include it in both sections?

brillout:
I see. Yes I still think it makes sense to list it in both sections.

How about just "Netlify" in static hosts and "Netlify Functions/Edge" in the other list? Or just "Netlify Functions" &mdash; I don't know whether "Netlify Edge" is a widespread terminology.

rand:
Thanks. I've pushed updates to reflect that.

brillout:
Wonderful. Thanks.

PR #813, opened 2023-04-24
Chore: Add Svelte (v1) example

gu-stav:
This adds a new (v1) example using Svelte. I've replicated the react-setup with the sidebar.


gu-stav:
@brillout Thanks for the fixes and the quick merge! 

> These framework prototypes showcase that Vite(-plugin-ssr) can be fully wrapped.

I agree and I think for any full-fledged meta-framework that would be the way to go. It is a detail but I think you could make an argument for having vite, vite plugin svelte and svelte itself dependencies of the users project in the end, because it allows them to update these dependencies independent from the meta-framework. In case of vite and svelte which are both very stable and deliver incremental improvements over time this might be beneficial ... (and having them as peer dependencies in the meta-framework).

Having the ability to compose several vite plugins on the top-level might also give a very neat DX.

> That's one of the reasons why I'm confident vite-plugin-ssr's flexibility can accomodate Strapi's plans. Strapi can fully wrap vite-plugin-ssr or just provide deep vite-plugin-ssr integrations, choose any UI framework, etc.

👍🏼 

> Btw. if you're on Discord, feel free to [PM](https://discord.com/invite/dSDMGGJZQy) us!

I am - will do. 🙏🏼

Issue #836, opened 2023-05-01
`pageContext._pageFilesAll` isn't available in the browser


rtritto:
### Description

- Clone https://github.com/brillout/vite-plugin-ssr/tree/main/examples/solid-ssr
- Create `/pages/index/index.page.server.tsx` and `/pages/about/index.page.server.tsx`
```ts
export async function onBeforeRender(pageContext) {
  return {
    pageContext: {
      pageProps: {
        ...
      }
    }
  }
}

export const passToClient = ['pageProps']
```
- `yarn vite`
- Go to `localhost:3000`
- Click on _About_ page button
- See console logs in browser

Related: https://github.com/brillout/vite-plugin-ssr/commit/c905a2bedbe814fd45a1f2fa3c80dfc3a294b9dc

### Error Message + Error Stack

```shell
Uncaught (in promise) Error: [vite-plugin-ssr@0.4.120][Wrong Usage] pageContext._pageFilesAll isn't available in the browser ('_pageFilesAll' is missing in the passToClient list ['pageProps', 'documentProps']). Did you forget to add '_pageFilesAll' to the passToClient list? See https://vite-plugin-ssr.com/passToClient
    createErrorWithCleanStackTrace createErrorWithCleanStackTrace.js:8
    assertUsage assert.js:51
    assertPassToClient releasePageContext.js:98
    get releasePageContext.js:76
    applyState dev.js:375
    reconcile dev.js:386
    updatePath dev.js:196
    setStore dev.js:216
    runUpdates dev.js:800
    batch dev.js:427
    setStore dev.js:215
    render _default.page.client.tsx:33
    hookResult executeOnClientRender.js:46
    callHookWithTimeout callHookWithTimeout.js:32
    callHookWithTimeout callHookWithTimeout.js:38
    executeOnClientRender executeOnClientRender.js:46
    renderPromise useClientRouter.js:188
    fetchAndRender useClientRouter.js:190
    useClientRouter useClientRouter.js:39
    onClick useClientRouter.js:230
    onLinkClick useClientRouter.js:215
    useClientRouter useClientRouter.js:37
    <anonymous> entry.js:7
```

brillout:
Can you publish the reproduction as Github repository?

rtritto:
https://github.com/rtritto/vps-pageFilesAll

brillout:
Thanks. And indeed I could reproduce. Let me have a look.

rtritto:
It works with this change (idk if it's ok):
[...]

PR #841, opened 2023-05-01
chore: improve solid render reconcile

rtritto:
Improve https://github.com/brillout/vite-plugin-ssr/commit/d5c001f11345841a1fd6d21a00ec7daeaf57b5d9
Related #836


brillout:
👍 You're right the changes should be applied to `/examples/solid-ssr-v1/` as well.

I reverted the changes to `/examples/solid-ssr/` since it's actually more accurate and I find the code more readable.

PR #866, opened 2023-05-07
Add eslint to solid examples

rtritto:
### TODO
- [ ] lint all
- [ ] run `pnpm i` to update _pnpm-lock.yaml_

### DIFF
--- a/examples/solid-spa-v1/.eslintrc.yml
+++ b/examples/solid-spa-v1/.eslintrc.yml
@@ -1,0 +1,24 @@
+extends:
+  - eslint:recommended
+  - plugin:@typescript-eslint/recommended
+  - plugin:solid/recommended
+  - plugin:unicorn/all
+parser: '@typescript-eslint/parser'
+parserOptions:
+  ecmaFeatures:
+    jsx: true
+  ecmaVersion: latest
+  sourceType: module
+plugins:
+  - '@typescript-eslint'
+  - solid
+rules:
+  # 0 = off
+  # 1 = warn
+  # 2 = error
+
+  #region eslint-plugin-unicorn
+  unicorn/filename-case: 0
+  unicorn/no-keyword-prefix: 0
+  unicorn/prevent-abbreviations: 0
+  #endregion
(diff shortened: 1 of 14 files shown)


rtritto:
Current:
```console
C:\workspace\vite-plugin-ssr\examples\solid-ssr-v1\renderer\+onRenderClient.tsx
  26:23  warning  Forbidden non-null assertion                                  @typescript-eslint/no-non-null-assertion
  46:3   error    Do not use "@ts-ignore" because it alters compilation errors  @typescript-eslint/ban-ts-comment
  48:3   error    Do not use "@ts-ignore" because it alters compilation errors  @typescript-eslint/ban-ts-comment

C:\workspace\vite-plugin-ssr\examples\solid-ssr-v1\renderer\Link.tsx
  7:67  error  Use `undefined` instead of `null`  unicorn/no-null

C:\workspace\vite-plugin-ssr\examples\solid-ssr-v1\renderer\types.ts
  3:25  error  Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `object` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
- If you really want a type meaning "any non-nullish value", you probably want `NonNullable<unknown>` instead  @typescript-eslint/ban-types

C:\workspace\vite-plugin-ssr\examples\solid-ssr-v1\renderer\usePageContext.tsx
  11:8   warning  The reactive variable 'props.pageContext' should be used within JSX, a tracked scope (like createEffect), or inside an event handler function, or else changes will be ignored  solid/reactivity
  12:35  warning  The reactive variable 'props.pageContext' should be used within JSX, a tracked scope (like createEffect), or inside an event handler function, or else changes will be ignored  solid/reactivity

✖ 7 problems (4 errors, 3 warnings)
```

`@typescript-eslint/ban-ts-comment` will be removed with VPS v1.

Some help is needed to others.

brillout:
Thanks for PR.

In genreal, the idea of `/examples/` is to be minimal and to showcase only one particular aspect. So I'm inclined to not add ESLint to the examples. But we're woking on Solid boilerplates and  we can then add ESLint over there.

Let's close in the meantime and let's continue the conversation once the boilerplates are ready.

PR #919, opened 2023-05-27
chore(docs): add note about parts of the pageContext not being availa…

### LINKED ISSUE #917: urlParsed.search is empty in prerender mode
### Description

Hi again, firstly thanks for the fix in #2784ac9. urlParsed is now making its way into the prerender build.

However, I have a follow-on problem. Take a look at the code below:

https://github.com/brillout/vite-plugin-ssr/compare/main...philwinder:vite-plugin-ssr:main?expand=1

In this code I have enabled prerendering, added the usual `usePageContext` script and added the `PageContextProvider`, like in the full react demo.

If you run this example with `yarn dev` and browse to http://localhost:3000/?hello=world#hash-hash then you will see:
```
pageContext.urlParsed: { "origin": null, "pathname": "/", "pathnameOriginal": "/", "search": { "hello": "world" }, "searchAll": { "hello": [ "world" ] }, "searchOriginal": "?hello=world", "hash": "", "hashOriginal": null }
```

But if you run `yarn preview` and browse to the same URL you will see:
```
pageContext.urlParsed: { "origin": null, "pathname": "/", "pathnameOriginal": "/", "search": {}, "searchAll": {}, "searchOriginal": null, "hash": "", "hashOriginal": null }
```

Note how the search* fields are now null, even though they work in dev. Also note how hash doesn't seem to work in either (according to my understanding of what hash should do).

I can't immediately see why this is; I suspect it's being overwritten somewhere else?

Thanks, Phil

philwinder:
…ble in pre-render mode

Fixes #917.

### DIFF
--- a/docs/pages/pageContext.page.server.mdx
+++ b/docs/pages/pageContext.page.server.mdx
@@ -5,6 +5,10 @@ The object `pageContext` holds information about the current page.
 > The `pageContext` object is completely discarded and created anew when a new page is rendered.
 > That's why it's called *page*Context (and not *app*Context).
 
+> `pageContext` is meant for rendering. When using pre-rendering you render at build-time.
+> Parts of the contenxt, like query parameters, don't exist at build-time. 
+> For example, `pageContext.urlParsed.searchAll` will always be null when pre-rendering.
+
 ## Built-in
 
 Built-in properties:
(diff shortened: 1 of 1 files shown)


brillout:
FYI https://github.com/brillout/vite-plugin-ssr/pull/919/commits/ea06b4557e034b7798d01a8d907b016da64c0ec0. Feel free to let me know if you believe that wouldn't have helped.

PR #936, opened 2023-06-06
Create dynamic metadata in react boilerplate

BanDroid:
usage example:

```tsx
async function getDocumentProps(pageProps) {
  const res = await fetch(process.env.API_URL);
  const data = await res.json();
  return {
    title: data.title,
    description: data.description
  }
}

function Page() {
  return <></>;
}

export { Page, getDocumentProps }
```


brillout:
The idea of the boilerplates is also to let users implement such features themselves. So I think it isn't worth it to add this.

I appreciate the intention.

PR #937, opened 2023-06-06
filesystemRoutingRoot: Needs a string but TS asks for a boolean

crummy:
I have this TS code which returns an error, `TS2322: Type 'string' is not assignable to type 'boolean'.`

```
export default {
  filesystemRoutingRoot: "/",
} satisfies Config;
```

Config seems to require `filesystemRoutingRoot` to be a boolean, but if I set it to a boolean, I get this error at runtime:
```
Error: [vite-plugin-ssr@0.4.131][Wrong Usage] /ssg/pages/+config.h.ts > `export default { filesystemRoutingRoot } should be a string
```

Therefore I believe it should be a string.


brillout:
👍 Indeed

PR #942, opened 2023-06-07
Update `all.page.server.js`, fixed server side route

zyxkad:
We should use `router.push(pageContext.urlOriginal)` but not `urlPathname`, for pass query and hash to the renderer


brillout:
👍 Makes sense.

PR #977, opened 2023-06-27
Add a next/styled-jsx css example

jeremypress:
Hi, 
I recently added SSR backed support of styled jsx in a vite SSR plugin i'm working on. It was painful to figure out, so I wanted to contribute an example. styled-jsx is a default in next.js, which we're moving from, so I think this will help others building out the their own stack with Vite. 

Let me know what else you'd like to see added, I was planning on adding the docs in another change

Todo
- [ ] regenerate the tests


brillout:
Neat 💯.

Actually, a lot of examples are going to be moved into their own repository. I think we should do the same for this example. Do you mind publishing it under `github.com/jeremypress`?

jeremypress:
sure thing! I'll stay consistent with the format/styles of this repo, but I think I would opt to drop CI and then link to it from your docs?

brillout:
Sounds good!

brillout:
I guess we can close this in the meantime. Looking forward to add your example to the docs 👀. (Feel free to just republish it under your GitHub name and I'll take care of the rest.)

Issue #1014, opened 2023-07-11
Out-dated content in react tour


bzy-debug:
### Description

Hi, thanks for your great work! I find some outdated content in the react tour section of the documentation:

In the example at start

``` jsx
// /renderer/_default.page.server.jsx
// Environment: server

import ReactDOMServer from "react-dom/server";
import React from "react";
import { escapeInject, dangerouslySkipEscape } from "vite-plugin-ssr";

export { render };

async function render(pageContext) {
  const { Page, pageProps } = pageContext;
  const viewHtml = ReactDOMServer.renderToString(
    <Page {...pageProps} />
  );

  const title = "Vite SSR";

  return escapeInject`<!DOCTYPE html>
    <html>
      <head>
        <title>${title}</title>
      </head>
      <body>
        <div id="page-view">${dangerouslySkipEscape(viewHtml)}</div>
      </body>
    </html>`;
}
```

``` jsx
// /renderer/_default.page.client.jsx
// Environment: browser

import ReactDOM from "react-dom";
import React from "react";

export { render };

async function render(pageContext) {
  const { Page, pageProps } = pageContext
  ReactDOM.hydrate(
    <Page {...pageProps} />,
    document.getElementById("page-view")
  );
}
```

We need to add `export passToClient = ["pageProps"]` at the end of `/renderer/_default.page.server.jsx` and use `hydrateRoot` in `/renderer/_default.page.client.jsx` to sync with react-18.

If you feel okay I can open a PR to update these. :)

brillout:
Yes PR welcome :)

PR #1015, opened 2023-07-12
Update doc

bzy-debug:
Update react-tour content to be consistent with current API
see #1014

### DIFF
--- a/docs/pages/react-tour.page.server.mdx
+++ b/docs/pages/react-tour.page.server.mdx
@@ -90,8 +90,8 @@ export { render };
 async function render(pageContext) {
   const { Page, pageProps } = pageContext
   ReactDOM.hydrateRoot(
-    <Page {...pageProps} />,
-    document.getElementById("page-view")
+    document.getElementById("page-view"),
+    <Page {...pageProps} />
   );
 }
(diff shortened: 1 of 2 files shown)


brillout:
Thanks!

docs/pages/react-tour.page.server.mdx
```diff
@@ -75,8 +75,6 @@ async function render(pageContext) {
       </body>
     </html>`;
 }
- 
-export const passToClient = ["pageProps"]
```
bzy-debug: Is this line really unnecessary?

I get an error without this line:

截屏2023-07-12 15 50 47

And the button is not interactive.

Correct me if I miss anything :)

brillout: You're right, it's required. But https://vite-plugin-ssr.com/react-tour#render-control is about showcasing render control and the code snippets don't aim to be complete. The `passToClient` part is actually explained down below at https://vite-plugin-ssr.com/react-tour#data-fetching.

bzy-debug: Make sense, thanks for your explanation!

PR #1024, opened 2023-07-20
Updates "Render Modes (SPA, SSR, SSG, HTML-only)" to fix a typo

infomiho:
Tiny fix to update the missing extension in a page filename.


brillout:
Good catch 👌

PR #1025, opened 2023-07-20
Fixes "Data Fetching" docs

infomiho:
There is duplicate `Page` declarations which results with this error:

Screenshot 2023-07-20 at 14 15 39

I believe the intent was to showcase the `pageProps` being available, so I removed the first declaration with `Page` and kept the second one.


brillout:
💯 Keep them coming.

PR #1048, opened 2023-08-04
fix: oven-sh/bun#3743 - Support Bun

timnghg:
While digging deeper to make [elysia-vite-plugin-ssr](https://github.com/timnghg/elysia-vite-plugin-ssr) work, I found a working solution, continue of [brillout/vite-plugin-ssr#d410d94](https://github.com/brillout/vite-plugin-ssr/commit/d410d94a021cb2b1a48db5148cc90ceaafffedf5)


brillout:
Neat 👌.

Does it work now? 👀 (I can't test as I only have a Linux 4 machine at the moment.)

timnghg:
I am watching the test failed as heavy rain drop! 
It seems that TS don't like `Bun` global var.

brillout:
It's fine, just fixed it.

PR #1122, opened 2023-09-15
Faster URL Parsing

dajinchu:
Improves https://github.com/brillout/vite-plugin-ssr/issues/1115 by removing try/catch to avoid error handling overhead when pathname is relative and no `base` is passed to URL constructor.

Benchmark using the `npm create vite-plugin-ssr starter` repo on the index page. Used [bombardier](https://github.com/codesenberg/bombardier) on Macbook with M1 Pro.

before:
```
Statistics        Avg      Stdev        Max
  Reqs/sec      1312.95     223.14    3744.83
  Latency       94.75ms    17.55ms   244.31ms
```
after:
```
Statistics        Avg      Stdev        Max
  Reqs/sec      2269.03     439.82    6707.62
  Latency       54.98ms    12.65ms   190.80ms
```

### DIFF
--- a/vite-plugin-ssr/utils/parseUrl.ts
+++ b/vite-plugin-ssr/utils/parseUrl.ts
@@ -132,31 +132,22 @@ function parseWithNewUrl(urlWithoutHashNorSearch: string, baseServer: string) {
     urlWithoutHashNorSearch = PROTOCOL_FAKE + urlWithoutHashNorSearch.slice(PROTOCOL_TAURI.length)
   }
 
-  let origin: string | null
-  let pathnameResolved: string
-  try {
-    // `new URL(url)` throws an error if `url` doesn't have an origin
-    const urlParsed = new URL(urlWithoutHashNorSearch)
-    origin = urlParsed.origin
-    pathnameResolved = urlParsed.pathname
-  } catch (err) {
-    // `url` has no origin
-    origin = null
-    // In the browser, this is the Base URL of the current URL
-    let base =
-      typeof window !== 'undefined' &&
-      // We need to access safely in case the user sets `window` in Node.js
-      window?.document?.baseURI
-    base = base || 'http://fake.example.org' + baseServer
-    // `new Url()` supports:
-    //  - `url === '/absolute/path'`
-    //  - `url === './relative/path'`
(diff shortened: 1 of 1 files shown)


vite-plugin-ssr/utils/parseUrl.ts
```diff
@@ -132,31 +132,22 @@ function parseWithNewUrl(urlWithoutHashNorSearch: string, baseServer: string) {
...
+
+  // `new Url()` supports:
+  //  - `url === '/absolute/path'`
+  //  - `url === './relative/path'`
+  //  - `url === '?queryWithoutPath'`
+  //  - `url === '''`
+  // `base` in `new URL(url, base)` is used for resolving relative paths (`new URL()` doesn't remove `base` from `pathname`)
+  const urlParsed = new URL(urlWithoutHashNorSearch, base)
+  let pathnameResolved = urlParsed.pathname
+  let origin = urlWithoutHashNorSearch.startsWith(urlParsed.origin) ? urlParsed.origin : null
```
dajinchu, line 150: If the parsed origin is inside `urlWithoutHashNorSearch`, that means that's actually the origin to return - if it isn't included, then it must have been a relative path, in which case we return null.

brillout: Good idea 👍

brillout:
> before:
> 
> ```
> Statistics        Avg      Stdev        Max
>   Reqs/sec      1312.95     223.14    3744.83
>   Latency       94.75ms    17.55ms   244.31ms
> ```
> 
> after:
> 
> ```
> Statistics        Avg      Stdev        Max
>   Reqs/sec      2269.03     439.82    6707.62
>   Latency       54.98ms    12.65ms   190.80ms
> ```

That's very surprising. I wonder why `new URL()` is so slow.

brillout:
Thanks for the PR!

dajinchu:
> That's very surprising. I wonder why new URL() is so slow.

Its less that URL is slow and more that error handling is slow

brillout:
> Its less that URL is slow and more that error handling is slow

I would find this even more surprising.

I guess the way to find out is to benchmark the difference between no `new URL()` usage at all and using it once per `parseUrl()` call 👀.

PR #1135, opened 2023-09-21
docs: minor improvements around conditional DOM hydration // #1133

AurelienLourot:


brillout:
👍 Although one thing I'm concerned about is that Bati scaffolds a V1 design app while the entire docs is currently documenting the old design.

AurelienLourot:
@brillout this should be better now

PR #1137, opened 2023-09-21
qual: make config hook types more precise

AurelienLourot:
@brillout, validated with `tsc --noEmit` on the example. Is this going in the right direction? If yes I would like to:

1. do the same for other hooks
2. add types to all such hooks in all existing typescript examples.

Is it OK with you?


AurelienLourot:
@brillout I think it is better now

vite-plugin-ssr/shared/page-configs/Config.ts
```diff
@@ -39,6 +40,14 @@ type Config = ConfigBuiltIn &
     | VikePackages.ConfigVikeSvelte
   )
 
+// Wrapper that transforms a sync function type into an async one.
+// See https://stackoverflow.com/questions/72916246/how-do-you-create-an-async-version-of-a-typescript-function-type
+type Async<T extends (...args: any) => any> = (...args: Parameters<T>) => Promise<ReturnType<T>>
+
+type OnBeforeRenderSync = <PC extends PageContextBuiltInCommon>(pageContext: PC) => { pageContext: Record<string, any> }
+type OnBeforeRenderAsync = Async<OnBeforeRenderSync>
+type OnBeforeRenderFunction = OnBeforeRenderSync | OnBeforeRenderAsync
```
AurelienLourot, line 49: Let's change this into something like

```
type OptionalPromise<T> = T | Promise<T>

type OnBeforeRenderFunction = <PC extends PageContext>(pageContext: PC) => OptionalPromise<{ pageContext: Record<string, any> }>
```

And let's wait for @brillout to have made `PageContext` available as a type here.

We also agreed to touch only this one `react-full-v1` example for now, but to touch other hooks like `onRenderClient` and `onRenderHtml`.

AurelienLourot:
@brillout does it now look more like what you had in mind? Thanks

AurelienLourot:
```
examples/auth $ pnpm exec tsc --noEmit
pages/+config.h.ts:8:3 - error TS2322: Type '{ onRenderHtml: "import:vike-react/renderer/onRenderHtml"; onRenderClient: "import:vike-react/renderer/onRenderClient"; passToClient: string[]; clientRouting: true; hydrationCanBeAborted: true; meta: { ...; }; }' is not assignable to type 'Config | `import:${string}` | Config[] | `import:${string}`[] | undefined'.
  Type '{ onRenderHtml: "import:vike-react/renderer/onRenderHtml"; onRenderClient: "import:vike-react/renderer/onRenderClient"; passToClient: string[]; clientRouting: true; hydrationCanBeAborted: true; meta: { ...; }; }' is not assignable to type 'ConfigBuiltIn & Config & ConfigVikeSvelte'.
    Type '{ onRenderHtml: "import:vike-react/renderer/onRenderHtml"; onRenderClient: "import:vike-react/renderer/onRenderClient"; passToClient: string[]; clientRouting: true; hydrationCanBeAborted: true; meta: { ...; }; }' is not assignable to type 'ConfigBuiltIn'.
      Types of property 'meta' are incompatible.
        Type '{ Head: { env: "server-only"; }; Layout: { env: "server-and-client"; }; title: { env: "server-and-client"; }; description: { env: "server-only"; }; favicon: { env: "server-only"; }; lang: { env: "server-only"; }; ssr: { ...; }; }' is not assignable to type '`import:${string}` | ConfigMeta | undefined'.
          Type '{ Head: { env: "server-only"; }; Layout: { env: "server-and-client"; }; title: { env: "server-and-client"; }; description: { env: "server-only"; }; favicon: { env: "server-only"; }; lang: { env: "server-only"; }; ssr: { ...; }; }' is not assignable to type 'ConfigMeta'.
            Property 'ssr' is incompatible with index signature.
              Type '{ env: "config-only"; effect: ConfigEffect; }' is not assignable to type 'ConfigDefinition'.
                The types returned by 'effect(...)' are incompatible between these types.
                  Type 'import("/home/lourot/Documents/git/vike/vite-plugin-ssr/node_modules/.pnpm/vite-plugin-ssr@0.4.142_vite@4.3.9/node_modules/vite-plugin-ssr/dist/esm/shared/page-configs/Config").Config | undefined' is not assignable to type 'import("/home/lourot/Documents/git/vike/vite-plugin-ssr/vike/dist/esm/shared/page-configs/Config").Config | undefined'.
                    Type 'ConfigBuiltIn & Config & ConfigVikeReact' is not assignable to type 'Config | undefined'.
                      Type 'import("/home/lourot/Documents/git/vike/vite-plugin-ssr/node_modules/.pnpm/vite-plugin-ssr@0.4.142_vite@4.3.9/node_modules/vite-plugin-ssr/dist/esm/shared/page-configs/Config").ConfigBuiltIn & Vike.Config & VikePackages.ConfigVikeReact' is not assignable to type 'import("/home/lourot/Documents/git/vike/vite-plugin-ssr/vike/dist/esm/shared/page-configs/Config").ConfigBuiltIn & Vike.Config & VikePackages.ConfigVikeReact'.
                        Type 'ConfigBuiltIn & Config & ConfigVikeReact' is not assignable to type 'ConfigBuiltIn'.
                          Types of property 'extends' are incompatible.
                            Type 'Config | `import:${string}` | `import:${string}`[] | Config[] | undefined' is not assignable to type 'Config | `import:${string}` | Config[] | `import:${string}`[] | undefined'.
                              Type 'ConfigBuiltIn & Config & ConfigVikeReact' is not assignable to type 'Config | `import:${string}` | Config[] | `import:${string}`[] | undefined'.
```

I'm pretty sure this is because the vike auth example +config.h.ts includes the new vike Config that I have touched, but also wants to satisty vike-react's Config, which wants to satisfy the old vike Config. In other words there is a kind of circular type dep now accross vike and vike-react.

@brillout I believe we should tackle this circular dep first as this will hit us back whenever we touch the Config type. What do you think? Does that make sense to you if I start looking into this?

brillout:
Yes, exactly.

If the user always extends `PageContext{Client,Server}` by using `Vike.PageContext` then I'm thinking we can even remove the generic type.

Btw. for `onBeforeRender()` specifically, since it's run on the server-side only, we should use `PageContextServer` instead of `PageContext`.

An interesting question is what happens when the user sets `onBeforeRender`'s env to `server-and-client`? So far, I believe the only solution would be to code-gen types.

AurelienLourot:
> If the user always extends `PageContext{Client,Server}` by using `Vike.PageContext` then I'm thinking we can even remove the generic type.

You made me realize I wanted to use `Vike.PageContext` and not `PageContext`. I mixed up. I'll fix that and see what happens if I remove the generic.

Regarding `"vike": "link:./vike/"`, we don't write that in the examples. Probably because we want the examples to work with `npm`?

Alternatively I was thinking of removing `statisfies Config` when using `extends: vikeReact`. Rationale:

```
// examples/auth/pages/+config.h.ts

import Layout from '../layouts/LayoutDefault'
import vikeReact from 'vike-react'

export default {
  Layout,
  passToClient: ['userFullName'],
  extends: vikeReact
} // `satisfies Config` not needed, as already implied by `extends vikeReact`, which is an object
  // that already states explicitly `satisfies Config`
```

What do you think?

> Actually, since `vike` is a peer dependency, it should resolve to the same/local `vike` package 🤔.

In this example actually `vike` is also a normal dependency. Maybe it shouldn't be. Let me play with that first. Thanks for all the pointers :)

brillout:
> Does that make sense to you if I start looking into this?

Yes. (I had a quick look and it does seem like pnpm is linking everything correctly, so I'm not sure why we get that TypeScript error.)

brillout:
> Probably because we want the examples to work with `npm`?

Yes exactly. That's why we use pnpm's overrides feature instead.

> I was thinking of removing `statisfies Config` when using `extends: vikeReact`

I believe this wouldn't type `passToClient` nor `Layout`; it would only type `vikeReact`.

> In this example actually `vike` is also a normal dependency.

The example itself declares a direct dependency to `vike` which is what I would expect. What I meant is that `vike-react` shouldn't declare a direct dependency to `vike` and its dependency to `vike` should be a peer dependency instead, which is actually already the case (I just double checked). So I believe `vike-react` and the example are doing things correctly. And since pnpm seems to link everything correctly, I wonder why that error occurs.

Let me know if you want me to dig into that TS error.

AurelienLourot:
@brillout I found the culprit: the example has a dependency to an old `vike-react` (0.2.0), which still has a dependency to `vite-plugin-ssr` instead of `vike`, so it doesn't benefit from `"vike": "link:./vike/"`. Switching to vike-react@0.3.0 solves the issue.

Thanks for the useful pointers. Updating the PR

brillout:
Ah, indeed, I forgot to update that one. Thanks 🙏.

vike/shared/page-configs/Config.ts
```diff
@@ -78,7 +81,11 @@ type ConfigBuiltIn = {
    *
    *  https://vike.dev/onBeforeRender
    */
-  onBeforeRender?: Function | ImportString | null
+  onBeforeRender?:
+    | ((pageContext: PageContextServer) => OptionalPromise<{ pageContext: Record<string, any> }>)
+    | ((pageContext: PageContextClient) => OptionalPromise<{ pageContext: Record<string, any> }>)
+    | ImportString
+    | null
```
AurelienLourot, line 88: @brillout better?

brillout: That's what I had in mind indeed. Interesting idea to have a union: I wonder whether this addresses the issue when a user changes the `env` of `onBeforeRender()` 🤔

brillout: Thinking, maybe we could even do this.

```diff
- OptionalPromise<{ pageContext: Record<string, any> }>)
+ OptionalPromise<{ pageContext: Vike.PageContext }>)
```

AurelienLourot: > I wonder whether this addresses the issue when a user changes the `env` of `onBeforeRender()`

oh am I missing something more from your previous points? I re-read them but I'm not sure which other issue there could be. The user can now freely do either

```
const onBeforeRender: Config['onBeforeRender'] = async (pageContext: PageContextServer) => {
```

or

```
const onBeforeRender: Config['onBeforeRender'] = async (pageContext: PageContextClient) => {
```

So any `env` is fine. Am I missing something?

However they can't do

```
const onBeforeRender: Config['onBeforeRender'] = async (pageContext: PageContextServer | PageContextClient) => {
```

so I need to add that to the union.



> Thinking, maybe we could even do this.
> 
> ```diff
> - OptionalPromise<{ pageContext: Record<string, any> }>)
> + OptionalPromise<{ pageContext: Vike.PageContext }>)
> ```

Let me give it a try

AurelienLourot: > However they can't do
> 
> ```
> const onBeforeRender: Config['onBeforeRender'] = async (pageContext: PageContextServer | PageContextClient) => {
> ```
> 
> so I need to add that to the union.

Actually this works already, because `PageContextClient | PageContextServer` and `PageContextClient` happen to be the same thing already, as `PageContextClient` is already `<something> | PageContextServer`. So that's covered.

> Thinking, maybe we could even do this.
> 
> ```diff
> - OptionalPromise<{ pageContext: Record<string, any> }>)
> + OptionalPromise<{ pageContext: Vike.PageContext }>)
> ```

I would have to make `Page` and `config` optional at https://github.com/vikejs/vike/blob/main/examples/react-full-v1/renderer/PageContext.ts , so `onBeforeRender()` isn't forced to put them in its return value. It doesn't feel quite right though, for example in `vike-react` we don't make `Page` optional in `Vike.PageContext`, which we use as parameter of `onRenderHtml()`. I see two options:

1. we're happy with just `Record` as it is for now
2. we go for something like

```
  namespace Vike {
    interface PageContextAdditionalData {
      title?: string
      abortReason?: string
    }
  }
```

and `Vike.PageContext` is defined based on `Vike.PageContextAdditionalData`. And `onBeforeRender()` is expected to return `Vike.PageContextAdditionalData`. What's your feeling on this?

brillout: Or how about [`Partial<Vike.PageContext>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype)?

> `Vike.PageContext` is defined based on `Vike.PageContextAdditionalData`

AFAICT it would work indeed. It's an intersting idea to differentiate between refining existing `PageContext` and extending it.

I've a slight preference for `Partial<Vike.PageContext>` because it's more minimal (feel free to disagree). I'm curious to see whether they'll be more use cases for `Vike.PageContextAdditionalData` 👀.

> `PageContextClient` is already `<something> | PageContextServer`

It's actually slightly different: conceptually `PageContextClient` needs to be `Partial<PageContextServer>` because of `passToClient`. I've also made changes recently that increases the difference. So I think we do need the union `PageContextClient | PageContextServer`. (We can also use `PageContext` as `PageContext = PageContextClient | PageContextServer`.)

AFAICT everything will work then. Very neat.

One little drawback is that if the user doesn't type his hook:

```ts
const onBeforeRender: Config['onBeforeRender'] = async (pageContext) => {
```

then, I guess, TypeScript will set `pageContext` to `PageContextClient | PageContextServer`? Which doesn't align with the default `env` that is `server-only`. I guess we can tell users to set `pageContext` to `PageContextServer`. I can't think of a solution that sets `pageContext` to `PageContextServer` by default while allowing users to set it to `PageContextServer | PageContextClient` 🤔.

Let me know if something's not clear in my response!

AurelienLourot: I didn't know `Partial<>`, all clear, thanks a lot :pray: I'll update the PR.

vike/shared/page-configs/Config.ts
```diff
@@ -78,7 +81,12 @@ type ConfigBuiltIn = {
    *
    *  https://vike.dev/onBeforeRender
    */
-  onBeforeRender?: Function | ImportString | null
+  onBeforeRender?:
+    | ((pageContext: PageContextServer) => OptionalPromise<{ pageContext: Partial<Vike.PageContext> }>)
+    | ((pageContext: PageContextClient) => OptionalPromise<{ pageContext: Partial<Vike.PageContext> }>)
+    | ((pageContext: PageContext) => OptionalPromise<{ pageContext: Partial<Vike.PageContext> }>)
+    | ImportString
+    | null
```
AurelienLourot, line 89: @brillout should be better now. Let me know if not clear why I need three signatures. I can illustrate with a TS Playground if needed

brillout: 👍 I think we're on the same page.

I'm seeing that the user *has* to type his `onBeforeRender()` hook because `Config['onBeforeRender']` is a union. In other words, it seems that, if `Config['onBeforeRender']` has only one function type then TypeScript will infer the type of `onBeforeRender()`.

That's quite a bummer as I was hoping that all the user had to do was to set `const onBeforeRender: Config['onBeforeRender'] = ...` and TypeScript would infer the `pageContext` and return type.

I can't think of a solution 🤔.

Anyways, it's better than `Function` as it ensures the user doesn't use wrong types.

AurelienLourot: Yes and there is another reason why the type should be set explicitly on `onBeforeRender()`, which is because of some otherwise-non-strictness in the return types, which I illustrate and explain [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApwOYQMIHtwQAeYyA3gFDJqYQBcyAzmFKBpcgvjMBvUyyDZUwwMABsIAfj7NW5AL7lyYAJ4AHFAHkQAIQgxcUCACUIIACbQAYgFcQCEfmQBeZAAo1NPAWL10Wb0hiAEoXAD4yZE8A-CCwPzgoETgxAB5-HFiiMAjFck4QJmR8PQMjUwtoAEZ6bVLDEzNLKFt7RxAXd1DnCIoqIzAbKA6+qiivLN8ydjHkEXE6ZAAiAFsVABVRCSWAGiVZqgB6Q+R1gAtgBmQAd0MAayuAIwgEOBsGFFwYZBB8AFp+MAHCAIAwrqBkAMhh1VBobmczMg4GCeCA5AdkMckcgYHYHMAnLCUGBcEjfmAEVAcXj2nN1BAZmNHmI4PQlszcI8lozFFRFHkCkUSvoGhVmgAmWq6EXlJrWGkEjquNzBBJJYApdITHw5cLTfoQQbDfVjaKZHX0UazeYSNlrTYLXb7A5YgByuGucwuVxYGDOYFxYjEKhxcGAYgYOy9cBADzmpIpKCIajEQNEkMN0LpGkZVGZrOWHK5PPY-PIQA). Let me know if unclear :) Otherwise I'll now keep extending this PR with other hook types.

examples/react-full-v1/renderer/+onRenderHtml.tsx
```diff
@@ -34,10 +34,13 @@ async function onRenderHtml(pageContext: PageContextServer) {
   return {
     documentHtml,
     // See https://vike.dev/stream#initial-data-after-stream-end
+    // temp(aurelien): interesting that this can be an async function. Can this be the case in
+    // onBeforeRender() as well? Or is this specific to onRenderHtml()?
```
AurelienLourot, line 38: @brillout see inline question, thanks!

examples/path-aliases-v1/renderer/onRenderHtml.tsx
```diff
@@ -5,10 +5,11 @@ import ReactDOMServer from 'react-dom/server'
...
 import { escapeInject, dangerouslySkipEscape } from 'vike/server'
 import { PageLayout } from './PageLayout'
-import type { PageContext } from '#root/types'
+import { ReactComponent } from '#root/types'
+import type { PageContextServer } from 'vike/types'
 
-function onRenderHtml(pageContext: PageContext) {
-  const { Page } = pageContext
+function onRenderHtml(pageContext: PageContextServer) {
+  const Page = pageContext.Page as ReactComponent
```
AurelienLourot, line 12: @brillout so I'm forced by TypeScript now to use `PageContextServer` in the signature, which is what we wanted. However it turns out that `PageContextServer['Page']` is of type `unknown`, thus the `as ReactComponent`. This is not elegant, to force the user to do that. Any thoughts?

AurelienLourot: Nevermind, I realize this is what interface merging is for:

```
declare global {
  namespace Vike {
    interface PageContext {
      Page: ReactComponent
    }
  }
}
```

I'll push a new commit

vike/shared/page-configs/Config.ts
brillout: Yea I've seen that as well. ~~As much as I wish TypeScript would infer the hook type, it doesn't seem like there is a way to do it that preserves correctness.~~ Actually I thought a bit more about this.

I'm thinking: maybe we can remove the union by setting the onBeforeRender's pageContext to `PageContextServer` while enabling users to change the type by using the global interface `Vike.Config`. Thoughts? Let me know if you feel like it's out of your PR's scope.

examples/path-aliases-v1/renderer/onRenderHtml.tsx
brillout: Yes, exactly. (It's actually on my to-do list to update the examples and boilerplates to use `Vike.PageContext` instead of `PageContextBuiltIn{Server,Client}`.)

vike/shared/page-configs/Config.ts
brillout: > which I illustrate and explain [here](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApwOYQMIHtwQAeYyA3gFDJqYQBcyAzmFKBpcgvjMBvUyyDZUwwMABsIAfj7NW5AL7lyYAJ4AHFAHkQAIQgxcUCACUIIACbQAYgFcQCEfmQBeZAAo1NPAWL10Wb0hiAEoXAD4yZE8A-CCwPzgoETgxAB5-HFiiMAjFck4QJmR8PQMjUwtoHThzem1SwxMzSyhbe0cQF3dQ5wiKKiMwGyhO-qooryzfMnZx5BFxOmQAIgBbFQAVUQllgBolOaoAeiPkDYALYAZkAHdDAGtrgCMIBDgbBhRcGGQQfABafjABwgCAMa6gZCDYadVQaW7nMzIODgnggOSHZAnZHIGB2BzAJxwlBgXDIv5gRFQXH4jrzdQQWbjJ5iOD0ZYs3BPZZMxRURR5ApFEr6RoVFoAcVwuFqyHqovKzWstMJnVcbmC9FIExiPni1CSwBS6UmerC8nCMwGECGIyt42imT1WqZwm2SzWm3dewOh2xADlcDd5pdriwMOcwHixGIVLi4MAxAxdiG4CBHvMyZSUEQ1GJgaIoTaYfSNK7kCy2StOdzeewBeQgA).

I'm also seeing that `satisfies` isn't strict as well. Also, even setting the return type isn't strict, which I find quite surprising (see `badAsWell: false`.) I udated your playground to showcase what I mean [here](https://www.typescriptlang.org/play?ssl=32&ssc=1&pln=33&pc=1#code/JYOwLgpgTgZghgYwgAgApwOYQMIHtwQAeYyA3gFDJqYQBcyAzmFKBpcgvjMBvUyyDZUwwMABsIAfj7NW5AL7lyYAJ4AHFAHkQAIQgxcUCACUIIACbQAYgFcQCEfmQBeZAAo1NPAWL10Wb0hiAEoXAD4yZE8A-CCwPzgoETgxAB5-HFiiMAjFck4QJmR8PQMjUwtoHThzem1SwxMzSyhbe0cQF3dQ5wiKKiMwGyhO-qooryzfMnZx5BFxOmQAIgBbFQAVUQllgBolOaoAeiPkDYALYAZkAHdDAGtrgCMIBDgbBhRcGGQQfABafjABwgCAMa6gZCDYadVQaW7nMzIODgnggOSHZAnZHIGB2BzAJxwlBgXDIv5gRFQXH4jrzdQQWbjJ5iOD0ZYs3BPZZMxRURR5ApFEr6RoVFoAcVwuFqyHqovKzWstMJnVcbmC9FIExiPni1CSwBS6UmerC8nCMwGECGIyt42imT1WqZwm2SzWm3dewOh2xADlcDd5pdriwMOcwHixGIVLi4MAxAxdiG4CBHvMyZSUEQ1GJgaIoTaYfSNK7kCy2StOdzefs5k8agBBBgAdQgMfo8CTjP5CiU+Xwwt0CqalSg1XMACYum4NZaxtC7WMqI7AtkXZiFhJ2estosfeXsRcrrcHs9Xu9PsUfn8QIDZCCwRDOkvYQyEUiUQw0Ri-ac4BpdpVVLEkyTTXBs2pPFgKJBly0rdkax5OZ5HrZlmzbDsxHoZgbF7ZABVCBg4BEBhuDBOURzKMcWjaAl8CAA).

examples/react-full-v1/renderer/+onRenderHtml.tsx
brillout: It's only implemented for `onRenderHtml()`. FYI the user case is this: https://vike.dev/stream#initial-data-after-stream-end.

AurelienLourot: Perfect, thanks

vike/shared/page-configs/Config.ts
AurelienLourot: > I'm also seeing that satisfies isn't strict as well.

Interesting. It seems to be another symptom of the same problem. More on this here:
* https://stackoverflow.com/questions/63006497/typescript-why-not-strict-return-type
* https://stackoverflow.com/questions/57214325/typescript-excess-properties-are-not-checked-in-lambdas-without-explicit-return

> I'm thinking: maybe we can remove the union by setting the onBeforeRender's pageContext to PageContextServer

That sounds good indeed. I'll give that a try.

examples/path-aliases-v1/renderer/onRenderHtml.tsx
```diff
@@ -1,13 +1,13 @@
...
 import ReactDOMServer from 'react-dom/server'
 import React from 'react'
 import { escapeInject, dangerouslySkipEscape } from 'vike/server'
 import { PageLayout } from './PageLayout'
-import type { PageContext } from '#root/types'
+import type { PageContextServer } from '#root/types'
+import type { Config, DocumentHtml } from 'vike/types'
 
-function onRenderHtml(pageContext: PageContext) {
+const onRenderHtml: Config['onRenderHtml'] = (pageContext: PageContextServer): DocumentHtml => {
```
AurelienLourot, line 10: @brillout fun fact: I suspect we are benefiting here from the return-type flexibility.

Here I am defining a function of type

```
(pageContext: PageContextServer) => DocumentHtml
```

which ends up satisfying / being assignable to

```
(pageContext: PageContextServer) => DocumentHtml | { documentHtml: DocumentHtml; pageContext?: Partial<Vike.PageContext> | Function }
```

vike/shared/page-configs/Config.ts
AurelienLourot: > If the user always extends `PageContext{Client,Server}` by using `Vike.PageContext` then I'm thinking we can even remove the generic type.

And now this sentence just clicked. I didn't fully understand it before. We're on the same page :+1:

brillout: 👍 Let me know if you have other questions.

AurelienLourot:
@brillout I'm happy with this now. Let me know if anything should be improved. Thanks!

brillout:
👍 Thinking out loud:
 - The type of `react-full-v1/pages/star-wars/@id/+onBeforeRender.tsx` is actually `(pageContext: PageContextServer | PageContextClient) => /* ... */` because of `/pages/star-wars/@id/+onBeforeRenderIsomorph.ts`. I think we can make it work by telling users that, if they modify `meta['onBeforeRender']`, then they should use `Vike.Config` to refine `Config['onBeforeRender']`.
 - It seems that `const onBeforeRender: Config['onBeforeRender'] = async (pageContext: PageContextServer): Promise<{ pageContext: Partial<Vike.PageContext> }>` is indeed the most correct/precise. I wonder if we can make it less verbose, but I guess we can't if we want to be strict about the return type. Which makes me wonder if the verbosity is worth the return type safety. Maybe I'm exaggerating but I imagine it to be tedious for the user to have to define 3 types for each hook (return type, argument type, and overall function type). I guess it's a trade-off situation. Thoughts? 
   - An idea: `const onBeforeRender: Config['onBeforeRender'] = async (pageContext: Parameters<Config['onBeforeRender']>[0]): ReturnType<Config['onBeforeRender']>` (the types `Paramaters` and `ReturnType` are utilities provided by TypeScript like `Partial`). I've actually just tried (see my / the latest commit to this PR): turns out it isn't as simple as that. But maybe we can make it simpler by having Vike providing some utility type (I can further dig later this week if you want me to).
 - The return type `Partial<Vike.PageContext>` seems to be working well, as I can't think of a use case where a user would override a built-in `pageContext` property https://vike.dev/pageContext#built-in.
 - `route()` (and `onBeforeRoute()`) is also called on the client-side if the user uses Client Routing => `pageContext: PageContextServer | PageContextClient`. Since we'll increasingly promote the `vike-*` packages, I think it's fine if we assume the user to use Client Routing (Server Routing users would need to use their own types).
 - `guard()` can also be called on the client-side.

Very much looking forward to provide users with type safe `Config['somHook']` 💫

examples/react-full-v1/pages/star-wars/index/+onBeforeRender.ts
```diff
@@ -1,11 +1,13 @@
...
 import { filterMoviesData, getStarWarsMovies, getTitle } from './getStarWarsMovies'
 
-const onBeforeRender: Config['onBeforeRender'] = async (
-  pageContext: PageContextServer
-): Promise<{ pageContext: Partial<Vike.PageContext> }> => {
+type OnBeforeRender = Extract<Config['onBeforeRender'], Function>
+
+const onBeforeRender: OnBeforeRender = async (
+  pageContext: Parameters<OnBeforeRender>[0]
+): Extract<ReturnType<OnBeforeRender>, Promise<any>> => {
```
AurelienLourot, line 10: My personal preference as a user would be to read just this

```
const onBeforeRender = async (
  pageContext: PageContextServer
): Promise<{ pageContext: Partial<Vike.PageContext> }> => {
```

Because if I need to modify the implementation of the hook, it tells me super clearly what's available in the parameters and what I need to return. For me, having the information I need in the code exactly where I need it has even more value than the actual safety provided by TypeScript.

But it's my personal taste so what about this: we provide various styles across examples:
* in some examples we have no explicit types (yes, not fully safe)
* in some examples we use only `Config['...']` and no other explicit types (yes, not fully safe)
* in some examples we use the style I prefer
* in some examples we use `Extract<>`

And then it's up to the user to use the style they like, if they prefer safety, or readability, or conciseness. I could even write documentation on vike.dev about this. What do you think?

brillout: Ok that's a good point: from a readability perspective, it is indeed much better.

> documentation on vike.dev 

Agreed: seems paramount to me to share with the user the knowledge we found out in this PR.

I'm actually tempted to make the verbose version our default choice, as I think it's a very good point that the readability of it is very neat. Maybe I(/we) can mull over it a bit more and then take a decision, but let me know if you feel blocked.

brillout: > I could even write

(👍)

AurelienLourot: > * But maybe we can make it simpler by having Vike providing some utility type (I can further dig later this week if you want me to).

Yes, for having some kind of simpler `Extract<>`-like utilities. I'll dig into that too but again that wouldn't be as a user/reader my favorite style, because it doesn't tell me what comes in and out of the function in simple, straightforward terms. If that's your favorite style though, I still think it would be cool to show/document different styles

AurelienLourot: > Maybe I(/we) can mull over it a bit more and then take a decision

no rush. I'll still want to push a few things to this PR regarding your other points, e.g. `PageContextServer | PageContextClient`, and if you need to think a bit more about it, take your time :pray:

examples/react-full-v1/pages/star-wars/index/+onBeforeRender.ts
```diff
@@ -1,13 +1,11 @@
...
 import { filterMoviesData, getStarWarsMovies, getTitle } from './getStarWarsMovies'
 
-type OnBeforeRender = Extract<Config['onBeforeRender'], Function>
-
-const onBeforeRender: OnBeforeRender = async (
-  pageContext: Parameters<OnBeforeRender>[0]
-): Extract<ReturnType<OnBeforeRender>, Promise<any>> => {
+const onBeforeRender: Hook<'onBeforeRender'> = async (
+  pageContext: HookParam<'onBeforeRender', 0>
+): HookReturnTypeAsync<'onBeforeRender'> => {
```
AurelienLourot, line 8: @brillout I came up with this

examples/react-full-v1/pages/star-wars/@id/+onBeforeRender.tsx
```diff
@@ -2,15 +2,15 @@
...
 import { filterMovieData } from '../filterMovieData'
-import type { Config, PageContextServer } from 'vike/types'
+import type { Config, PageContextClient, PageContextServer } from 'vike/types'
 import type { MovieDetails } from '../types'
 import { render } from 'vike/abort'
 import React from 'react'
 
 const onBeforeRender: Config['onBeforeRender'] = async (
-  pageContext: PageContextServer
+  pageContext: PageContextServer | PageContextClient
```
AurelienLourot, line 11: @brillout we touched on this in the past but for some reason, this just works although `Config['onBeforeRender']` is defined to accept only `PageContextServer`. Do you want me to dig more?

AurelienLourot:
> * The return type `Partial<Vike.PageContext>` seems to be working well, as I can't think of a use case where a user would override a built-in `pageContext` property https://vike.dev/pageContext#built-in.

Not sure I get this point. Is this about a particular line of code? Do you want me to change something?

examples/react-full-v1/pages/star-wars/@id/+onBeforeRender.tsx
brillout: My (wild) guess/intuition is that it's an inversion of control thing: Vike calls the hook and always provides `pageContext: PageContextServer`, which means that if the user defines a hook that accepts `pageContext: PageContextServer | SomethingElse` then that's fine: everything will work. Maybe that's what TypeScript means with "function type A is assignable to function type B".

AurelienLourot: [You seem to be right](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApwOYQMIHtwQAeYAytAG7TIDeAUMmphAFzIDOYUoG9yC+MYBlYcuIHgxi5cIzt1oBfWrVCRYiFOix4CxbABtgEcDV5aW7OeN78Qg4ZbETkAIzhRZTxcrABPAA4oAPIgAEIQUlAQAErGACbQAGIAriAIYMD4yAC8yAAU-kw6kMSs5sVEpBTQAJQ5AHzI5LjAccq2HMj44ZEx8dCsIT24UbEgCVApaRlZuQVF+CVgZQu6VVCUUMgAPozai5UGRuA1rM2tDTRKQA).

So is it fine for you as it is now? I.e. `Config['onBeforeRender']` is defined to accept only `PageContextServer` but the user can still define a hook accepting `PageContextServer | PageContextClient`. I would cover that in the documentation.

brillout: Sound good 👍

AurelienLourot:
@brillout I think I addressed everything. I'll give a try at documenting now all this, either in this PR or in another one.

brillout:
I just woke up from a nap and got an idea, I hope it's ok if I share it with you. I don't know if it bothers you but I'm thinking it's worth it to take our time on this one. Let me know if it does bother you.

The idea:

```diff
- Hook<'onBeforeRender'>
- HookParam<'onBeforeRender'>
- HookReturn<'onBeforeRender'>
+ OnBeforeRender
+ OnBeforeRenderParam
+ OnBeforeRenderReturn
```

Why? Because, that way, the user can go-to-defintion:

```ts
// This is the Vike source code the user sees when he goes-to-the-definition of
// OnBeforeRender{Param,Return}
type OnBeforeRenderParam = PageContextServer
type OnBeforeRenderReturn = { pageContext?: Partial<Vike.PageContext> } | undefined
type OnBeforeRender = (pageContext: OnBeforeRenderParam) => OnBeforeRenderReturn
```

Whereas going-to-the-definition-of `Hook{Param,Return}` isn't helpful.

One drawback of `OnBeforeRender{Param,Return}` is that the source of truth isn't `Config` anymore, which means that the user cannot use `Vike.Config` to override the hook types. (Whereas with `Hook{Param,Return}` the source of truth is `Config` => if the user modifies `Config` by using `Vike.Config` then his changes will be reflected by `Hook{Param,Return}`.) I think a semi-satisfactory solution would be to tell users to not use `OnBeforeRender{Param,Return}` if they change the `meta` of `onBeforeRender`.

What do you think? Let me know if something is unclear.

Also, some random thoughts (clearly out-of-scope for this PR): I'm thinking that `Vike.PageContextCustom` could have been a better name than `Vike.PageContext` (which, I guess, is similar to what you had in mind back then with `Vike.PageContextAdditionalData`). Let's see how things turn out.

Thanks for the conversations btw. Your input has been very helpful. While it does delay the V1 design release, I think it's very well worth it: providing a neat TypeScript DX is super convenient for the user.

> I'll give a try at documenting now all this, either in this PR or in another one.

I'm thinking, if we provide a clear path for users, then we may not need to teach him TypeScript knowledge at all. (Saving us time while lifting the user from the need for deep TypeScript knowledge: I can imagine that some/(many?) users don't want to dig into TypeScript.) That said, in the future, some extra docs explaining TypeScript could be really neat. So I agree: we can keep the docs minimal for now and see when/whether we'll need it.

AurelienLourot:
No worries, you're right it's important to get this right. And it is interesting for me.

Regarding `OnBeforeRender{Param,Return}` it's a tiny bit more complicated but I have just pushed a minimum working commit: https://github.com/vikejs/vike/pull/1137/commits/2213c796 . I have added a bunch of inline `NOTE(aurelien)` as prompts for discussion, and I will eventually remove them. Let me know what you think.

Once we're happy with this part of the story, I'll reply on the other topics :) Namely:
* `Vike.PageContextCustom`
* documentation

examples/react-full-v1/pages/star-wars/index/+onBeforeRender.ts
```diff
@@ -1,11 +1,12 @@
...
+import type { OnBeforeRender, OnBeforeRenderParam, OnBeforeRenderReturn } from 'vike/types'
 import { filterMoviesData, getStarWarsMovies, getTitle } from './getStarWarsMovies'
 
-const onBeforeRender: Hook<'onBeforeRender'> = async (
-  pageContext: HookParam<'onBeforeRender'>
-): HookReturn<'onBeforeRender'> => {
+// NOTE(aurelien): that's the alternative style, still with full safety. I'm wondering if we should
+// keep `Config['onBeforeRender']` instead of `OnBeforeRender` in order for the user to easily modify
+// `Config` through `Vike.Config` without unexpected results?
+const onBeforeRender: OnBeforeRender = async (pageContext: OnBeforeRenderParam): OnBeforeRenderReturn => {
```
AurelienLourot, line 9: The other reasons why this style isn't perfect are:
1. If that hook has to run on client too, you probably don't want to use `OnBeforeRenderParam` but `PageContextServer | PageContextClient` instead.
2. The direct definition of `OnBeforeRenderReturn` is still cryptic as it is `Extract<OnBeforeRenderReturnType, Promise<any>>`.

vike/shared/page-configs/Config.ts
```diff
@@ -56,6 +67,18 @@ type HookReturnSync<Name extends keyof Config> = Exclude<HookReturnType<Name>, P
...
+// NOTE(aurelien): to be extended. I'm wondering if this is all worth it as I'm wondering if people
+// will/should really use this style.
+type OnBeforeRenderParam = PageContextServer
+// NOTE(aurelien): unfortunately the user can't use directly `OnBeforeRenderReturnType` but needs
+// to explicitly select either the Async or the Sync version.
+type OnBeforeRenderReturnType = OptionalPromise<{ pageContext: Partial<Vike.PageContext> } | void>
+type OnBeforeRenderReturnAsync = Extract<OnBeforeRenderReturnType, Promise<any>>
+type OnBeforeRenderReturnSync = Exclude<OnBeforeRenderReturnType, Promise<any>>
+// NOTE(aurelien): but as discussed we still provide a simpler name for Async as default.
+type OnBeforeRenderReturn = OnBeforeRenderReturnAsync
```
AurelienLourot, line 79: In order not to have a cryptic direct definition of `OnBeforeRenderReturn` for the user, maybe I should write instead something like this:

```ts
type OnBeforeRenderReturnType = OptionalPromise<{ pageContext: Partial<Vike.PageContext> } | void>
type OnBeforeRenderReturnAsync = Promise<{ pageContext: Partial<Vike.PageContext> } | void>
type OnBeforeRenderReturnSync = { pageContext: Partial<Vike.PageContext> } | void
type OnBeforeRenderReturn = Promise<{ pageContext: Partial<Vike.PageContext> } | void>
```

brillout: I had the same idea :)

examples/react-full-v1/pages/star-wars/index/+onBeforeRender.ts
brillout: My idea was this:

> I think a semi-satisfactory solution would be to tell users to not use `OnBeforeRender{Param,Return}` if they change the `meta` of `onBeforeRender`.

The `meta` feature is mostly meant for `vike-*` packages and, so far, I expect only few users to directly use the `meta` feature. That's why I'm thinking my semi-satisfactory proposal to be viable.

brillout:
> `NOTE(aurelien)`

Good idea 👍 and I started doing the same.

> Once we're happy with this part of the story, I'll reply on the other topics :) Namely:

👍 Sounds like a good plan.

brillout:
> ```ts
> // NOTE(aurelien): this is still my favorite style as a user. And as a Vike contributor I'll
> // probably personally advertise for this.
> const onBeforeRender: Config['onBeforeRender'] = async (
>   pageContext: PageContextServer | PageContextClient
> ): Promise<{ pageContext: Partial<Vike.PageContext> }>
> ```

One thing I like about `OnBeforeRender{Param,Return}` is that it's simple for the user to set up: the user merely has to remember the `${HookName}{Param,Return}` name convention. Whereas with the expanded style, the user has to go to `https://vike.dev/onBeforeRender#typescript` and copy-pase the hook type.

Also with `OnBeforeRender{Param,Return}` it's obvious that the hook type corresponds to the offical type and is therefore correct, whereas with the expanded style this isn't clear.

Another neat thing is that we can add JSDocs to `OnBeforeRender{Param,Return}`:

Screenshot 2023-10-06 10 50 55
Screenshot 2023-10-06 10 51 00

The screenshots also shows that the type is shown like in the expanded style.

Although, unfortunately, this doesn't work for `OnBeforeRenderParam`:

image

I've created a TypeScript Feature Request about it: https://github.com/microsoft/TypeScript/issues/56010.

That said, from a purely reading perspective (without TypeScript's quick info IntelliSense for example in a Pull Request review), the expanded style is indeed superior.

brillout:
I'm thinking maybe we could/should forgo `OnBeforeRenderParam` while keeping `OnBeforeRender{Return}`:

```ts
const onBeforeRender: OnBeforeRender = async (pageContext: PageContextServer): OnBeforeRenderReturn
```

examples/react-full-v1/pages/star-wars/@id/+onBeforeRender.tsx
```diff
@@ -1,15 +1,16 @@
...
+import type { OnBeforeRender, OnBeforeRenderReturn, PageContextClient, PageContextServer } from 'vike/types'
 import type { MovieDetails } from '../types'
 import { render } from 'vike/abort'
 import React from 'react'
 
-async function onBeforeRender(pageContext: PageContextBuiltInServer) {
-  const dataUrl = `https://star-wars.brillout.com/api/films/${pageContext.routeParams.id}.json`
+const onBeforeRender: OnBeforeRender = async (
+  pageContext: PageContextServer | PageContextClient
+): OnBeforeRenderReturn => {
```
AurelienLourot, line 12: @brillout is this what you had in mind? If yes I'd go ahead and make change all the hooks to this pattern

AurelienLourot (commented):
> `Vike.PageContextCustom`

Yes I like this name more. Should we work on a renaming/alias?

examples/react-full-v1/pages/star-wars/@id/+onBeforeRender.tsx
brillout: Yes, although I'm still a little bit hesitating between `pageContext: PageContextServer | PageContextClient` and  `pageContext: OnBeforeRenderParam`.

What do you think? 👀

brillout:
> > `Vike.PageContextCustom`
> 
> Yes I like this name more. Should we work on a renaming/alias?

I'm not sure about this yet, so I'd suggest we first finish this PR and then, at some point later, we revisit whether `Vike.PageContextCustom` turns out to be indeed better. (As always, feel free to disagree.)

examples/react-full-v1/pages/star-wars/@id/+onBeforeRender.tsx
brillout: Let me think a little bit more about it; I'll circle back on this tentatively later today.

> I'm not a big fan of the entire `OnBeforeRender*` set of types

(We can further discuss this if you want, as you wish.)

brillout: Alright, I'm very happy with 7bc32ed79c5a50f5b081708a26fd1a5265c58ae6. While we tell users:

> We recommend to always use `ReturnType<OnBeforeRender>` which is needed for making TypeScript strict about return types, see [some-ts-playground-link] and [some-external-link-explaining-it].

I feel like this nails it. 

As for the expanded styles, users can do it if they want to although, IMO, I wouldn't encourage it nor mention it in the docs: I think IntelliSense's QuickInfo is more than enough.

Screenshot 2023-10-10 11 49 25

image

Let me know if you have objections (I'm more than open to further discuss things if you us want to).

AurelienLourot: Ah well done, this looks great now. Doesn't feel clunky anymore. Let me go ahead and apply that to all hooks

brillout: 👍 I'm glad you like it as well.

Let me know if repeating the JSDocs gets too annoying, I think we can write a little macro script and append it to `vike/dist-cjs-fixup.mjs` (I guess we should rename it to `vike/build-fixup.js` then).

AurelienLourot: It might be a bit of an overkill, especially because the duplication is not spread across different files

brillout:
FYI https://github.com/vikejs/vike/pull/1137/commits/b577406b1c23b895d4e4ae2813567bc22de6f315: I pushed a change to `main` so that hooks can be exported as `export { hookName }` instead of `export default`.

brillout:
> I pushed a change to `main`

The CI is failing because there is a bug in the changes I pushed. I'll fix it tomorrow.

AurelienLourot:
> FYI [b577406](https://github.com/vikejs/vike/commit/b577406b1c23b895d4e4ae2813567bc22de6f315): I pushed a change to `main` so that hooks can be exported as `export { hookName }` instead of `export default`.

Nice!

brillout:
> I'll fix it tomorrow.

Done. The CI is now green.

AurelienLourot:
@brillout I think it looks good now. What do you think?

Should I work on adding all this to the documentation or do you think we can get away with not documenting this for now?

brillout:
👍 I'm seeing you added the type utils for the hooks 💯.

Async `route` is actually deprecated, so the default should be sync. ~~I also think the default for `onPageTransition{Start,End}` hooks should be sync~~ `onPageTransition{Start,End}` hooks don't return any value so I think we may not need any (a)sync variant. Same for `onHydrationEnd`.

I'm thinking whethter other hooks should be sync by default. E.g. I expect most `guard()` hooks to be sync. Alternative: we always use async `guard()` hooks in the docs and examples. Any thoughts on this? (I'm thinking it could be nice to always use async hook when a hook can be async: it's then obivious that the hook can be async.)

brillout:
I just tried and this works:

```ts
type OnHydrationEnd = (pageContext: PageContextClient) => Promise<void | undefined> | void | undefined;
```

So we indeed can remove the variants for hooks that return `void | Promise<void>`.

On the other hand, it's kinda nice to keep the convention `HookName{Sync}` consistent across all hooks.

Hm 🤔

AurelienLourot:
> Async `route` is actually deprecated, so the default should be sync. ~I also think the default for `onPageTransition{Start,End}` hooks should be sync~ `onPageTransition{Start,End}` hooks don't return any value so I think we may not need any (a)sync variant. Same for `onHydrationEnd`.

Make sense, yes

----------

> I'm thinking it could be nice to always use async hook when a hook can be async: it's then obivious that the hook can be async.

Yes I like this

----------

> ```ts
> type OnHydrationEnd = (pageContext: PageContextClient) => Promise<void | undefined> | void | undefined;
> ```

The problem with this is that we can't use `ReturnType<OnHydrationEnd>` anymore:

```ts
// error TS1064: The return type of an async function or method must be the global Promise<T> type. Did you mean to write 'Promise<void>'?
const onHydrationEnd: OnHydrationEnd = async (): ReturnType<OnHydrationEnd> => {
  console.log('Hydration finished; page is now interactive.')
}
```

This is why we stopped mixing `Promise<something> | something` in return-types. We could say it's OK not to support `ReturnType<HookThatReturnsVoid>` but I would prefer just providing the async variant instead.

examples/vue-full-v1/renderer/+config.h.ts
```diff
@@ -6,7 +6,7 @@ export default {
   passToClient: ['pageProps', 'title'],
   clientRouting: true,
   prefetchStaticAssets: 'viewport',
-  onHydrationEnd,
+  onHydrationEnd, // NOTE(aurelien): this assignment breaks if we stop supporting the sync variant
```
AurelienLourot, line 9: See inline comment. It seems like we're introducing a breaking change by not supporting the sync variant of void-returning hooks. So I think we shouldn't do that either.

My suggestion is that we keep `OnHydrationEnd | OnHydrationEndSync`, yet from now on, as you said, we try to make examples use async variants as much as we can. What do you think?

brillout:
Ah, you're right, `The return type of an async function or method must be the global Promise<T> type` was the issue and I just checked: it's also problematic even for a return type `void`.

Ok let's go with your with your suggestion. In other words, looking at the list:

```js
  Guard,
  GuardSync,
  OnBeforePrerenderStart,
  OnBeforePrerenderStartSync,
  OnBeforeRender,
  OnBeforeRenderSync,
  OnBeforeRoute,
  OnBeforeRouteSync,
  OnHydrationEnd,
  OnPageTransitionEnd,
  OnPageTransitionStart,
  OnPrerenderStart,
  OnPrerenderStartSync,
  OnRenderClient,
  OnRenderClientSync,
  OnRenderHtml,
  OnRenderHtmlSync,
  Route
```

Only `Route` should be sync by default (i.e. `Route | RouteAsync`). All other hooks should be `HookName` | `HookNameSync` (async by default).

> we try to make examples use async variants as much as we can. What do you think?

Agreed. I think it's easier for the user if we keep things consistent.

brillout:
Idea: we provide `OnBeforeRenderAsync | OnBeforeRenderSync` instead of `OnBeforeRender | OnBeforeRenderSync`. More verbose but clearer.

I think I've a slight preference for that. Thoughts?

Let me think this a bit.

AurelienLourot:
> Only `Route` should be sync by default (i.e. `Route | RouteAsync`). All other hooks should be `HookName` | `HookNameSync` (async by default).

Actually, there is a caveat I think. When browsing Vike's source code, I came to the conclusion that the following hooks are sync-only: route, onPageTransitionStart, onPageTransitionEnd. The reason is that there is no `await` being done in Vike's source code for exactly these three hooks, so I think there shouldn't be an async variant for them. Or am I mistaken? I can dig that up again if needed.

AurelienLourot:
> I think I've a slight preference for that.

Yes, I prefer explicit `Async/Sync` as well

brillout:
> Yes, I prefer explicit `Async/Sync` as well

👍 Ok let's move with that then: we can add an alias `type OnBeforeRender = OnBeforeRenderAsync` later on anyways 👌.

brillout:
> Let me check about `onPageTransition*`.

Fixed: https://github.com/vikejs/vike/commit/58f822b8f8880f9168478a1956ff8ed6997d06da.

brillout:
> Yes, I prefer explicit `Async/Sync` as well

Since we're explicit, I think we don't need to use async hooks in all examples. Some hooks are trivial and obviously sync: I think it's fine to use the `*Sync` type for them.

AurelienLourot:
@brillout I think it looks good now :)

brillout:
A real beauty 😍

I've added a couple of commits (see the entire commit messages); let me know if you object. (I'm actually not 100% sure about my `type Route = RouteSync` commit but it feels weird to offer a `RouteSync` if `RouteAsync` is deprecated. Not sure. I guess we can keep my commit and experiment how we feel about it later on (we can always deprecated `Route` in favor of `RouteSync`).)

I'm 👍 for proceeding. Btw. we can already merge; we can add docs and update examples in an extra PR. Whatever works best for you.

vike/shared/page-configs/Config.ts
```diff
@@ -39,6 +65,153 @@ type Config = ConfigBuiltIn &
...
+      // See https://vike.dev/stream#initial-data-after-stream-end
+      pageContext: Partial<Vike.PageContext> | (() => Promise<Partial<Vike.PageContext>>)
+    }
+/** @deprecated Use a sync route() with an async guard() instead */
+type RouteAsync = (pageContext: PageContextServer | PageContextClient) => Promise<{ routeParams: Record<string, string> }>
+/** The page's URL(s).
+ *
+ *  https://vike.dev/route
+ */
+type RouteSync = (pageContext: PageContextServer | PageContextClient) => { routeParams: Record<string, string> }
```
AurelienLourot, line 208: This won't work unfortunately, we had this topic earlier in this PR: `PageContextServer` won't be assignable to `PageContextServer | PageContextClient`. It works the other way around. In other words:

* `(A | B) => void` is assignable to `(A) => void`, but
* `(A) => void` is not assignable to `(A | B) => void`

So the user won't be able to stop using client routing and create a `route()` that takes only `PageContextServer` as argument. I can create a TS playground to illustrate if needed.

AurelienLourot:
(all the rest looks good to me :+1:)

vike/shared/page-configs/Config.ts
brillout: Yes I know. My idea is to tell Server Routing users to not use the `Route` type. We can increasingly assume users to use Client Routing because of `vike-{react,vue,solid}`.

My thinking is: users who use `vike-{react,vue,solid}` are more likely to be junior-ish and it's particularly important for them to provide friendly types.

Current users who use Server Routing are "expert users"; I think it's fine to be less friendly. (Alternatively we can provide `RouteWithServerRouting`, like `PageContextClientWithServerRouting`.)

In a nutshell: we highly polish things for Client Routing and neglect Server Routing. (Honestly, I'm even considering deprecating Server Routing.)

brillout:
(We can ignore Cloudflare's CI failing: we've reached Cloudflare's free tier daily rate liming.)

vike/shared/page-configs/Config.ts
AurelienLourot: @brillout even an expert won't be able to perform such an assignment if they do it in `+config.h.ts`, similarly to the problem I was showing here: https://github.com/vikejs/vike/pull/1137#discussion_r1356928529

AurelienLourot: In otherwords, yes if you use a `+route.ts` file then you are free not to use the `Route` type and no assignment type checking will be performed, true, but this won't work when directly assigning in `+config.h.ts`

brillout: True 👍. A solution would be to use `pageContext: PageContextServer` for `Config['route']` (instead of setting it to `RouteSync | RouteAsync`). Alternatively, we create `RouteWithServerRouting` and set `Config['route']` to `RouteSync | RouteAsync | RouteWithServerRoutingSync | RouteWithServerRoutingAsync`. Either way is fine with me.

I know it's very ugly for Server Routing users and only slightly prettier for Client Routing users. That's how much I think we should prioritize Client Routing over Server Routing... (I'd be up for deprecating Server Routing after all).

AurelienLourot: > A solution would be to use `pageContext: PageContextServer` for `Config['route']`

That's brilliant, let me push that

AurelienLourot: actually, I think it's not necessary. As a user, if I disable Client Routing, and I'm forced to still define a `route(PageContextServer | PageContextClient)`, although I know `PageContextClient` will never be passed, so what? Why would this bother me? I'm still going to be able to write a pretty decent implementation. Am I missing something?

AurelienLourot: (if you agree, let's merge :rocket:)

brillout: 👍 I think you're right, the slight mismatch won't bother Server Routing users that much.

To answer your question: yes, there are slight mismatches between `PageContextServer` and `PageContextClient`, the most notable one is that everything is `Partial` because TypeScript cannot now the value of `passToClient`. (See `vike/shared/types.ts`.)

Actually... I'm just realizing that `PageContextClient` is only correct for Client Routing users... For Server Routing users it should actually be `PageContextClientWithServerRouting`. I've just checked and `PageContextBuiltInClientWithServerRouting` and `PageContextBuiltInClientWithClientRouting` (that's the core diff) are quite similar though: I think this mismatch is ok.

This is actually a good example for why I'm inclined to deprecate Server Routing: in my experience the maintaince costs it adds isn't worth it.

After thinking more about it: I suggest we keep the implementation as is while we tell Server Routing users to not use these new types introduced by this PR. This clearly marks Server Routing users as second-class citizen which fits the long-term plan of encouraging users to use `vike-{react,vue,solid}` while eventually deprecating Server Routing as soon it becomes too annoying to maintain.

The issues we're having with it in this PR makes me think it's a reason enough to deprecate it. I'm now inclined to think we should deprecate Server Routing before releasing `v1.0.0`.

> (if you agree, let's merge )

👍 Let me know if you don't object with that last comment and I'll merge.

AurelienLourot: This sounds scary - aren't there valid use cases for not wanting client routing? I'll trust your guts on this

AurelienLourot: (whether or not we deprecate server routing, I think this PR is good enough to be merged)

brillout: While Server Routing is marginally faster on first page load (less client-side code to load), Client Routing is tremendously faster on subsequent navigation. I find this is a reason enough to recommend Client Routing over Server Routing.

Also there are use cases that only work with Client Routing (e.g. a music player that should still play music when navigating pages). That's the biggest reason why I'd always recommend using Client Routing. (You never know what use case you may need in the future.)

The idea of supporting Server Routing was neat but, in hindsight, I think it was a mistake.

All in all, I think it's worth it to deprecate it.

brillout:
Merged 🎉 

Thanks for the PR. Can't wait to release these really neat types.

PR #1161, opened 2023-10-04
Update Flightcontrol link

flybayer:
Thanks for including us! This adds a ref tag to the link so we know users who come from here.


brillout:
Thank *you* for building Flightcontrol 💯.

Issue #1138, opened 2023-09-23
Guard doesn't appear to work for multiple files.


DreamwareDevelopment:
### Description

Hi! I'm just building the redirect logic of trying to access a page behind login without being authenticated. I got it working by putting the guard statement in the same location as each page (duplicating files), however, applying guard to multiple pages from either the /pages/index.page.route.ts root or /pages/admin/index.pages.ts doesn't work as documented. Maybe this functionality isn't under test and broke at some point?

DreamwareDevelopment:
Accidentally made as enhancement and not bug, sorry @brillout

brillout:
Make sure to update to the latest version: there was a patch about this recently.

Closing but let me know if you still run into issues and I'll then re-open.

Bob-Jung:
It still doesn't seem to work. @brillout

brillout:
@Bob-Jung Yes that's a 0.4 design limitation: is upgrading to the V1 design an option for you? (It's going to be officially released very soon.)

PR #1173, opened 2023-10-13
Apply guard to multiple files

Bob-Jung:
This is a patch for [this bug](https://github.com/vikejs/vike/issues/1138).


brillout:
This doesn't work as it executes any `guard()` hook (e.g. from a sibling page file).

PR #1180, opened 2023-10-17
examples: leverage Vike.PageContext to add types to hooks

AurelienLourot:
By adding types to all hooks in all examples, I stumble upon little refactorings in the examples that need to be done. Instead of working for a long time on a monster PR, I think it's easier to review if I do one PR per category of changes.

This PR is about `Vike.PageContext` for the `+onRender*` hooks. If you like this change, I'll push here other changes that follow exactly the same pattern.


brillout:
👍 Agreed (both the changes of this PR and your proposal for step-by-step PRs.)

examples/react-full/renderer/_default.page.client.tsx
```diff
@@ -10,10 +10,10 @@ import React from 'react'
 import ReactDOM from 'react-dom/client'
 import { PageShell } from './PageShell'
 import { getPageTitle } from './getPageTitle'
-import type { PageContextClient } from './types'
+import type { OnRenderClientAsync } from 'vike/types'
 
 let root: ReactDOM.Root
-async function render(pageContext: PageContextClient) {
+const render: OnRenderClientAsync = async (pageContext): ReturnType<OnRenderClientAsync> => {
```
AurelienLourot, line 16: do you think it's fine to add types to non-V1 examples or do you find it misleading? I like it

brillout: Yes it's fine since the hook types work for the non-V1 design as well (and we actually document them in the current non-V1 docs).


That said, if you feel lazy, you can skip the non-V1 examples (since they'll be removed soon). As you wish.

brillout:
(@AurelienLourot I just removed a couple of JavaScript examples from `examples/*` but since they didn't use TypeScript it shouldn't affect your endeavour.)

AurelienLourot:
I'm happy with this one now @brillout

brillout:
Loving these new type hooks 😍

Issue #1189, opened 2023-10-19
`dist/foo/bar.page.js` files unrelated to Vike cause `vite build` to fail


AurelienLourot:
### Description

`pnpm run build --emptyOutDir` calls `vite build --emptyOutDir` in the examples, which should allow us to perform a clean build, however the `--emptyOutDir` option seems to have no effect.

I suspect the problem to be here: https://github.com/vikejs/vike/blob/main/vike/node/plugin/shared/isViteCliCall.ts

AurelienLourot:
Actually, it kind of works already and `outDir` gets emptied as expected, however when setting an `outDir` in an example (as in https://github.com/vikejs/vike/blob/main/examples/react-full-v1/vite.config.ts#L10 where we set it to `dist/nested/`), we seem to manipulate `outDir` to e.g. `dist/nested/client/` in https://github.com/vikejs/vike/blob/main/vike/utils/getOutDirs.ts before passing it to Vite.

So Vite ends up emptying `dist/nested/client/` but technically not the full `dist/nested/`. Instinctively I think this should still be good enough but a user on Discord reported a concrete problem around that so I'll ask them to post a reproducer here and I'll then re-open.

robinelvin:
The Vite build only empties/removes `dist` and `client` directories in `dist`. Any extra directories are left intact which isn't a problem in itself as other build outputs may produce different artifacts. However the problem arises if these other directories exist in `dist` containing files with the word "page" in their name and with content that will produce an error.

Steps to reproduce:

1. Use `examples/react` from this repo
2. `yarn install` or `npm install`
3. Run `yarn build` or `npm build`
4. Observe successful build
5. `mkdir dist/stuff`
6. `echo "import 'foo'" > dist/stuff/index.page.js`
7. `yarn build` or `npm build`
8. Build now fails

AurelienLourot:
@robinelvin thanks a lot. It sounds like a bug on our side. It sounds like manipulating `outDir` isn't a safe thing to do and I'm going to have a look and see if this can be done differently.

Out of curiosity: how did you stumble into this? Did you have at first a vite plugin that was writing files in `dist/`, which you then removed from your setup and the files lingered? Which plugin was it?

robinelvin:
[...]

Issue #818, opened 2023-04-25
Optional include/exclude pattern for faster page files crawl


bagr001:
### Description

Hi,
I have noticed drastic slowdown of project startup time after I placed a "cache" folder in my project containing thousands of files.
I suggest adding optional include/exclude pattern for page files location to VPS configuration, which will increase crawling speed.

```ts
vps({
    pageFiles: {
        include: ['/pages/**/*'],
        // or
        exclude: ['/.cache/**/*'],
    }
});
```

Thanks!

brillout:
Good idea. Let me implement this. ETA today.

brillout:
Actually, while vite-plugin-ssr can implement this, this will require changes on Vite's side as well as Vite [currenty doesn't support such option](https://github.com/vitejs/vite/blob/07bd6d14e545d05c6a29cf341f117fcfe9536ba4/packages/vite/src/node/plugins/importMetaGlob.ts#L394-L402).


I could do a Vite PR. Although I usually do this kind of extra-mile work only for sponsors. I also believe it's a unique feature that other frameworks like Next.js don't support.

brillout:
Actually, I think I'm mistaken: frameworks probably don't crawl the whole repository.

Let me know if this is a blocker for you and I'll have a second look at this.

brillout:
A workaround is to create a hidden directory (i.e. a dir with a name that starts with `.`) and move the files you want to be skipped in there.

brillout:
`fast-glob` (which both Vite and vite-plugin-ssr uses) ignores hidden dirs: https://github.com/mrmlnc/fast-glob#dot.

brillout:
We could also automatically exclude the patterns defined by `.gitignore`.

I wonder whether that would cover most/(all?) uses cases.

@bagr001 @lookyman @tausa17 @Andrewsville @enri90 @stof66 @blagodus Would that work for you?

Let us know if you need more control, i.e. if a `type ConfigCrawl = { gitignore: boolean, exclude: string[], include: string[] }` is needed.

bagr001:
Perfect! I thing excluding the patterns defined by `.gitignore` would cover most use cases.

Thanks!

brillout:
> I thing excluding the patterns defined by `.gitignore` would cover most use cases.

Done by @nitedani in https://github.com/vikejs/vike/pull/1264.

[...]

PR #1192, opened 2023-10-21
fix: don't glob page files in dist/

AurelienLourot:
(attempt to solve #1189)

@brillout @robinelvin with this small patch, an empty `dist/foo/bar.page.js` file doesn't get globbed anymore. Validated with `pnpm run dev|preview|build`, however with `build` there seems to still be a problem, as it fails with

```
vike v0.4.143 pre-rendering HTML...
Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined.
    at Xc (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:77:77)
    at Z (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:78:89)
    at Yc (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:81:98)
    at Xc (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:73:145)
    at Z (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:78:89)
    at Yc (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:81:98)
    at Xc (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:73:145)
    at Z (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:78:89)
    at Xc (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:72:481)
    at Z (/home/.../vike/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server.node.production.min.js:78:89)
 ELIFECYCLE  Command failed with exit code 1.
```

Still digging on my own but @brillout please let me know if this rings a bell too. Thanks!

### DIFF
--- a/vike/node/plugin/plugins/buildConfig.ts
+++ b/vike/node/plugin/plugins/buildConfig.ts
@@ -13,7 +13,7 @@ import {
   unique,
   assertPosixPath,
   assertUsage,
-  getOutDirRootFromResolvedConfig
+  getOutDirs,
 } from '../utils.js'
 import { virtualFileIdImportUserCodeServer } from '../../shared/virtual-files/virtualFileImportUserCode.js'
 import { getVikeConfig } from './importUserCode/v1-design/getVikeConfig.js'
@@ -91,7 +91,7 @@ async function getEntries(config: ResolvedConfig): Promise<Record<string, string
   const pageFileEntries = await getPageFileEntries(config, configVike.includeAssetsImportedByServer) // TODO/v1-release: remove
   const { pageConfigs } = await getVikeConfig(
     config.root,
-    getOutDirRootFromResolvedConfig(config),
+    getOutDirs(config).outDirRoot,
     false,
     configVike.extensions
   )
--- a/vike/node/plugin/plugins/buildConfig.ts
+++ b/vike/node/plugin/plugins/buildConfig.ts
@@ -13,7 +13,7 @@ import {
   unique,
   assertPosixPath,
   assertUsage,
-  getOutDirRoot
+  getOutDirRootFromResolvedConfig
 } from '../utils.js'
 import { virtualFileIdImportUserCodeServer } from '../../shared/virtual-files/virtualFileImportUserCode.js'
 import { getVikeConfig } from './importUserCode/v1-design/getVikeConfig.js'
@@ -91,7 +91,7 @@ async function getEntries(config: ResolvedConfig): Promise<Record<string, string
   const pageFileEntries = await getPageFileEntries(config, configVike.includeAssetsImportedByServer) // TODO/v1-release: remove
   const { pageConfigs } = await getVikeConfig(
     config.root,
-    getOutDirRoot(config.build.outDir),
+    getOutDirRootFromResolvedConfig(config),
     false,
     configVike.extensions
   )
--- a/vike/node/plugin/plugins/buildConfig.ts
+++ b/vike/node/plugin/plugins/buildConfig.ts
@@ -12,7 +12,8 @@ import {
   removeFileExtention,
   unique,
   assertPosixPath,
-  assertUsage
+  assertUsage,
+  getOutDirRoot
 } from '../utils.js'
 import { virtualFileIdImportUserCodeServer } from '../../shared/virtual-files/virtualFileImportUserCode.js'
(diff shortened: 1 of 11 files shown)


brillout:
👍 This works for the current design but, if I ain't mistaken (correct me if I'm wrong), we don't have access to Vite's config with the V1 design (see the `fast-glob` usage in `getVikeConfig.ts`).

We could still merge but, on the other hand, this means the V1 design will be regression for a couple of users.

brillout:
As for the React issue, it's one of these painful React issues that is unfortunately quite hard to debug. I don't think / I wouldn't know how it's related to your PR.

brillout:
FYI this React error happens quite often because of CJS/ESM issues, e.g. if you import `undefined` instead of the actual component.

AurelienLourot:
> if I ain't mistaken (correct me if I'm wrong), we don't have access to Vite's config with the V1 design (see the `fast-glob` usage in `getVikeConfig.ts`).

I'll have a look at that :)

> As for the React issue, it's one of these painful React issues that is unfortunately quite hard to debug. I don't think / I wouldn't know how it's related to your PR.

It doesn't happen if I delete the my empty `dist/foo/bar.page.js`. What I'm trying to achieve here is that files outside `dist/client/` and `dist/server/` like this one shouldn't impact us at all, no matter how wrong they are. I'm not sure why this file gets picked up as it's now not part of the glob results

brillout:
There is also Vite's `import.meta.glob` usage in `getVirtualFileImportUserCode.ts` which may explain this. (This is only used by current design; the V1 design only uses `fast-glob`.)

AurelienLourot:
Indeed the culprit was `import.meta.glob`. I've just pushed a fix for that. So now it seems to work well with the current (non-V1) design, as in: dummy page files under `dist/` aren't picked up anymore.

> if I ain't mistaken (correct me if I'm wrong), we don't have access to Vite's config with the V1 design (see the `fast-glob` usage in `getVikeConfig.ts`).

I still would like to have a look at that

AurelienLourot:
I have now solved the problem for V1 in exactly the same way: `config.build.outDir` was available everywhere needed. Validated locally with an empty `dist/+dummy.ts` file. Without this patch, `pnpm run build|dev` would fail. With this patch the dummy file doesn't get globbed and the build succeeds.

I'm quite happy with this PR now

AurelienLourot:
The posix path failure on Windows is caused by me. Digging

AurelienLourot:
Looks good now :)

brillout:
Neat, I'm positively surprised `getVikeConfig()` is able to require Vite's `config`.

Let me think through if I can foresee an issue with the upcoming `vike.config.h.js`.

brillout:
I think we're good.

Exciting. It has been a recurring pain for users &mdash; I'm very glad it's fixed now ✨

Let me review/polish some details.

brillout:
Alright, did some refactoring to `getOutDirs.ts`. (Still dirty but it's slowly getting slightly cleaner, at least not *as* dirty as it used to be :-). FYI I added the many assertions to get the whole thing under control as it's an intricate set of interactions with Vite's config.)

brillout:
Thanks for the PR! I'm super excited about this: one pain less for many users 👌.

Btw. I think https://github.com/vikejs/vike/issues/818 is now implementable.

AurelienLourot:
Your last changes seem to have broken it, let me have a look

PR #1194, opened 2023-10-22
refactor: update rollup warning types

bluwy:
Since [Rollup 3.25.0](https://github.com/rollup/rollup/releases/tag/v3.25.0), it introduced the `RollupLog` type in favour of `RollupWarning` as it supports `this.info`, `this.warn`, etc. `RollupWarning` was made an alias of `RollupLog`.

In Rollup 4 which Vite 5 uses, `RollupWarning` was removed. In this PR, I updated to use `RollupLog` instead.


brillout:
Thanks!

Issue #578, opened 2022-12-27
V1 Design


brillout:
The `vite-plugin-ssr@1.0.0` design in a nutshell:

```ts
// /pages/product/+config.ts

import type { Config } from 'vite-plugin-ssr'

export default {
  Page: './Page.vue',
  route: '/product/@id',
  onBeforeRender: './onBeforeRender.ts',
  onPrerender: './onPrerender.ts',
  prerender: true,
  ssr: false // render as SPA
} satisfies Config // `satisfies` is a new TypeScript 4.9 operator
```

I.e. `+config.js` replaces `.page.js`, `.page.server.js`, `.page.client.js`, and `.page.route.js`.

It also replaces `_default.page.js` as well as VPS configurations set in `vite.config.js`:

```ts
// /pages/+config.ts

import type { Config } from 'vite-plugin-ssr'

// Applies as default to all pages
export default {
  onRenderHtml: './config/onRenderHtml.tsx',
  onRenderClient: './config/onRenderClient.tsx',
  prerender: {
    partial: true // Currently lives in vite.config.js
  },
  ssr: true
} satisfies Config
```

VPS's architecture stays the same and therefore migration is relatively simple.

### Succint syntax

Under consideration is the possibility to create new pages in a succint manner:

```js
/pages/index/+Page.js
/pages/about/+Page.js
/pages/jobs/+Page.js
```

Here, a new `+config.js` file isn't needed each time a new page is created.

I'm currently leaning towards *not* supporting such succint syntax (in favor of "Single Config File", see next section). But I may reconsider if users want this.

### Single Config File (aka Single Route File)

```ts
// /pages/+config.ts

import type { Config } from 'vite-plugin-ssr'

const marketingPagesConfig = {
  Layout: './layouts/LayoutMarketingPages.vue',
  ssr: true,
}
const adminPagesConfig = {
  title: 'Admin Panel',
  Layout: './layouts/LayoutAdminPages.vue',
  ssr: false
}

export default {
  // If `singleConfigFile: true` then only one `+` file is allowed (this file). If there is
  // anothoer `+` file, then VPS displays a warning.
  singleConfigFile: true,
  pages: [
    { ...marketingPagesConfig, Page: './LandingPage.vue', route: '/'        , title: 'Awesome Startup'                          },
    { ...marketingPagesConfig, Page:    './JobsPage.vue', route: '/jobs'    , title: "We're hiring!"                            },
[...]

PR #1206, opened 2023-10-24
V1-first documentation

AurelienLourot:
Long living (merging ETA: 2023-11-30), collaborative branch for making the documentation teach the V1 design (see #578) first.

Feedback and commits welcome at any time. Please avoid force-pushing.

Preview: https://vike-v1-staging.lourot.dev.


docs/pages/render-hook.page.server.mdx
```diff
@@ -1,3 +1,13 @@
+import { Link, Warning } from '@brillout/docpress'
+
+<Warning>
+  `render()` is deprecated in favor of `onRenderHtml()`, see
+  <ul>
+    <li><Link href="/onRenderHtml" /></li>
+    <li><Link href="/migration/v1-design#renamed-hooks" /></li>
+  </ul>
+</Warning>
```
brillout, line 9: I think we can also be lazy and just remove that page - as you wish.

brillout:
🚀 Neat.

> Please avoid force-pushing.

👍  (Now that Vike's code is increasingly collaborative, I'll stop force pushing on non-`brillout/*` branches.)

brillout:
> commits welcome

Just made one. Feel free to object to any of my (future) commits.

docs/pages/render-hook.page.server.mdx
AurelienLourot: Good to know :pray: So far it's easy to keep these old pages but if it becomes cumbersome I'll remove them

docs/pages/add.page.server.mdx
```diff
@@ -13,6 +13,8 @@ To add Vike to an existing Vite app:
 
    - <RepoLink text="Example" path="/boilerplates/boilerplate-vue/server/index.js" />
 
+{ /* NOTE(aurelien) modify the boilerplates to use the V1 design */ }
```
AurelienLourot, line 16: @brillout I think I should do this in this same PR, so that the V1 docs and the V1 boilerplates land together. What do you think?

brillout: Yes makes sense 👍

AurelienLourot: done :) feel free to review `/boilerplates/**` in this PR if you feel like it

brillout: Looks good 💯

docs/pages/route-string.page.server.mdx
```diff
@@ -50,13 +50,14 @@ If you define multiple globs, then the glob values are available at `pageContext
 You can use a <Link href="#globs">glob</Link> to catch all URLs:
 
 ```js
-// catch-all.page.route.js
+// /pages/catch-all/+route.js
```
AurelienLourot, line 53: @brillout is this the right way of doing `catch-all` in V1?

brillout: Yes, this also works (as long as the Route String value set by `+route.js` is `/*` or `*`).

AurelienLourot: Ah yes, I got confused for a moment and felt like `catch-all` was a special string, but it can be anything of course. Thanks

brillout: 👍

docs/pages/multiple-renderer.page.server.mdx
```diff
@@ -1,4 +1,9 @@
-import { Link } from '@brillout/docpress'
+import { Link, Warning } from '@brillout/docpress'
+
+<Warning>
+  Having multiple `/renderer` folders is deprecated in favor of having `+*.js` files in higher page folders,
+  see <Link href="/default-page" />
+</Warning>
```
AurelienLourot, line 6: @brillout is this fine with you? I have reworked `/default-page` to teach that

docs/pages/file-structure.page.server.mdx
```diff
@@ -10,26 +10,26 @@ while for basic apps we recommend a single `pages/` directory.
...
 
 # Code that specifies how our pages are rendered
-/renderer/_default.page.server.js
-/renderer/_default.page.client.js
-/renderer/_error.page.js
+/renderer/+onRenderHtml.js
+/renderer/+onRenderClient.js
 /renderer/PageShell.js  # The root (React/Vue/...) component that wraps the `Page` component
-/renderer/PageShell.css # exported by `/pages/**/*.page.js`.
+/renderer/PageShell.css # exported by `/pages/**/+Page.js`.
```
AurelienLourot, line 32: @brillout did you mean "imported"? Not sure I understand this comment

docs/pages/prerender-meta.page.server.mdx
```diff
@@ -0,0 +1,73 @@
...
+
+Note that <Link text={<><code>onBeforePrerenderStart()</code> hooks</>} href="/onBeforePrerenderStart" /> are always
+called, regardless of `prerender`. If we want to call `onBeforePrerenderStart()` hooks conditionally as well:
+
+{ /* NOTE(aurelien): rewrite this last part after Vike bug fixed where `prerender` seems to have no effect */ }
+
+```js
+export const doNotPrerender = someCondition() ? true : false
+export const prerender = doNotPrerender ? undefined : myPrerenderHook
+```
```
AurelienLourot, line 73: @brillout That's the last missing bit - we talked about it offline

docs/pages/multiple-renderer.page.server.mdx
brillout: 👍 I like the idea of showcasing default settings increasingly as `/pages/+someConfig.js` and less and lass as `/renderer/+someConfig.js`. It makes sense since less and less users are going to have a `renderer/` directory as they'll increasingly use `vike-{react,vue,solid}`.

That said, for users with custom renderers, such as Ecosia, I'm thinking `renderer/` can still make sense. IMO tucking the rendering logic into a `renderer/` directory is a neat way to seperate render logic from defining pages.

Also I'm increasingly thinking that it would be neat to support `$ eject vike-react` which would move the renderering logic from `node_modules/vike-react/src/renderer` to `renderer/` so that the user can gain full control. So I'm thinking the `renderer/` directory could also be useful in that context.

So while I agree that we should showcase `renderer/` less and less, I think it would be nice to still use it in docs that showcase custom renderers. Or, alternatively, we can just mention it as a note "Btw, for a clear separation of concerns, you can create a `renderer/` directory in order to...") .

WDYT?

docs/pages/file-structure.page.server.mdx
brillout: Indeed, it doesn't make sense 😅. I'm 👍 for removing ~~# exported by /pages/**/+Page.js~~.

docs/pages/migration/v1-design.page.server.mdx
```diff
@@ -1,6 +1,4 @@
-import { Link, Warning } from '@brillout/docpress'
-
-<Warning>The V1 design is in beta.</Warning>
+import { Link } from '@brillout/docpress'
```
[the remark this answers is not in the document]
brillout: 💯

AurelienLourot (commented):
Happy with this PR now. Apart from the minor ongoing discussions in this PR, I'm done with reworking the docs for teaching the V1 design. I did the following:

* I browsed and read the entire doc, following the navigation menu
    * I adapted the page contents to be V1
    * I marked some pages as deprecated and moved them to "detached"
    * I took some existing "detached" pages and attached them to the navigation menu
* I browsed all detached pages
    * I adapted the page contents to be V1, if they weren't deprecated
* I slightly moved some pages within the navigation menu and made sure everything still looks sound
* I added some more typescript docs
* I updated all the links to examples in order to point to `*-v1/` examples
* I converted all the boilerplates to V1
* Finally, I grepped for some strings to make sure I forgot nothing, e.g. `.page`

docs/pages/prerender-meta.page.server.mdx
brillout: 👍 I'll have a look at it after I'm done with [this feature request](https://github.com/vikejs/vike/issues/1231) (which is a blocker for [a high-tier sponsor](https://github.com/Alignable)).

brillout:
> I did the following:

Sounds good. I'll have a look at it once I'm done with that feature request.

docs/pages/multiple-renderer.page.server.mdx
AurelienLourot: I think there is a slight misunderstanding. This page isn't about the root `/renderer` folder but about some `/**/renderer` subfolders, which I think aren't so useful, as things can just be inside `/**/pages` instead. I'm still happy with advertizing a root `/renderer` folder. What do you think?

docs/pages/file-structure.page.server.mdx
AurelienLourot: done :) (removed `# exported by /pages/**/+Page.js`)

docs/pages/multiple-renderer.page.server.mdx
brillout: I see, although I think the following POV still holds:

> That said, for users with custom renderers, such as Ecosia, I'm thinking `renderer/` can still make sense. IMO tucking the rendering logic into a `renderer/` directory is a neat way to seperate render logic from defining pages.

E.g. the classic example of having an admin panel rendered as SPA while having landing/marketing pages rendered with SSR.

I think it's neat to have these two very different rendering logics live in two separate `renderer/` directories.

AurelienLourot: alright, makes sense, I'll revert the deprecation notice and adapt that page too

AurelienLourot: fixed

docs/pages/index/index.page.server.tsx
```diff
@@ -20,15 +20,13 @@ function Page() {
...
+function GlobalNote() {
   return (
     <>
       <div style={{ display: 'flex', justifyContent: 'center', textAlign: 'center' }}>
         <blockquote>
           <p>
-            <a href="https://vite-plugin-ssr.com"></a>
-            <code>vite-plugin-ssr</code> has been renamed Vike, see{' '}
-            <a href="https://vite-plugin-ssr.com/vike">migration guide</a>.
+            The <i>V1 design</i> has been released, see <Link text="migration guide" href="/migration/v1-design" />.
```
AurelienLourot, line 29: Is this ok?

brillout: Good idea 💯

brillout: I'm seeing that there are still [1/3 of people on vite-plugin-ssr](https://www.npmjs.com/package/vite-plugin-ssr). How about we have both? E.g. with `<ul>`.

brillout:
(I managed to fix the flaky test; we should get much less false CI failures. I also have an idea about the failing CloudFlare tests when we hit CF's API rate limit.)

docs/pages/index/index.page.server.tsx
AurelienLourot: Fixed, I've decided to go for two thinner notes on top of each other:

Screenshot from 2023-11-04 10-52-31

brillout: 👍

AurelienLourot:
Staging preview: https://vike-v1-staging.lourot.dev/

docs/pages/prerender-meta.page.server.mdx
brillout: I just checked https://vike.dev/doNotPrerender#conditional-pre-rendering and the use case described there (disabling pre-rendering for "deploy live previews") should, AFAICT, also work with the V1 design.

The idea being to set an env `DISABLE_PRERENDER` and then setting `prerender: !process.env.DISABLE_PRERENDER` and `onBeforePrerenderStart: process.env.DISABLE_PRERENDER ? null : onBeforePrerenderStart`.

AurelienLourot: indeed it's easier to do

```js
export const doNotPrerender = someCondition() ? true : false
export const prerender = someCondition() ? undefined : myPrerenderHook
```

that

```js
export const doNotPrerender = someCondition() ? true : false
export const prerender = doNotPrerender ? undefined : myPrerenderHook
```

the second one would probably require an effect, and would be quite hard to describe in the documentation. I'll go for the first one then.

AurelienLourot: So this doesn't work:

```js
import { onBeforePrerenderStart } from './myOnBeforePrerenderStartHook.js'

export default {
  prerender: someCondition(),
  onBeforePrerenderStart: someCondition() ? onBeforePrerenderStart : undefined
} satisfies Config

function someCondition() {
  return false
}
```

if the condition is `false`, of course we get

```
Error: [vike][Wrong Usage] /pages/about/+config.h.ts imports the following:
  import { onBeforePrerenderStart } from "./myOnBeforePrerenderStartHook.js";
But the import onBeforePrerenderStart isn't re-exported at export default { ... } and therefore has no effect, see explanation at https://vike.dev/header-file
```

That's understandable but a bit of a pity. This will have to be solved with a `+onBeforePrerenderStart.js` file. Giving it a try

brillout: I think we can remove that assertUsage(). The trade off here is that we lose a bit on educating users about config.h.js but it could be worth it 🤔.

Or we just turn it into a warning. At least it isn’t blocking then.

AurelienLourot: There seems to be a barrier in every path I take:

```js
// +onBeforePrerenderStart.js

export { onBeforePrerenderStart }

import { someCondition } from './someCondition'

const onBeforePrerenderStart = someCondition()
  ? async () => {
      return ['/about']
    }
  : null

// fails with
// Error: [vike][Wrong Usage] Set onBeforePrerenderStart to null in a +config.h.js file instead of /pages/about/+onBeforePrerenderStart.js
```

```js
// +onBeforePrerenderStart.js

export { onBeforePrerenderStart }

import { someCondition } from './someCondition'

const onBeforePrerenderStart = someCondition()
  ? async () => {
      return ['/about']
    }
  : undefined

// fails with
// Error: [vike][Wrong Usage] Hook onBeforePrerenderStart() defined by /pages/about/+onBeforePrerenderStart.js should be a function
```

I don't think there is a working V1-way for unsetting a hook based on a condition at the moment. Am I missing something?

AurelienLourot: maybe the assertUsage that could be lifted, because not-dangerous, is the one preventing me from exporting `null` in `+onBeforePrerenderStart.js`?

brillout: > // Error: [vike][Wrong Usage] Set onBeforePrerenderStart to null in a +config.h.js file instead of /pages/about/+onBeforePrerenderStart.js

This one is necessary because Vike needs to know whether a global hook has been suppressed (i.e. set to `null`). (Vike doesn't load hooks at config-time, so the only way to tell Vike that a hook is `null` at config-time is to set it in `+config.h.js`.)

<hr>

How about we do the following? It seems to be the quickest way to unblock us. Thoughts?

> Or we just turn it into a warning. At least it isn’t blocking then.

It'll give us some more time until users complain about the warning. (And I think it's okay-ish if we don't tell users about the warning in the docs if we feel lazy.)

AurelienLourot (commented):
(nice, LGTM, thanks :heart:)

AurelienLourot (commented):
Latest changes look good to me :pray:

brillout:
> docs: fix copy-paste mistakes and typo

🙏

pnpm-lock.yaml
```diff
@@ -24,7 +24,7 @@ importers:
         version: 0.5.25
       '@brillout/test-types':
         specifier: ^0.1.10
-        version: 0.1.10(typescript@4.9.5)
+        version: 0.1.10(typescript@4.8.4)
```
AurelienLourot, line 27: I don't understand why `cd docs && pnpm install @brillout/docpress@latest` changed the typescript version everywhere. I tried several times by doing `pnpm reset` just before but I couldn't prevent pnpm from touching these lines. Any ideas?

brillout: I also noticed some behavior that seemed quite unexpected in that regard; I didn't dig into it yet.

brillout:
I think https://vike-v1-staging.lourot.dev/header-file is now really good and I now consider it done. (Finally 😅 it wasn't easy to find explanations that is both succint and clear 🙂). But let me know if you feel something's missing / not good in there.

docs/components/HeaderFileIntro.mdx
```diff
@@ -12,18 +12,16 @@ export default {
...
+const onBeforeRender = 'import:./some-other-file:default';
 
 export default {
   Page,
   onBeforeRender
 }
 ```
 
-> `.h.js` files are only meant for config files (i.e. any `.h.js` file should be a `+config.h.js` file).
+> This transformation is only for config files. (Any `.h.js` file is a `+config.h.js` file.)
```
AurelienLourot, line 27: Does it mean "only `+config.h.js` is supported"? or does it mean "`anything.h.js` will be treated as `+config.h.js`"?

docs/pages/header-file.page.server.mdx
```diff
@@ -62,38 +45,57 @@ export default {
...
+
+Instead of `+config.h.js` you can define a `+config.js` file (without `.h`):
+
+```js
+// +config.js
+
+export default {
+  Page: 'import:./some-file:someExport',
+  onBeforeRender: 'import:./some-other-file:someOtherExport'
+}
```
AurelienLourot, line 75: Actually I can also do that in a `+config.h.js` (I just tried it out). Is there anything I can only do in a `+config.js` but not in a `+config.h.js`?

AurelienLourot (commented):
Great job! Two things aren't very clear to me and can maybe be improved?

docs/pages/header-file.page.server.mdx
AurelienLourot: Ok actually I think I know the answer: in a `+config.js` if I do imports they won't be transformed. So I can import functions and use them but I won't be able to re-export them. It can be useful in some cases. Maybe that's what we should show here? Otherwise if we don't want to advertise the possibility to import without transformation, then maybe we shouldn't advertise `+config.js` at all?

docs/components/HeaderFileIntro.mdx
AurelienLourot: Could I create a `+Page.h.js`  to feed the Page config and have transformed imports in there, even if this makes no sense?

brillout: 2e58a6c33bd22ee07bd84231c1945b7a6dcc43cf &mdash; hopefully it's clearer now.

docs/pages/header-file.page.server.mdx
brillout: I see. I'm realizing the docs should further explain why the transformation is needed (and explain why code extraction is problematic). I tried but I realized, while trying, that it would be quite time consuming to properly explain all that. So I'm thinking to postpone working on this after the V1 design release.

brillout: And I think you're right: 9a4d978d2576dd05f1d485072641c1b9fde8bc9a. I'm realizing it's contra productiv to open the "can of worms" without properly explaining everything.

brillout:
The overall navigation structure LGTM now. (It isn't perfect but I think it's good enough for now.)

I will now review the docs changes on a page-by-page basis. I expect it to be mostly about proofreading and, therfore, to be fairly quick.

brillout:
Getting there 🙂 image

AurelienLourot (commented):
Everything still LGTM so far :pray:

brillout (approved):

brillout:
This now LGTM.

@AurelienLourot Let me know if my latest changes are ok with you. I'll then implement the deprecating warning for the current design and release. Let's prepare the release tweets for Monday.

AurelienLourot (commented):
LGTM :rocket:

brillout:
> LGTM 🚀

🙏

Let me release the new version and I'll then merge.

brillout:
@AurelienLourot Huge thanks for the PR 🙏 it has been a massive help.

afkcodes:
Amazing work by @AurelienLourot thanks a lot, and thanks @brillout for all the hardwork.

brillout:
Everything is ready. Let's merge tomorrow at noon before the release tweet.

Issue #1205, opened 2023-10-24
"constructor" search param in URL causes internal server error


uifox:
### Description

Using vike@0.4.144 I experience a server side error when a search param with key "constructor" is present in the URL, e.g. http://www.example.com/provider/wok?constructor=12

```
TypeError: (searchAll[key] || []) is not iterable
    at file:///home/appuser/node_modules/vike/dist/esm/utils/parseUrl.js:54:46
    at Array.forEach (<anonymous>)
    at parseUrl (file:///home/appuser/node_modules/vike/dist/esm/utils/parseUrl.js:52:59)
    at modifyUrlPathname (file:///home/appuser/node_modules/vike/dist/esm/utils/parseUrl-extras.js:73:72)
    at normalizeUrlPathname (file:///home/appuser/node_modules/vike/dist/esm/utils/parseUrl-extras.js:57:27)
    at normalizeUrl (file:///home/appuser/node_modules/vike/dist/esm/node/runtime/renderPage.js:319:27)
    at renderPageAndPrepare (file:///home/appuser/node_modules/vike/dist/esm/node/runtime/renderPage.js:84:40)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at file:///home/appuser/node_modules/vike/dist/esm/node/plugin/shared/getHttpRequestAsyncStore.js:68:35
```

This is caused by https://github.com/vikejs/vike/blob/926313b298dbffac434ab52a44be7b24c1d16d7b/vike/utils/parseUrl.ts#L80

where `searchAll['constructor']` is never `undefined`. It is in fact a function (the constructor of `{}`) causing the `searchAll[key] = [...(searchAll[key] || []), val];` expression to fail.

```
> [...(searchAll['constructor'] || []), 'val'];
Uncaught TypeError: (searchAll.constructor || []) is not iterable
```

brillout:
> @brillout Let me know if I can help fix this issue. I'd be glad to.

Ok!

uifox:
@brillout I cloned this repo, created a branch, fixed this issue, but I can't push my changes. What's the process I should follow?

```
[uifox/issue-1205 afbfceb27] fix: allow "constructor" search param in URL
 2 files changed, 8 insertions(+), 4 deletions(-)
[...]

PR #1215, opened 2023-10-27
Allow "constructor" search param in URL

uifox:
This fixes https://github.com/vikejs/vike/issues/1205

### What I did
* I made sure search params with the `constructor` key in URLs don't make parseUrl() blow up.
* Also fixed a few typos

### How I did
In following code

```ts
searchAll[key] = [...(searchAll[key] || []), val]
```

when `searchAll[key]` is accessed the first time it is typically `undefined` and `searchAll[key] || []` evaluates as an empty array resulting in `searchAll[key] = [val]`. However, if `key` is `'constructor'` `searchAll['constructor']` is a function resulting `searchAll['constructor'] || []` to be a function, which is not iterable, hence `...(searchAll[key] || [])` throws an error.

One possible solution to the issue is checking if `searchAll[key]` is an array and spread it only if so, otherwise fall back to the empty array.

### DIFF
--- a/vike/utils/parseUrl.spec.ts
+++ b/vike/utils/parseUrl.spec.ts
@@ -125,28 +125,28 @@ describe('parseUrl', () => {
       searchAll: { q: ['apples'] },
       searchOriginal: '?q=apples'
     })
-    expect(parseUrl('/shop?fruits=apples&candies=chocolate,lolipop', '/')).toEqual({
+    expect(parseUrl('/shop?fruits=apples&candies=chocolate,lollipop', '/')).toEqual({
       ...resultBase,
       pathnameOriginal: '/shop',
       pathname: '/shop',
-      search: { fruits: 'apples', candies: 'chocolate,lolipop' },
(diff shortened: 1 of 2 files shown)


brillout:
Thanks for the PR! How about we use [hasOwnProperty](https://stackoverflow.com/a/455366/270274)? I think the intent would be clearer.

brillout:
And thanks for the typos 🙂

uifox:
> Thanks for the PR! How about we use [hasOwnProperty](https://stackoverflow.com/a/455366/270274)? I think the intent would be clearer.

Sure, that works too. I changed it.

brillout:
Merged 🎉

Pre-released as `vike@0.4.144-commit-68c730d`.

Btw. would your company be up for [sponsoring](https://github.com/sponsors/brillout)?

uifox:
> Btw. would your company be up for [sponsoring](https://github.com/sponsors/brillout)?

I im'd the decision makers. Still waiting. Will pop it up again in a week or so.

brillout:
Thanks 💚 sponsoring makes a *big* difference.

PR #1249, opened 2023-11-08
docs: add new react-query integration example

AurelienLourot:
See vikejs/vike-react#30

FYI @phonzammi


AurelienLourot:
Is it me or some tests have become more flaky recently?

brillout:
I'm aware of this one flakiness; I did dig into it yet because it occurred only very seldomly. 

> more flaky recently

Yea it's being a couple of weeks I have the same impression. Maybe something changed with GitHub's Actions.

brillout:
@AurelienLourot So far I was pushing new examples at the top of lists, I'm not sure which way is the best. As you wish.

AurelienLourot:
makes sense! I've reordered and pushed directly to main. Hope that's OK

brillout:
Totally 💯

PR #1251, opened 2023-11-10
docs: add integration example with Grommet

AurelienLourot:
FYI @tdonhauser

### DIFF
--- a/docs/headings.tsx
+++ b/docs/headings.tsx
@@ -451,6 +451,11 @@ const headings = [
   },
   {
     level: 2,
+    title: 'Grommet',
+    url: '/grommet'
+  },
+  {
+    level: 2,
     title: 'Sass / Less / Stylus',
     url: '/sass'
   },
--- a/docs/pages/css-in-js.page.server.mdx
+++ b/docs/pages/css-in-js.page.server.mdx
@@ -1,3 +1,4 @@
+import { Link } from '@brillout/docpress'
 import { CSSinJSTool } from '../components'
 
 <CSSinJSTool prefix="Some CSS-in-JS tools allows us to" />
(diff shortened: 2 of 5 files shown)


AurelienLourot:
FYI @brillout I've put this example in a separate repo because:
1. I didn't feel like putting it in this repo because it's based on vike-react
2. I didn't feel like putting it in the vike-react repo because it's based on having ejected `onRenderHtml()` and a bunch of files it depended on.

So I feel like the example would look nicer if it was either based on Vike directly, or if we had a nicer way of having it on top of vike-react without ejecting so much code (maybe a bit the same spirit as https://github.com/vikejs/vike-react/issues/28 )

So it's a bit of a misfit right now but maybe later we can make it a first-class example.

AurelienLourot:
Sorry I realize I should have merged "rebase", so without a merge commit. I'll do that next time.

brillout:
> So I feel like the example would look nicer if it was either based on Vike directly, or if we had a nicer way of having it on top of vike-react without ejecting so much code (maybe a bit the same spirit as [vikejs/vike-react#28](https://github.com/vikejs/vike-react/issues/28) )

Yes, that'd be 💯:
 1. Making `vike-react` ejectable. (FYI [stem-react](https://github.com/brillout/stem-react), which is the experimental precursor of `vike-react`, [is ejectable](https://github.com/brillout/stem-react/blob/main/eject.config.ts) but it's using the old design. We could port that functionality to `vike-react` and the new design.)
 2. Having an ejectable integration `vike-gromment` that seamlessly integrates with Vike/`vike-react`.

If we manage to pull it off, I think such ejectable integrations would be a tremendous added value.

> Sorry I realize I should have merged "rebase", so without a merge commit. I'll do that next time.

No problems, the commit history is fairly messy anyways.

PR #1255, opened 2023-11-13
docs: add new tanstack query example

AurelienLourot:


AurelienLourot:
I forgot again about the merge strategy :/ Will try to remember next time. There is no way on GitHub to set a default strategy per project unfortunately and I'm used to choose the merge-commit strategy without thinking about it

PR #1259, opened 2023-11-16
docs: extend /broken-npm-package

AurelienLourot:


brillout:
I just pushed some improvements.

This PR LGTM; feel free to merge if you agree with my changes.

I'm realizing we could actually already start catching this error and then pointing users to this section of the docs.

PR #1262, opened 2023-11-16
Write prerendered files after each page render, don't keep the html in memory

nitedani:


nitedani:
Benchmark: 
`prerender: { parallel: 1 }`
55mb large `pages/hello/index.page.ts`
in `writeHtmlFile` : `gc(); console.log(process.memoryUsage().heapUsed)`


Before: multiple instances of the page are kept in memory, for the whole run
before

After: the page gets cleared from the memory after each page render
after

brillout:
Writing files early is definitely a big improvement 💯

Let me know if you're on twitter and I'll mention you on the next release announcement tweet.

PR #1264, opened 2023-11-18
add readGitignore option

nitedani:


nitedani:
If `pages` is in gitignore, but it contains page files, the files still get included in the manifest, and the build fails in `getManifestEntry`

brillout:
Very neat 👌

Couple of thoughts:
 - How about we always read `.gitignore`? So that we don't need an extra setting `readGitignore`, and so that everything "just works" for the user. There is the risk that the user mistakenly adds `pages/` to `.gitignore` and then wonders "why isn't Vike working" but I think it's a very seldom situation we can ignore or, alternatively, we tell the user that Vike uses `.gitignore` when Vike throws its error "No page file found" and/or adding a note to the documentation.
 - I wonder what `.gitignore` patterns `fast-glob` supports. (Hopefully there isn't any pattern mismatch leading to false interpretations by `fast-glob` that leads to ignoring too much.)
 - Because `fast-glob` search's root is `userRootDir`, I believe we can only consider `${userRootDir}/.gitignore`, otherwise we may ignore too much. To support monorepos, we could search between `/` and `userRootDir` while aborting if we stumble upon the `.git` directory.

WDYT?

nitedani:
I found the missing piece. Now it works as I would expect. I tried to think of a case when the user deliberatly wants to add something to the gitignore that would break the build, but I'm not sure it's realistic.
So ok, let's go with your idea. We can change it later if there is an issue.

Test cases:
image
image
image

vike/node/plugin/plugins/importUserCode/getVirtualFileImportUserCode.ts
```diff
@@ -317,10 +323,11 @@ function getGlobs(
 }
 
 function getGlobRoots(config: ResolvedConfig, configVike: ConfigVikeResolved): GlobRoot[] {
+  const gitignoreLines = getGitignoreLines()
   const globRoots: GlobRoot[] = [
     {
       includeDir: '/',
-      excludeDir: path.posix.relative(config.root, getOutDirs(config).outDirRoot)
+      excludeDir: [path.posix.relative(config.root, getOutDirs(config).outDirRoot), ...gitignoreLines]
```
AurelienLourot, line 330: This won't work unfortunately because `gitignoreLines` can contain negative patterns like `!something`. But then on line 308 we prefix `excludeDir` also with an exclamation mark `!`. So we will end up with double exclamation marks `!!` being fed to `import.meta.glob()`, which I don't think works.

> Note: At other places however where we just pass all the lines of gitignore to `fast-glob`, things are fine as `fast-glob` [supports negative patterns](https://github.com/mrmlnc/fast-glob)

> Other note: if `excludeDir` can now contains several dirs, it should be renamed `excludeDirs`

AurelienLourot (changes requested):
Good idea in general but see my concern about negative gitignore patterns, thanks!

vike/node/plugin/plugins/importUserCode/getVirtualFileImportUserCode.ts
nitedani: What if readAndSetGitignoreLines filtered the lines starting with `!`? I think we can ignore those lines.

AurelienLourot: It might go wrong in some cases though:

```
/ignore/this/folder/
!/ignore/this/folder/but/not/this/subfolder/  # we would miss what is in there if we ignore this line
```

But fine with me as this is, I think, quite rare. We could also print a warning in the console if we encounter such lines in the gitignore.

Another possibility is to not care too much, as this is the logic for the old V0.4 design (the `.page` files), not for the V1 design (the `+` files), so maybe it's fine if we don't even support gitignore at all for the old design? What do you think @brillout ?

AurelienLourot: So my vote would be: we don't touch this V0.4 old logic at all. One way or another we can't leave it as is, as if we end up writing `!!` things will fail in the user's project in cryptic ways

vike/node/plugin/shared/readGitignore.ts
```diff
@@ -0,0 +1,46 @@
...
+  for (;;) {
+    const res = await readdir(currentDir)
+    const gitignoreFile = res.find((fname) => fname.toLowerCase() === '.gitignore')
+    if (gitignoreFile) {
+      const content = await readFile(join(currentDir, gitignoreFile), 'utf8')
+      const lines = content
+        .replace(/\r\n/g, '\n')
+        .split('\n')
+        .filter((line) => line.length)
+      allLines.push(...lines)
```
AurelienLourot, line 30: I see another problem unfortunately: collecting all lines of all .gitignore files isn't correct actually, because where the .gitignore file is located plays a role. Let me elaborate:

```
# /foo/.gitignore
/bar
!/bar/baz
```

means actually:

```
/foo/bar
!/foo/bar/baz
```

Each line should be kind of concatenated with the folder in which the `.gitignore` is located (here `/foo`), but then of course there is a spacial handling for:
* question marks `!`
* leading or no leading `/`

So in other words, globbing all .gitignore in a repo and aggregating all their contents is quite a complex endeavor. Is there a library out there we could use for this?

> Note: and another problem I just noticed is that it doesn't seem that you are removing `# comments` from the .gitignore files? or did I miss it?

AurelienLourot (changes requested):

vike/node/plugin/plugins/importUserCode/getVirtualFileImportUserCode.ts
brillout: The underlying issue is https://github.com/vikejs/vike/issues/818 which also holds for the V1 design. If we manage to apply `.gitignore` to Vike's `fast-glob` usage that'd be quite neat as it'd solve the issue in the (vast?) majority of cases.

I wonder if it's feasible to apply `.gitignore` in a safe way, albeit only a subset of `.gitignore`.

AurelienLourot: so would you be fine solving it for V1-only?

brillout: Yes adding this only for the V1 design is fine.

vike/node/plugin/shared/readGitignore.ts
AurelienLourot: Sorry I misread the code. I thought you were globbing all `.gitignore` files. But we support only a `.gitignore` in the root folder?

AurelienLourot: Alternatively, instead of not perfectly supporting `.gitignore` and leading to problems and surprises for the user, we could support a very simple `.vikeignore` file instead and say:
* no comments
* no `!`
* only one in the whole repo, at the root of the repo.

What would you think of that @brillout @nitedani ?

vike/node/plugin/plugins/importUserCode/getVirtualFileImportUserCode.ts
brillout: Vite's `import.meta.glob()` uses `fast-glob` so I believe also supporting the V0.4 design doesn't add much complexity. That said, if omitting the V0.4 design makes our life easier, then that'd be more than fine because we're going to release the V1 design in the next coming days and eventualy deprecate the V0.4 design.

vike/node/plugin/shared/readGitignore.ts
brillout: Yes, if supporting `.gitgnore` (or only a subset) turns out to be too complex, then I guess a new config is the only option. (Feel free to object but I think a config defined in `vite.config.js` instead of `.vikeignore` would be better.)

The neat thing of using `.gitignore` is that it "just works" for the user, doubling down on Vike's zero-config nature. But, yea, if it's too complex / time consuming to implement then let's go for a new config.

So I think the bottom line here is: is supporting `.gitignore` too complex / time consuming?

AurelienLourot: Also note that there are edge cases where using `.gitignore` won't be what the user wants, for example imagine the following setup:

```
# .gitignore
/pages/autogenerated
```

```
// package.json
{
  "scripts": {
    "build": "generate /pages/autogenerated && vite build",
// ...
```

Obviously in this case the user won't be happy that the pages in `/pages/autogenerated` aren't found by vike.

I vote for going away from the idea of parsing `.gitignore` and instead having our own `ignoreDirs` (or similar) vite config

brillout: I did think of this, but only a handful of users are doing that. My thinking was that, in the long term and depending on what needs arise, we can have both a config `crawl: { skipGitignore: boolean, ignore: string[] }` as well as applying `.gitignore` (or only a subset of it) by default.

For example, a simple subset would be to only apply trivial patterns from `.gitignore`:

```
# .gitignore
dist/ # <-- this one is picked
/some/precise/path/to/data/ # <-- this one is picked as well
/foo/**/boo # <-- Ignore? Or maybe we can pick this as one as well while checking if a `!` line negates it
```

Maybe, with a new Vitest `.spec.ts` suite of unit tests, this can be made reliable?

@nitedani WDYT?

nitedani:
I removed the code from v0.4 paths, and only read the `.gitignore` in config.root. Correctly parsing all of the gitignore files in a monorepo would be too difficult.

vike/node/plugin/shared/readGitignore.ts
nitedani: Yes, I tried to write an algorithm to parse all of the gitignore files in the monorepo, and while it worked, it would be too heavy to maintain. So I think we should only use a subset of the gitignore in config.root.

AurelienLourot (approved):
Looks good to me, thanks @nitedani

brillout:
> I removed the code from v0.4 paths

👍 

> Correctly parsing all of the gitignore files in a monorepo would be too difficult.

Do you mind elaborating?

I believe the current implementation is faulty given the following:

```
/generated
!/generated/pages/
```

While supporing only a subset is fine, I think ignoring too much is a no-go.

I wonder: does `fast-glob` support the following distinction?

```
# Ignore all dist/ folders
dist/
# Only ignore the dist/ folder at the root
/dist
```

The discrepancy between `.gitignore`'s patterns and `fast-glob`'s patterns is error prone; I'm inclined to think that a new `.spec.ts` with thorough tests is needed.

If you believe it's too much effort, feel free to abandon this PR.

AurelienLourot:
> > Correctly parsing all of the gitignore files in a monorepo would be too difficult.
> 
> Do you mind elaborating?

I think @nitedani mean supporting `.gitignore` in subfolders of the vite root folder is too complicated, as I described earlier, because suddenly you have to apply conversions like `/foo/.gitignore : bar` -> `/foo/**/bar`. So @nitedani decided to stick to supporting only one .gitignore file at the vite setup's root

brillout:
Makes sense 👍. And how about that?

```
# monorepo root
package.json
.gitignore
# Vike app
packages/my-app/package.json
```

Alternatively, I'd be 👍 for merging a very simple `.gitignore` support (which would already benefit a *lot* of users). But at least we should fix this: 

> I believe the current implementation is faulty given the following:
> 
> ```
> /generated
> !/generated/pages/
> ```

Depending on whether @nitedani (or someone else) wants to, we can (now or later) implement a more thorough implementation with thorough unit tests.

brillout:
Actually... how about we use `git ls-files` instead of `fast-glob`? That would be simple, reliable, and fast.

We can fallback to `fast-glob` if the user doesn't use Git.

@nitedani FYI if you want you can also push a bunch of commits instead of force-pushing a single commit &mdash; as you wish and whatever works best for you. (Your `readGitignore.spec.ts` is really neat 💯, although I guess we don't need it anymore if we go for the `git ls-files` approach.)

vike/node/plugin/shared/readGitignore/index.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    if (line.endsWith('/')) {
+      line = `${line}**`
+    }
+
+    const slashPosAfter = line.indexOf('/')
+    if (slashPosAfter === -1) {
+      const startsWithAsterisk = line.startsWith('*')
+      const endWithAsterisk = line.endsWith('*')
+      while (line.startsWith('*')) line = line.substring(1)
```
AurelienLourot, line 66: We remove all leading `*` here, right? That means we transform `*foo` into `foo`, and `startsWithAsterisk` is `true`, so later on line 78 we will turn it into `**/*foo*`, which is not the same, right? a `barfoobaz` file wouldn't have matched the original `*foo` expression, but now it will match the new `**/*foo*` expression.

I suspect we can come up with other problems here and I think Rom's idea of using an already existing tool for this instead feels safer. He suggested for example `git ls-files` but I discovered also `git status --ignored`. I have played a bit with it and it seems to be exactly what we want. See also https://stackoverflow.com/questions/466764/git-command-to-show-which-specific-files-are-ignored-by-gitignore

It seems to cover all the edge cases we have mentioned so far, like having .gitignore in ancestor folders or child folders. It won't yield any star or any exclamations marks. It may yield lines starting with `../` if the `.gitignore` comes from an ancestor folder, and all we have to do is skip them. We could then be quite confident to have a working implementation. What do you think?

AurelienLourot: `git status --ignored [--porcelain]` isn't perfect either I just discovered, sorry I stand corrected, but `git ls-files` instead of `glob()` as suggested by Rom seems to be a viable option indeed

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
```diff
@@ -1111,20 +1116,48 @@ async function findPlusFiles(
...
+
+  let result: string[] = []
+
+  // If there is no .git folder, this will error
+  try {
+    const gitIncludePatterns = scriptFileExtensionList.map((ext) => `"**/+*.${ext}"`)
+    // -o lists untracked files only(but using .gitignore because --exclude-standard)
+    // -s adds the tracked files to the output
+    const { stdout } = await execA(
+      `git ls-files ${gitIncludePatterns.join(' ')} -os --exclude-standard --exclude="**/node_modules/**"`,
```
AurelienLourot, line 1128: Why do we need to exclude `node_modules`? Isn't it in `.gitignore` anyway?

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
```diff
@@ -1111,20 +1116,48 @@ async function findPlusFiles(
...
+      {
+        cwd: userRootDir
+      }
+    )
+
+    result = stdout
+      .split('\n')
+      .map((line) => line.split('\t').pop() || '')
```
AurelienLourot, line 1136: It seems like you need this because you are using `-os --exclude-standard`. It seems like everything is fine if you use `-oc` instead. Do you confirm?

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
AurelienLourot: My mistake, I understand, it's for the case in which there is no `.gitignore` in the repo :+1:

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
```diff
@@ -1111,20 +1116,48 @@ async function findPlusFiles(
...
-      ...ignorePatterns
-    ],
-    cwd: userRootDir,
-    dot: false
-  })
+
+  let result: string[] = []
+
+  // If there is no .git folder, this will error
+  try {
```
AurelienLourot, line 1123: This looks great, do you think we could extract this logic to a separate `glob()` function that we call here?

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
nitedani: -oc works, nice

AurelienLourot (approved):
LGTM, thanks @nitedani

brillout:
I did some polishing. I fixed two issues around pattern matching, and added a neat trick: fallback to `fast-glob` if `git ls-files` doesn't find any plus file (this should cover most users that generate plus files).

@nitedani @AurelienLourot Let me know if this is LGTM to you and I'll merge 🎉

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
```diff
@@ -1170,6 +1184,43 @@ async function findPlusFiles(
...
+    throw err
+  }
+
+  let files = stdout.split('\n').filter(Boolean)
+
+  // --exclude only applies to untracked files
+  assert(!outDir.startsWith('/'))
+  files = files.filter(
+    (line) => !line.includes('node_modules/') && !line.includes('.telefunc.') && !line.startsWith(`${outDir}/`)
+  )
```
AurelienLourot, line 1219: Aren't we excluding all this already on lines 1191-1198 anyway?

AurelienLourot (approved):
LGTM, just one nitpick but feel free to merge anyway :pray:

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
nitedani: Those `--exclude` only applies to the untracked files. So if, for some reason `outDir` is committed and tracked, we need to filter it here.

brillout: Yep. (And FYI that's why I decided to not factor out the patterns between `fast-glob` and `git ls-files`.)

I improved the comment to make it clearer why we repeat the exclusion logic.

brillout:
Merged 🎉

Thanks @nitedani & @AurelienLourot, it was a nice PR "sparring session" 🦾

It's very neat that the crawling is now both more performant and more clever.

More PRs very much welcome. Also about a `vike-react-query` package and, in general, further Vike packages for automatic tool integrations beyond `vike-{react,vue,solid}`.

PR #1265, opened 2023-11-18
fix double call to onBeforeRender

AurelienLourot:
If `pageContextInit` has some client data, and onBeforeRender's env is `{ server: true, client: true }`, then the onBeforeRender hook was being called both on server- and client-side upon navigation.


vike/client/client-routing-runtime/getPageContextFromHooks.ts
```diff
@@ -165,8 +170,9 @@ async function getPageContextAlreadyRouted(
     }
   }
 
-  {
+  if (!pageContextFetchedFromServer) {
```
AurelienLourot, line 173: this is the fix. All the rest of the PR is just comments and typescript-happiness

brillout:
👌

Issue #962, opened 2023-06-19
New hook `onBoot()`


brillout:
### Description

A common use case is to fetch global initialization data that is needed by all pages. (For example i18n data.)

```js
// /pages/+onBoot.js

export default onBoot

async function onBoot() {
  const someInitData = await fetchSomeInitData()
  return { pageContext: { someInitData } }
}
```
The `onBoot()` hook is executed as soon as the server is started and its result is awaited before any page starts rendering.

It's executed when the production/development/preview server starts as well as when the pre-rendering process starts. So it's guaranteed to have been executed before any page starts (pre-)rendering.

Workaround until this is implemented:
 - Add your initial data at [`renderPage()`](https://vike.dev/renderPage), see [API > `pageContext` > Custom](https://vike.dev/pageContext#custom).
 - Use a single global [`onBeforeRender()`](https://vike.dev/onBeforeRender) hook that applies to all your pages:
   ```js
   const cache = {}
   // Or:
   // globalThis.cache = {}
   export async function onBeforeRender(pageContext) {
     if (!cache.someData) {
      cache.someData = await fetch(...)
     }
     pageContext.someData = cache.someData
   }
   ```
 - If you already use `onBeforeRender()`, then create a new [custom hook](https://vike.dev/meta)

Additional consideration:
 - Also for the client-side, by defining `+onBoot.client.js`.
   - `+onBoot.js` => server-side only.
   - `+onBoot.client.js` => client-side only.
   - `+onBoot.shared.js` => server- and client-side. (I can't think of a use case, but supporting `.shared` wouldn't induce any additional implementation cost.)
   - You could even define both `+onBoot.js` and `+onBoot.client.js`.

See also:
 - https://vike.dev/client (which is an alternative if you don't need to populate `pageContext`)
 - https://github.com/vikejs/vike/issues/1908
 - https://github.com/vikejs/vike/issues/2361
 - https://github.com/vikejs/vike/issues/1613

If not having `onBoot()` is a blocker for you then write a comment down below.

brillout:
Alternative name: `onServerInit()`.

gdev219:
that is great I was customizing the express server and fetch some data before it renders the page. this will be very helpful thanks!

brillout:
@gdev219 👍 Let me know if not having `onServerStart()` is a blocker. (Since you have an Express.js server, you should be able to easily workaround it.)

brillout:
[...]

PR #1267, opened 2023-11-19
feat: data hook

AurelienLourot:
Initial implementation. Should be working but I haven't actually tested it, so opening as a draft for now until I at least push a working example. Feedback already welcome :pray:


brillout:
💯 I will have a look at it after we merge the V1 docs.

brillout:
Just had a quick high-level look at it: looks good so far 👍. I will have a closer look after the V1 docs merge and/or after you feel this PR is ready.

vike/shared/assertDataHookReturn.ts
```diff
@@ -0,0 +1,20 @@
...
+): asserts hookReturnValue is undefined | null | { [key in Keys[number]]?: unknown } {
+  if (hookReturnValue === undefined || hookReturnValue === null) {
+    return
+  }
+  const errPrefix = `The data() hook defined by ${hookFilePath}` as const
+  assertUsage(
+    isPlainObject(hookReturnValue),
+    `${errPrefix} should return a plain JavaScript object, ${pc.cyan('undefined')}, or ${pc.cyan('null')}`
+  )
+  assertHookReturnedObject(hookReturnValue, [] as const, errPrefix)
```
phonzammi, line 19: Hi @AurelienLourot, what about these `[] as const`, what should we put inside these array?, Is it ['data'] ?

AurelienLourot: This assertion is wrong actually, good catch. Any key should be allowed here. The idea is that a `data()` hook that looks like

```js
function data(pageContext) {
  // ...
  return {
    foo: 42,
    bar: 'baz'
  }
}
```

will result in a `pageContext` that looks like

```js
{
  data: {
    foo: 42,
    bar: 'baz'
  }
}
```

I'll fix that :pray:

brillout: (@phonzammi Btw. we added a couple of tickets to https://github.com/vikejs/vike/issues?q=is%3Aissue+is%3Aopen+label%3A%22contribution-welcome+%3Atwo_hearts%3A%22 in case that's something you'd be interested in.)

phonzammi: Sure, 👍

phonzammi: > Btw. we added a couple of tickets to https://github.com/vikejs/vike/issues?q=is%3Aissue+is%3Aopen+label%3A%22contribution-welcome+%3Atwo_hearts%3A%22 in case that's something you'd be interested in.)

I'll see if I can, I am still junior

AurelienLourot:
I'm happy with this PR now. It is covered with E2E tests and I'm confident it works as intended.

I want to wait for #1206 to land before I write the docs and an example, and I'm fine doing that in a follow-up PR. So as far as I'm concerned this can be merged

vike/shared/assertDataHookReturn.ts
brillout: > I'll see if I can, I am still junior

I you're willing to dig on your own (e.g. trying to become knowledgeable around about the topics of your PR) then it's totally fine. You can have a try and we see how it goes if you want. Looking forward to some PR "sparring" 🦾

phonzammi: Thanks @brillout 
i really appreciate that

brillout:
FYI I started reviewing this.

One issue Vike's repo has is that is has `78` monorepo packages, which massively slows down E2E tests and `pnpm`. So I'm trying to reduce the number of packages we create. For example I started merging bunch of tests into `/test/misc/`.

I wonder: could we maybe merge the unit tests (which is really neat) of `/test/data-hook/` and `/test/hook-override/` into a single test setup?

> ```
> // NOTE(aurelien) `data` may not exist at all if the user doesn't use `data` hooks, thus the `?`.
> data?: Data
> ```

I thought for a moment whether Vike should always set `data` to an object, but I actually like that `data` is `undefined` if there is, well, no data. So I'm thinking it's good like like that 👌.

AurelienLourot:
Your last commit looks good, thanks!

Yes I'll merge `/test/data-hook/` and `/test/hook-override/`, good idea

brillout:
I wonder about the execution order. I struggle to find a sensible order. I'm actually thinking maybe we shouldn't execute the onBeforeRender hook if there is a data hook: that way we can postpone that ordering decision, and take the decision later when use cases that need both hooks pop up. But I'll mull over it a bit more and maybe a sensible order will occur to me.

brillout:
I think it's a net win for the user if Vike executes both. I thought about superseeding only because it would allow us to postpone the ordering decision.

> 2\. ([New hook `onStartup()` #962](https://github.com/vikejs/vike/issues/962), [Global /renderer/+onBeforeRender and Local /page/+onBeforeRender behavior Vike V1 Client Routing #1226](https://github.com/vikejs/vike/discussions/1226),

Neat, thanks for that list of use cases, that's very convenient.

The first use case needs `onBeforeRender()` to be called before `data()`, whereas the second use case needs the order to be the other way around.

I'm thinking, a simple solution would be to add one more hook `onBeforeData` with following order:
 1. `onBeforeData`
 2. `data`
 3. `onBeforeRender`
 4. `onRender{Html,Client}`

So my proposal would be this: we change this PR to always execute `data` before `onBeforeRender`. Then later, in another PR, we implement `onStartup()`. After that and if the need eventually arises, we implement `onBeforeData`.

Thoughts? Do you foresee an issue with executing `data` before `onBeforeRender`?

AurelienLourot:
Thanks, this sounds good and makes sense. I'll swap the order

AurelienLourot:
I'm happy with this PR :)

brillout:
👌 I'll have a second look at it. (It didn't review everything last time so new questions may pop up.)

brillout:
- I'm thinking a hook `useData()` that is just an alias for `usePageContext().data` could be neat.
   ```ts
   // The Data generic is required, see example below.
   const data = useData<Data>()
   /* Equivalent to:
   const { data } = usePageContext() as { data: Data }
   */
   ```
 - One neat thing about the data hook is that the user doesn't have to think in terms of `pageContext` and, instead, directly thinks in terms of data; that's why I'm thinking an alias `getData()` could be neat for that reason as well: we abstract away `pageContext`.
 - Since each page has its own `Data` structure, I wonder how much sense `Vike.Data` makes. Maybe it would make more sense to make `pageContext.data` globally `undefined | Record<string, unknown>` while forcing the user to always specify Data at the relevant places (at `useData()` and `+data.js`).
 - I guess TypeScript doesn't have the notion of filesystem precedence 🤔 so I'm inclined to think that the user does always need to specify `Data` at the relevant places.
 - I thought about renaming `+data.js` to `+load.js` but I'm hesitating &mdash; I guess `+data.js` is just fine.

Example:

```ts
// /pages/star-wars/types.ts

export type { Data, Movie }

type Data = {
  movies: Movie[]
}
type Movie = {
  id: string
  title: string
  release_date: string
}
```

```ts
// /pages/star-wars/+Page.ts

export { Page }

import React from 'react'
import type { Data } from './types'

// Note how we don't need the whole pageContext.pageProps thing
function Page() {
  return (
    <>
      <h1>Star Wars Movies</h1>
      <MovieList />
    </>
  )
}

function MovieList() {
  // The Data generic is required
  const data = useData<Data>()
  const { movies } = data
  return (
    <ol>{
      movies.map(({ id, title, release_date }) => (
        <li key={id}>
          <a href={`/star-wars/${id}`}>{title}</a> ({release_date})
        </li>
      ))
    }</ol>
  )
}
```
```ts
// /pages/star-wars/+data.ts

export { data }

import type { Data, Movie } from './types'
import type { DataAsync } from 'vike/types'

// Make the Data generic required here as well? It's useless but it educates the user 🤔
const data: DataAsync<Data> = async (
// While this data hook doesn't need routeParams, it's a frequent use case.
{ routeParams }
): Data  => {
  const response = await fetch('https://star-wars.brillout.com/api/films.json')
  const movies: Movie[] = ((await response.json()) as any).results
  return movies
}
```

AurelienLourot:
> * I'm thinking a hook `useData()` [...]

Super good ideas, thanks a lot. I'll do that. I'll see if I can modify an existing example to make use of `useData()` (an example that isn't already pointed at by the documentation in order to illustrate pageContext). If you have one in mind, let me know. If I can't find any, I'll write a new example.

brillout:
👍 Let's see how/if we can remove/merge examples/tests later.

AurelienLourot (commented):
LGTM :100:

AurelienLourot (commented):
LGTM, I'll turn `DataAsync` into a generic `DataAsync<...>` type as discussed

AurelienLourot:
Done :)

brillout:
Neat :)

I'm hesitating a bit whether my idea was good.

WDYT? Which do you prefer?

```ts
const data: DataAsync = async (pageContext): Promise<Data> => {
  // ...
}
```
```ts
const data: DataAsync<Data> = async (pageContext): Promise<Data> => {
  // ...
}
```
```ts
const data: DataAsync<Data> = async (pageContext): ReturnType<DataAsync<Data>> => {
  // ...
}
```

Can't wait to merge :)

AurelienLourot:
I prefered `const data: DataAsync = async (pageContext): Promise<Data> => {`

But `const data: DataAsync<Data> = async (pageContext): ReturnType<DataAsync<Data>> => {` has the advantage, as you said offline, to keep the usual hint about return type that we have on other hooks.

So either of these two choices are fine by me.

brillout:
Makes sense. There is also the third option of making the generic `unknown` by default. Consequently, the user can choose which one of two variants he wants to use. I think I like this most, but I'll mull a bit over it 💭

brillout:
How about this: https://github.com/vikejs/vike/pull/1267/commits/d47fa19f01ffe1423eef91b64d7a729a3e7b34d6?

AurelienLourot:
Nice move, I'd do https://github.com/vikejs/vike/pull/1267/commits/553b070ed4 on top of that. WDYT?

AurelienLourot:
It keeps the important part IMHO of your change, being: `+Page` imports the `Data` type from `./+data`

brillout:
I'm thinking inferring can be quite important. Let me elaborate.

In our examples declaring instead of inferring `type Data` inside `+data.ts` is easy, but, if the user uses a tool such as Prisma with database query results that are automatically typed, then it can be cumbersome to declare instead of inferring `type Data`. Even very cumbersome for ORM queries that use `.select()`.

Also the neat thing about inferring is that you can factor/move data fetching logic out of `+data.ts` without breaking `type Data`. Otherwise you'd need some brain power to preserve the declared `type Data` when refactoring.

AurelienLourot:
Makes sense. So if we're going to infer `Data` from the return type, that means any return type is correct. There is no need to do type checking on the return type, thus there is no need for the generic. Let me push a commit according to that

AurelienLourot:
No I understand now. I couldn't come up with anything better than what you did before. Just reverted my last commit.

brillout:
I suggest we:
 - Always showcase the inferring way in the docs and examples.
 - We keep the `DataAsync<Data>` type for folks who want 1. the `Data` type to be defined next to and by the UI component and 2. want to re-use the usual recommendation.

This PR now LGTM. As always, feel free to disagree & nitpick.

brillout:
Merged 🎉

Big 🙏 for the quite complex PR. I'm thrilled: it's the first PR implementing a major feature 💯

I can't wait for us to release these new lovely `data()` hooks 👀 It's quite a big DX improvement.

AurelienLourot:
@brillout can we release this as v0.4.151? I'm implementing `useData()` in vike-react and need this. Thanks :pray:

brillout:
I ain't sure, but I think it's actually already released in `0.4.150`. (IIRC we didn't add a `feat: ` commit so it's missing in the changlog.)

Anyways, I just pre-released `main` as `vike@0.4.150-commit-e6c3150`. Let me know if that isn't enough.

AurelienLourot:
@brillout no it's not in 0.4.150 if you look at the git history. It got merged later. And the first commit says `feat: data hook` 

I don't want to write `^0.4.150-commit-e6c3150` in vike-react, so eventually I'd need it to be released. No rush though, but just curious: is there any specific reasons why you're not comfortable releasing 0.4.151?

brillout:
This release isn't as trivial as usual because Telefunc needs to be co-released (the shared dependency https://github.com/brillout/vite-plugin-import-build has a breaking change).

But let me see if I can release now.

brillout:
`0.4.151` released. (FYI I had to manual fix the changelog; we added some `fix: ` and `feat: ` prefix where we shouldn't. Although sometimes it isn't always easy to know when to set them, e.g. I couldn't know whether I would update `vite-plugin-import-build` again in the future.)

AurelienLourot:
Thanks :) Yes I'll be more careful too with my commit messages with regards to the changelog

PR #1271, opened 2023-11-21
Fix partial hydration while SSR streaming

AurelienLourot:
with `useAsync()` and `<Suspense>`.

FYI @nitedani


examples/react-streaming-v1/pages/star-wars/index/+Page.jsx
```diff
@@ -0,0 +1,57 @@
...
+import { useAsync } from 'react-streaming'
+import fetch from 'cross-fetch'
+import { usePageContext } from '../../../renderer/usePageContext'
+import { Counter } from '../../../components/Counter'
+
+function Page({ movies }) {
+  return (
+    <>
+      <h1>Star Wars Movies</h1>
+      Interactive while loading: <Counter />
```
AurelienLourot, line 13: Unfortunately this isn't true: it's not interactive until the MovieList has fully streamed. So it gets hydrated only once the streaming has finished :/

Same in the already existing `cloudflare-workers-react-full-v1` example.

Is it a known issue and/or am I missing something?

brillout: It used to work. I guess there is a regression somewhere. (Inspecting what stream chunk is sent when is usually quite insightful for what the problem is.)

AurelienLourot: Ok thanks for the reply, I'll dig a bit more myself :pray:

nitedani: https://github.com/vikejs/vike/blob/19e44c2e2a1e2eb95ea981f0493d0ad870d34bff/vike/node/runtime/html/injectAssets/inferHtmlTags.ts#L31

this is the issue

nitedani: Was this changed to defer, because the pageContext can be async?

AurelienLourot: I have been playing with this but I'm a bit struggling. Locally I have turned all `defer` back into `async` and at the place where we are accessing the pageContext (https://github.com/vikejs/vike/blob/main/vike/client/shared/getPageContextSerializedInHtml.ts#L8) I inserted a loop waiting for the pageContext to show up:

```ts

async function getPageContextSerializedInHtml(): Promise<{ _pageId: string; _hasPageContextFromServer: true }> {
  const id = 'vike_pageContext'

  // Wait for `vike_pageContext` to be parsed:
  let elem: HTMLElement | null = null
  for (let remainingAttempts = 2000; remainingAttempts > 0; --remainingAttempts) {
    elem = document.getElementById(id)
    if (elem !== null) break
    await new Promise((resolve) => setTimeout(resolve, 10))
  }

  assertUsage(
    elem,
    `The element #${id} (which vike automatically injects into the HTML) is missing from the DOM. This may happen if your HTML is malformed. Make sure your HTML isn't malformed, and make sure you don't remove #${id} from the HTML nor from the DOM.`
  )
```

and of course I decorated the calls to `getPageContextSerializedInHtml()` with `await`. 

Unfortunately the `<script>` making use of this comes in the HTML before the `pageContext` JSON `<script>` and my await-loop blocks rendering, so the `pageContext` never shows up, so all examples are broken. According to https://addyosmani.com/blog/script-priorities/ I wouldn't have expected it to fully block the parser, but instead to give the parser opportunities to continue whenever the loop is sleeping 10 ms. Same problem if I increase the sleep to 1000 ms.

I have spent hours on this actually and I can't come up with a solution except if I put the pageContext `<script>` even earlier in the HTML. Is this really the only solution? Am I missing something? Are there arguments against putting the pageContext `<script>` very early in the HTML? Input welcome @nitedani @brillout. Happy to push my attempt here if this helps :pray:

AurelienLourot: Ok I think I understand, here is another nice explanation: https://stackoverflow.com/a/39711009/1855917

(I'm now a bit skeptical about the distinction between "interrupting" and "blocking" the parser in https://addyosmani.com/blog/script-priorities/ - it seems to be the same)

The key is that the parser always completely freezes until the script has fully executed (no matter if `defer`, `async`, or nothing, the parser will never proceed while a script has been started). So my await-loop prevents the script from terminating, so freezes the parser and prevents it from reaching the pageContext JSON `<script>`.

Does anything speak against this easy solution? putting at the end of the body first the pageContext JSON `<script>`, followed by the entry `<script defer>`, the followed by the stream's `<script>`s. I don't see a reason why having the entry `<script>` in `<head>` would be better. We used to have it there (at least in dev), but with `defer`, so it would be executed only after the end of the body is reached anyway.

I think I'll go in this direction. Thanks for listening :pray:

brillout: > await-loop prevents the script from terminating, so freezes the parser

I didn't expect that, good to know.

I'm thinking maybe Vike should insert things differently depending on whether the HTML is being streamed 🤔

AurelienLourot: > I'm thinking maybe Vike should insert things differently depending on whether the HTML is being streamed

It doesn't make a difference in terms of performance to differentiate both cases. It would complicate the code without gains, except if I'm missing something. Let me lay it out. 

First of all, the position of the `<script>` in the document has an impact on when downloading the script would start, but this is irrelevant for us because we're talking about inline scripts. So it's really all about when we want the script to be executed.

Here is the train of thoughts:
1. We can't use `defer` when streaming because it breaks progressive hydration for the reasons we know. So special care in this case at least needs to be taken so the entry script runs after the pageContext JSON script. The only way is
    a. the JSON script must come before the entry script in the HTML, and can't be `async` (scripts that are depended upon by other scripts can't be async, because no guarantees can be made on ordering when using async)
    b. both scripts could be placed in `<head>` but I think it's counter-productive. I'd rather have the rest of the HTML be rendered first. I don't see why I would try to prioritize hydrating code higher up. I'm more inclined to have these two scripts at the bottom of `<body>` (before the streaming)
2. Now without streaming, yes you're right in theory we could put `defer`, but it has no gain because our scripts are already at the bottom of `<body>`

Let me know if I'm missing something. I'll push a fix here soon with useful comments at strategic places, also referring to this conversation.

Thanks for you're input. It's quite helpful for me to lay everything out here and think it through.

AurelienLourot: I have just pushed a fix according to all this and tested it thoroughly:
* on react-full-v1 dev
* on auth dev
* on auth prod
* on react-streaming-v1 dev (partial hydration works)
* on react-streaming-v1 prod (partial hydration works)

I'm pretty confident that this fix is correct and I tried to make the inline comment as clear as possible. I expect a few tests to turn red as I didn't run them locally and I'm sure I'll have a few test expectations to adjust.

AurelienLourot:
Alright, I'm happy with this PR :)

examples/react-streaming-v1/pages/star-wars/index/+Page.jsx
AurelienLourot: And thanks a lot to @nitedani for having figured out the culprit in the first place! (`defer`)

brillout: Thanks for digging into all that, it's a *massive* help 🙏

Two thoughts:
 - There seems to be a conundrum when `pageContext` values are determined *during* the stream, see https://vike.dev/stream#initial-data-after-stream-end. In particular, I'm not sure Vike can detect whether `onServerPrefetech()` is being used to determine `pageContext` values. I've added `test/vue-stream-onServerPrefetch/` to play around that (I guess its tests are green because the example is "lucky" to determine `pageContext.initialStoreState` early enough, but I think the tests can be made red by pushing more things to the stream).  
 - The `<link rel="modulepreload">` tags should be injected in `<head>` (I just checked and they are being injected along the `<script>` tags which makes these preload tags useless).

AurelienLourot: > * The `<link rel="modulepreload">` tags should be injected in `<head>`

Interesting because `main` does it already "wrong" then, i.e. `main` already puts `<link>` in `<body>`, see test expectations: https://github.com/vikejs/vike/blob/main/test/preload-v1/__snapshots__/prod.spec.ts.snap#L60

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link) though:

> A `<link>` element can occur either in the [`<head>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head) or [`<body>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body) element, depending on whether it has a [link type](https://html.spec.whatwg.org/multipage/links.html#body-ok) that is body-ok

And `modulepreload` is body-ok: https://html.spec.whatwg.org/multipage/links.html#body-ok

So I would leave this as is. What do you think?

----------

> * There seems to be a conundrum [...]

OK I'll wrap my head around this topic and address that. Thanks for the super useful hints!

brillout: Ah, you're right, and I just remembered why I did that: AFAIK there is no way to tell the browser "here are all the assets the page needs but load JS only after you're done loading evertyhing else" which is quite a bummer. So it makes sense to delay injecting all `<script>` tags as much as possible so that it doesn't slow down the critical path consisting of HTML, CSS, and images.

Good catch and thanks for pushing back on that 👌

> OK I'll wrap my head around this topic and address that.

The dilemma can be summed up like this: in `test/vue-stream-onServerPrefetch/` the value of `pageContext.initialStoreState` can be determined anywhere during the stream (using Vue's `onServerPrefetch()`) and potentially after the stream writes the `pageContext` JSON `<script>` but the `pageContext` JSON needs to contain the value of `pageContext.initialStoreState`.

brillout: >  is what happens:

That's surprising, even more so because it sets `enableEagerStreaming` to `true`.

PR #1274, opened 2023-11-22
Minor update examples/auth

phonzammi:
Minor refactor of examples/auth

Since onRenderHtml() of vike-react v0.3.4 has condition : [isSsrDisabled](https://github.com/vikejs/vike-react/blob/24ebe5902e7a86b2145af280574530098647f30c/vike-react/renderer/onRenderHtml.tsx#L36C3-L55C78)

If we don't create our own server/index.ts file (server entry if i am not mistaken), it will just works.

But if we do create custom server/index.ts (like in this example) then we have to return
[stream-pipe](https://vike-v1-staging.lourot.dev/stream#stream-pipes)
```
const { statusCode, headers, earlyHints } = httpResponse
if (res.writeEarlyHints) 
res.writeEarlyHints({ link: earlyHints.map((e) => e.earlyHintLink) })
headers.forEach(([name, value]) => res.setHeader(name, value))
res.status(statusCode)
httpResponse.pipe(res)
```
or [stream-to-string](https://vike-v1-staging.lourot.dev/stream#stream-to-string)
```
const body = await httpResponse.getBody()
      if (typeof body !== 'string') {
        return next()
      }
headers.forEach(([name, value]) => res.setHeader(name, value))
res.status(statusCode).send(body)
```

cmiiw
_**no need to merge in case of mess**_

### DIFF
--- a/examples/auth/package.json
+++ b/examples/auth/package.json
@@ -16,9 +16,9 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "typescript": "^5.0.2",
-    "vike-react": "^0.3.2",
-    "vite": "^4.0.3",
-    "vike": "0.4.147"
+    "vike": "0.4.147",
+    "vike-react": "^0.3.4",
+    "vite": "^4.0.3"
   },
   "type": "module"
 }
--- a/examples/auth/server/index.js
+++ b/examples/auth/server/index.js
@@ -72,7 +72,11 @@ function vike(app) {
     if (!httpResponse) {
       return next()
     } else {
-      const { body, statusCode, headers } = httpResponse
+      const { statusCode, headers } = httpResponse
+      const body = await httpResponse.getBody()
+      if (typeof body !== 'string') {
+        return next()
+      }
       headers.forEach(([name, value]) => res.setHeader(name, value))
       res.status(statusCode).send(body)
     }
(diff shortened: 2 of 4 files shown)


brillout:
Good catch 👍.

I'd use `.pipe()` so that we preserve the stream. The CI is failing; have a look at the warning printed there.

phonzammi:
Thanks @brillout,
> I'd use .pipe() so that we preserve the stream

I'll change to .pipe().
> The CI is failing; have a look at the warning printed there.

I'm testing it on local right now

phonzammi:
> The CI is failing

I've checked, I think it's failing to find [hasText: 'Neumann'](https://github.com/vikejs/vike/blob/74d84f54c2f024cd9b47b1fa7582dfb710003035/examples/auth/.testRun.ts#L24) 

by adding userAgent fixed it

brillout:
Neat 💯

brillout:
Looking forward to further PRs 💫

phonzammi:
Thanks, I just learning & trying from little problem to be fixed, I hope I can help make a few good feature for vike ❤️

brillout:
Sounds good 🙂. If you want we can create a new label `easy-contribution 🌱` along `contribution-welcome 💕`.

phonzammi:
Well that would be nice, more specific for newbies like me

PR #1282, opened 2023-11-23
examples: fix stale pageContext when hitting back button

### LINKED ISSUE #1275: Unexpected pageContext behaviour when navigating to page via the URL bar and back
### Description

I ran into this issue when implementing authentication following the [auth example](https://github.com/vikejs/vike/tree/main/examples/auth).

After logging in the user object is defined in pageContext and the Vike reload function is called. From there, if the user navigates to any page by typing in the URL in the address bar and pressing enter and then navigates back, the user object which was set in pageContext becomes null.

I'm using the React preset and V1 just like in the auth example. In fact, the same issue can be reproduced in the example by doing the following:

1. Load up the app and navigate to the log in page `/login` via the URL bar.
2. Log in (doesn't matter which one).
3. Navigate to `/` via the URL bar again.
4. Navigate back.

Thanks :)

AurelienLourot:
In `examples/auth`, fix stale `pageContext` when hitting the back button in the browser. Fixes #1275

CC @jgalminas

### DIFF
--- a/examples/auth/server/index.js
+++ b/examples/auth/server/index.js
@@ -72,11 +72,17 @@ function vike(app) {
     const pageContext = await renderPage(pageContextInit)
     const { httpResponse } = pageContext
     if (!httpResponse) {
+      // not a page
       return next()
     } else {
       const { statusCode, headers, earlyHints } = httpResponse
       if (res.writeEarlyHints) res.writeEarlyHints({ link: earlyHints.map((e) => e.earlyHintLink) })
       headers.forEach(([name, value]) => res.setHeader(name, value))
+
+      // Disable browser-caching for all pages. This prevents stale pageContext in browser when hitting the back button.
+      // Relax this if your app allows. See https://github.com/vikejs/vike/issues/1275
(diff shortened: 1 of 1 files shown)


brillout:
I wonder if this shouldn't be the default? For example, we can add this to `pageContext.httResponse.headers` whenever the HTML contains the `pageContext` JSON script.

Alternatively, one could argue that it's just how browsers work and that it's expected that the user should hit the reload button.

AurelienLourot:
> I wonder if this shouldn't be the default? For example, we can add this to `pageContext.httResponse.headers` whenever the HTML contains the `pageContext` JSON script.

Good idea :+1: On top I would create a config for the user to be able to change this behavior. For some pages the user knows that the pageContext is always going to be the same and may want to allow caching. What do you think?

brillout:
Sounds good 👍. Correct first, then performant.

AurelienLourot:
Awesome, let me reject this PR and I'll come up with a better one soon

brillout:
👍 If you do the PR before the V1 design release, then I think we can push the docs directly to V1 docs branch (same for the data hook docs btw.).

PR #1292, opened 2023-11-26
Enable any protocol support for redirects

### LINKED ISSUE #1245: Redirects: Support URLs with other protocols such as `mailto:` or `ipfs://`
### Description

Currently, redirects only support URLs of type `http`, `https` and local routes.

I propose adding support to those of type `mailto` as well.

Discussion about the topic started on https://github.com/vikejs/vike/discussions/1244.

### Motivation
- `mailto` URLs are the next more common type of emails. Some people might want to setup a certain route to a `mailto`.
- For example, in order to be easily contactable, I might setup a route in my personal website called `/mail-me` and having it redirect to my current email, as opposed to giving my email to a person (which might be a complex email prone to error when typing) I can simply give an easy to memorize link
- On top of that, `mailto` urls also support a bunch of parameters such as setting the default body and subject — which might be useful for certain use-cases, such as job applications, etc.

usk94:
resolves #1245

This PR introduces the capability to utilize the mailto protocol within the redirect configuration. 
This allows for configured path to be redirected to email addresses, triggering the opening of the user's default email client. 

config example
```javascript
// vite.config.js

import vike from 'vike/plugin'

export default {
  plugins: [
    vike({
      redirects: {
        '/email': 'mailto:foo@bar.test'
      }
    })
  ]
}
```

### DIFF
--- a/vike/shared/route/resolveRedirects.ts
+++ b/vike/shared/route/resolveRedirects.ts
@@ -23,14 +23,10 @@ function resolveRedirects(redirects: Record<string, string>, urlPathname: string
 function resolveRouteStringRedirect(urlSource: string, urlTarget: string, urlPathname: string): null | string {
   assertRouteString(urlSource, `${configSrc} Invalid`)
   assertUsage(
-    urlTarget.startsWith('/') ||
-      urlTarget.startsWith('http://') ||
-      urlTarget.startsWith('https://') ||
-      urlTarget.startsWith('mailto:') ||
-      urlTarget === '*',
+    urlTarget.startsWith('/') || /^[a-zA-Z]+:/.test(urlTarget) || urlTarget === '*',
     `${configSrc} Invalid redirection target URL ${pc.cyan(urlTarget)}: the target URL should start with ${pc.cyan(
       '/'
-    )}, ${pc.cyan('http://')}, ${pc.cyan('https://')}, ${pc.cyan('mailto:')}, or be ${pc.cyan('*')}`
+    )}, a valid protocol, or be ${pc.cyan('*')}`
   )
   assertParams(urlSource, urlTarget)
   const match = resolveRouteString(urlSource, urlPathname)
@@ -46,7 +42,7 @@ function resolveRouteStringRedirect(urlSource: string, urlTarget: string, urlPat
     assertUsage(!urlResolved.includes('@'), 'URL should not contain "@" unless it is a mailto link.')
   }
   if (urlResolved === urlPathname) return null
-  assert(urlTarget.startsWith('/') || urlTarget.startsWith('http') || urlResolved.startsWith('mailto:'))
+  assert(urlTarget.startsWith('/') || /^[a-zA-Z]+:/.test(urlTarget))
   return urlResolved
 }
 
--- a/vike/shared/route/resolveRedirects.ts
+++ b/vike/shared/route/resolveRedirects.ts
@@ -4,6 +4,7 @@ export { resolveRedirects }
 export { resolveRouteStringRedirect }
 
 import { assertIsNotBrowser } from '../../utils/assertIsNotBrowser.js'
+import { isUriWithProtocol } from '../../utils/parseUrl-extras.js'
 import { assert, assertUsage } from '../utils.js'
 import { assertRouteString, resolveRouteString } from './resolveRouteString.js'
 import pc from '@brillout/picocolors'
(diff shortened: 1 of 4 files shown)


vike/shared/route/resolveRedirects.ts
```diff
@@ -41,9 +42,11 @@ function resolveRouteStringRedirect(urlSource: string, urlTarget: string, urlPat
     }
     urlResolved = urlResolved.replaceAll(key, val)
   })
-  assert(!urlResolved.includes('@'))
+  if (!urlResolved.startsWith('mailto:')) {
+    assert(!urlResolved.includes('@'), 'URL should not contain "@" unless it is a mailto link.')
+  }
   if (urlResolved === urlPathname) return null
-  assert(urlTarget.startsWith('/') || urlTarget.startsWith('http'))
+  assert(urlTarget.startsWith('/') || urlTarget.startsWith('http') || urlResolved.startsWith('mailto:'))
```
brillout, line 49:
```suggestion
  assertUsage(urlTarget.startsWith('/') || urlTarget.startsWith('http') || urlResolved.startsWith('mailto:'))
```

vike/shared/route/resolveRedirects.ts
```diff
@@ -41,9 +42,11 @@ function resolveRouteStringRedirect(urlSource: string, urlTarget: string, urlPat
     }
     urlResolved = urlResolved.replaceAll(key, val)
   })
-  assert(!urlResolved.includes('@'))
+  if (!urlResolved.startsWith('mailto:')) {
+    assert(!urlResolved.includes('@'), 'URL should not contain "@" unless it is a mailto link.')
```
brillout, line 46:
```suggestion
    assertUsage(!urlResolved.includes('@'), 'URL should not contain "@" unless it is a mailto link.')
```

vike/shared/route/resolveRedirects.ts
```diff
@@ -41,9 +42,11 @@ function resolveRouteStringRedirect(urlSource: string, urlTarget: string, urlPat
...
+  }
   if (urlResolved === urlPathname) return null
-  assert(urlTarget.startsWith('/') || urlTarget.startsWith('http'))
+  assertUsage(urlTarget.startsWith('/') || urlTarget.startsWith('http') || urlResolved.startsWith('mailto:'))
```
brillout, line 49:
```suggestion
  assert(urlTarget.startsWith('/') || urlTarget.startsWith('http') || urlResolved.startsWith('mailto:'))
```
I suggested the wrong assertion.

usk94: 🙆‍♂️

usk94:
@brillout 
> I wonder if this could be extended to support any other protocol such as ipfs:, ipns:, magnet:, etc. If all protocols match /^[a-zA-Z]:/, then I guess we could simply replace the http:// checks with that regex.

I wonder `javascript:` and `data:` are at risk of xss vulnerability, so how about validating them?https://cheatsheetseries.owasp.org/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.html

brillout:
> I wonder `javascript:` and `data:` are at risk of xss vulnerability, so how about validating them?https://cheatsheetseries.owasp.org/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.html

What kind of vulnerability do you have in mind? Do you see a problem with setting the `Location` header to any arbitrary string? I could be wrong, but I'd assume browsers to already validate and not do anything funky upon arbitrary `Location` values.

As for https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html#dangerous-url-redirect-example-1 it don't think it applies to us: it's something the Vike user needs to careful about and I don't see how Vike could implement additional protections.

usk94:
@brillout 
Understood. 
I was concerned more about the possibility of malicious developers exploiting end users rather than xss vulnerabilities. 
But you're right, that's something developers need to be cautious about, and there's nothing we can do about it on our end👍
Thank you for pointing that out!

brillout:
👍 That said, I could very well miss out a problematic situation, but so far I can't think of one. I guess it's ok if we risk it.

usk94:
I have switched to using regex🙆‍♂️

brillout:
I did some polishing, everything's perfect IMO :) Do we agree and shall we merge?

usk94:
It's fine. Thank you for your reviewing and refactoring!

vike/utils/parseUrl-extras.ts
```diff
@@ -109,3 +110,8 @@ function addUrlOrigin(url: string, origin: string | null): string {
   const urlModified = createUrlFromComponents(origin, pathnameOriginal, searchOriginal, hashOriginal)
   return urlModified
 }
+
+function isUriWithProtocol(uri: string): boolean {
+  // Is it correct to assume that all protocols (http:, https:, ipfs:, magnet:, ...) match this RegExp?
```
ndavd, line 115: Albeit not common, URL schemes may contain digits, `+`, `.` and `-`.

ndavd:
Added a small comment @brillout

ndavd:
> Do you have an example?

https://en.wikipedia.org/wiki/List_of_URI_schemes

brillout:
Neat. Fixed: 72ede8a12e5482405590656f79a2219374d0213c.

ndavd:
Looks great 🔥

PR #1296, opened 2023-11-29
Create plugin for asserting file runtime

### LINKED ISSUE #744: Use `.server.js` and `.client.js` to assert the JavaScript environment
### Description

Now that the V1 design removes the filename suffixes such as `.page.server.js` and `.page.client.js`, we can use `.server.js` and `.client.js` for another purpose.

I suggest we use these to assert where the file is loaded.

```js
// /some/where/SomeClientOnlyComponent.client.js

export function SomeClientOnlyComponent() {
   // ...
}
```

```js
// /some/where/secrets.server.js

export const databaseConnection = {
  password: '4H34CXA3cMa'
}
```

Enabling the user to ensure that the password won't ever be loaded in the browser.

usk94:
resolves #744

I referred to https://gist.github.com/dajinchu/2c03ec791ad135e481dbb96bd3538121 for this implementation. 

I encountered a different error from the one I implemented when using `es-module-lexer` while starting the server (I don't understand the reason for this error). so I used regex to extract import statements from the source code. 

Also, in the error messages, I made it clear which file was executed at a different runtime despite client/server only, and from which file it was imported.


usk94:
> I encountered an error when using `es-module-lexer` while starting the server 

Oops, when I used the same source code of the referenced page, there was no error🤔
Was I not using it right?

usk94:
@brillout 
May I ask for a review?

> I encountered an error when using `es-module-lexer` while starting the server

Also, it is possible that I did not use `es-module-lexer` correctly for this ( https://github.com/vikejs/vike/pull/1296#issuecomment-1832339000 ), but I have already made an implementation that does not use this library.
I am fine with either implementation and would like to hear your opinion😀

brillout:
Hello Yusaku, thanks for the PR 💚

Couple of thoughts:
 - This `transform()` applies to every module, so I wonder how much this would slow down DX speed?
 - I wonder how reliable the regex is. One common use case where it will fail:
   ```js
   /* We comment out problematic import, but Vike still throws the error
   import something from './file.server.js'
   */
   ```

brillout:
Btw. if we feel lazy we can go for the dumber implementation: simply using `resolveId()` instead of `transform()`. I actually wonder, maybe there is a Rollup API that provides both `id` and `importer`, which would give us the information we need for providing the neat DX.

> Certainly, the current regular expression does pick up imports within comments, which is a problem. There might be other loopholes as well, so I think it's best to stick with `es-module-lexer`.

I'm also inclined to think `es-module-lexer` would be better.

> Regarding the benchmarking method, would it be appropriate to run build and dev several times in one of the directories under `examples`, and then calculate the average for each?

AFAICT I think so.

brillout:
Actually, I'm just seing that [`resolveId()`](https://rollupjs.org/plugin-development/#resolveid) does provide both the `id` (we can get it by calling `const { id } = await this.resolve(source, importer, options)`) and the `importer`. So AFAICT we can simply use `resolveId()` instead of `transform()` and skip the whole `es-module-lexer` / RegExp parsing.

> Wrote another version using resolveId instead of transform but transform gives a nice error message showing where the import happened. [quote from [here](https://github.com/vikejs/vike/issues/744#issuecomment-1686322750)]

@dajinchu AFAICT `resolveId()` would work as well since it provides both the `id` and `importer` &mdash; let us know if you foresaw a problematic edge case for `resolveId()`.

brillout:
>  find the line of code that contained the import

I see. I guess it's fine if we only say which file is the offender without line number.

usk94:
> Also thanks to source we know the import path, I guess file + source is enough.

Ok, I am going to use `resolveId`.

usk94:
And I've encountered the `es-module-lexer` error again, and I might have figured out the cause. 
This library does not support parsing JSX. https://github.com/guybedford/es-module-lexer/issues/47

So, if we want to meet the requirements of this issue, it seems better to either use regular expressions or look for another library.

usk94:
P.S. 
With resolveId, there is no need to parse the source code because it provides an import statement by default.
So we may no longer have to use either `es-module-lexer` or regular expressions.

brillout:
> does not support parsing JSX

Yes, the transform needs to run after the JSX (and any other) transform.

> P.S.
> With resolveId, there is no need to parse the source code because it provides an import statement by default.
> So we may no longer have to use either `es-module-lexer` or regular expressions.

Neat, that's what I was hoping for 👌

usk94:
And the build benchmark results are as follows. 
As a method of measurement, I cleared the cache (deleted node_modules/.vite) before each `build` and `dev`, but somehow the performance was faster after the addition. 

I feel there might be something wrong with the way I'm measuring. By the way, even without clearing the cache, the average was faster after the addition.

### build
- before: 4.81s (1st: 4.90s, 2nd: 5.06s, 3rd: 4.48s)
- after: 4.40s (1st: 4.27s, 2nd: 4.34s, 3rd: 4.59s)

usk94:
> Rergading performance, I think it's fine and we don't need to test performance if we use resolveId().

🙆‍♂️

> Why enfore: "pre"? Maybe I'm missing something but wouldn't enforce: "post" be better?

When using `enforce: "post"`, it did not throw an error at build time, even though there was a file importing `foo.server.tsx`, and did not even output `console.log` put in the line 1 of the resolveId.

This may be similar to the following:
https://github.com/vitejs/vite/issues/1899#issuecomment-775259950

usk94:
> I think we need to use this.resolve(), otherwise we will miss aliases.

I fixed it🙆‍♂️

> How about we create a little test at /test/misc? See expectLog() which enables to assert errors (see usage in examples/ and test/).

Wait a moment.

phonzammi:
@usk94 Have you try with 'pnpm run test:e2e'

brillout:
@usk94 Make sure to follow the usual structure of tests, e.g. always wrap tests with `test()` async handlers. (Make sure they are async.)

usk94:
@brillout 
I thought that since this is a test for a plugin, it would be more of a unit test rather than an e2e test, so I implemented it using vitest. 
I'm not very confident about it, so please feel free to comment if I missed something😅

brillout:
I agree creating a new e2e test would be overkill, but I was thinking more of adding a new page `pages/forbidden-import/+Page.js` that would do some forbidden import at `test/misc/`. And then add a new test to `test/misc/.testRun.ts` that uses `expectLog()`.

usk94:
I am currently adding a test like this in test/misc/.testRun.ts, but it outputs Error: `The following log was expected but it wasn't logged.`

```ts
  test('assertFileRuntime plugin - invalid import', async () => {
    await editFile('./pages/forbidden-import/+Page.tsx', (content) =>
      content.replace(/^/, `import ServerOnly from './ServerOnly.server';\n`)
    )
    await page.goto(getServerUrl() + '/forbidden-import')
    expectLog('Failed to load resource: the server responded with a status of 404 (Not Found)')
  })
```

I have created `pages/forbidden-import/+Page.tsx`, but if I directly write imports including `.server`(or `.client`) there, it actually fails during the build within the test. 
So, I am trying to add the import statement from e2e using editFile, but it's not working successfully.

usk94:
And this is now `test/misc/vite.config.ts`.
```ts
import react from '@vitejs/plugin-react'
import vike from 'vike/plugin'
import { assertFileRuntime } from '../../vike/node/plugin/plugins/runtimeValidation/fileRuntime'

export default {
  build: {
    outDir: `${__dirname}/../../test/misc/dist/nested`
  },
  plugins: [
    react(),
    vike({
      prerender: {
        noExtraDir: true
      }
    }),
    assertFileRuntime()
  ]
}

usk94:
If I put this plugin in the `vite.config.ts`, the test does not pass by simply doing an `expectLog`, as the plugin is triggered and throws a `Client-only module...` error before the page is displayed in e2e (it is quite possible that I am not doing it right).

I still think that unit test is more suitable than e2e, but what do you think?

And I would like to know if there are any plugins that test with e2e as well, as it would be helpful.

brillout:
> I am currently adding a test like this in test/misc/.testRun.ts, but it outputs Error: `The following log was expected but it wasn't logged.`

Can you push the commits that lead to that error? I'll have a look at it then. (In general, feel free to push commits that aren't finished to your PRs.)

I still think adding a test to `test/misc/` would be better. Let's see.

brillout:
Add `(log) => log.logSource === 'stderr'` as second argument to `expectLog()`.

usk94:
@brillout 
Thank you so much. 

After implementing this approach, I managed to eliminate the `Error: The following log was logged as expected, but it didn't match the logFilter() you provided` error. 

However, the test continued to fail with a 500 error status. Consequently, I added the prefix tags, the timestamp, and '\n' in an effort to achieve an exact match in the log entries. Despite these adjustments, the test is still not passing...

usk94:
I feel like I'm gradually getting closer to the right solution😐
This test may pass if I can properly clean up the invalid import.

brillout:
See `vike/node/plugin/index.ts` which is the place where Vike's internal plugins are registered.

usk94:
Let's organize the situation.

As you can see, registering the plugin in `vike/node/plugin/index.ts` causes various tests to fail.
Therefore, I narrowed down the registration to `test/misc/vite.config.ts`, but even then, the tests still fail.
The way they fail, as shown in the screenshots, is that although the test case I added(`assertFileRuntime plugin - invalid import`) pass, the test itself fails afterwards.

I suspect this is because the plugin's scope is not only extending to the virtual environment used for e2e testing but also to the environment for running the tests.
That's why I thought it might be good to insert an invalid import statement containing `.client` only during the test cases and revert this edit at the end. 

However, even with this approach, the test itself fails after all the test cases have passed.

image

usk94:
@brillout 
If there are any errors in my assumptions, please do not hesitate to correct me. I admit that I have limited knowledge about e2e testing and Vite plugins, and there is a possibility that I might be off the mark in some of my statements, but I am committed to completing this PR.

brillout:
@usk94 👍 If you `expectLog('HTTP response /forbidden-import 500', (log) => log.logSource === 'stderr')` then `@brillout/test-e2e` won't complain about it.

usk94:
@brillout 
I finally realized something basic. 
I had to catch all error statements in `expectLog`.

Thank you for your advice.
How about this fix?

And regarding the remaining test failing due to a type error, it seems to be related to content within `test/misc`, but at first glance, it appears unrelated to my PR. Should I fix this as well?

brillout:
@usk94 I refactored the PR a bit, I hope it's ok. Let me know if you disagree with some of it. If not then I'll merge 🎉

Thanks for the PR, it was a very useful preparation 👌 (It would have taken me a lot of time to find out that `enforce: 'pre'` and  `skipSelf` are needed.)

nitedani:
skipSelf defaults to true after Vite 5 https://vitejs.dev/guide/migration#rollup-4

brillout:
@nitedani Thanks, comment updated.

usk94:
@brillout 

Thank you very much for doing a lot of refactoring (or rather, almost finishing it)! 

I struggled a lot with how to test for this plugin, but I'm glad I was able to learn about vite's plugins and e2e testing. 

I have no objections to the changes! Please merge it🎉

brillout:
Merged 🎉

Apologies for having marked the ticket as https://github.com/vikejs/vike/labels/easy-contribution%20%F0%9F%8C%B1 as it turned out to be quite far from easy.

Let me know if you want to tackle another ticket and I'll dig a little bit more into whether I foresee the ticket to be a relatively easy contribution.

Thanks for your PR and looking forward to more 💯

Issue #1299, opened 2023-11-30
Some preload tags are missing in `injectFilter`


brillout:
### Description

https://vike.dev/injectFilter should provide every possible preload tags, so that the user can easily toggle and position preload tags.

brillout:
Actually, I just checked and it should already be the case: all JavaScript chunks listed in `dist/asset.json` (and that are used by the page currently being rendered) are passed to `injectFilter()`.

@jakubfiala Reproduction welcome. Closing but I'll re-open if I can get my hands on a reproduction.

Note that the JavaScript *entry*, e.g. `entry_server_routing.js`, is purposely not passed to `injectFilter` for various reasons, but I think it's possible to pass the entry as well. I'll implement this today/tomorrow.

brillout:
> this is not currently the case for chunks which are generated from dynamic imports.

It could be nice to add them to `injectFilter` while setting them to `inject: false` by default but, yeah, it could be a somewhat risky DX. I'll think and have a look at it this next week.

> Note that the JavaScript _entry_, e.g. `entry_server_routing.js`, is purposely not passed to `injectFilter` for various reasons, but I think it's possible to pass the entry as well. I'll implement this today/tomorrow.

(Postponing this to next week.)

PR #1308, opened 2023-12-03
Add preload tags to `injectFilter`

Kenzo-Wada:
Issue: https://github.com/vikejs/vike/issues/1299

I've added preload tags such as `video`, `audio` and more (all from [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-as)

If there are something wrong or missing, i'd be happy if you let me know...!!

### DIFF
--- a/docs/pages/preload.page.server.mdx
+++ b/docs/pages/preload.page.server.mdx
@@ -28,7 +28,7 @@ type PageContext = {
   httpResponse: {
     earlyHints: {
       earlyHintLink: string, // Header Line for the Early Hint Header
-      assetType: "image" | "script" | "font" | "style" | null
+      assetType: "image" | "script" | "font" | "style" | "audio" | "video" | "document" | "fetch" | "track" | "worker" | "embed" | "object" | null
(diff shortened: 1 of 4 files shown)


brillout:
Wonderful. Also good choice of prioritization.

I actually didn't really elaborate what the issue https://github.com/vikejs/vike/issues/1299 is about, apologies for that. That's the issue: https://github.com/vikejs/vike/issues/1299#issuecomment-1837587717. Is that something you'd be up to have a look at?

We can merge the PR as-is or we can contine working using this PR, as you wish.

Also, a small thing: the commit message should start with `feat: ` otherwise it won't make it into `CHANGELOG.md`. (You can put the emoji after `feat: ` or at the end 🙂)

brillout:
I think we should postpone fixing #1299 until we get a reproduction.

Let's merge 🎉

Thanks for the PR 💚

Looking forward to more PRs - if you want 🙂

Kenzo-Wada:
allright! thanx!

PR #1314, opened 2023-12-05
Remove implication of no bugs in Vike

zcolleyz:
I think you should be kind to yourself here and not invite people to be annoyed if there are bugs in the future ❤️

### DIFF
--- a/docs/pages/error-page.page.server.mdx
+++ b/docs/pages/error-page.page.server.mdx
@@ -62,7 +62,7 @@ The error page is rendered if:
  2. One of your hooks threw an error (`500 Internal Error`).
  3. You called <Link href="/render" text={<code>throw render(abortStatusCode)</code>} />.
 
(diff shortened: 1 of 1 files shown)


brillout:
Thanks for caring 🙂

We actually plan to double down on this and further emphasis the fact that Vike has "no known bugs" in the upcoming new marketing pages.

But you're right: we should explain what exactly we mean with "no known bug", which we'll do. I'm declining your PR in the meaning, I hope it's ok.

Thanks for the PR and, again, thanks for caring :)

zcolleyz:
If it's for marketing I think it'll have the adverse affect, people will trust you less because all code has bugs.

brillout:
Makes sense and you make me realize that we really should make it clearer what we mean with "no known bug" (or maybe using another wording 🤔). Essentially, what we mean is that we prioritize reliability over fanciness (in contrast to many frameworks which are very marketing/fanciness driven).

PR #1321, opened 2023-12-05
qual: add warning when streaming and pageContext promise

AurelienLourot:
follow-up on #1271


brillout:
👌

brillout:
I'm thinking, while the warning is on point, many(/most?) users may not understand its meaning; it could be nice to add a section on https://vike.dev/stream e.g. "Partial hydration" and, in the warning, add a link to it.

AurelienLourot:
Makes sense, I'll work on that

PR #1343, opened 2023-12-09
Docs: update assetsType of injectfilter

Kenzo-Wada:
related to: https://github.com/vikejs/vike/pull/1308

fix: I have forgot to update assetType of injectfilter of the document


brillout:
👍 Maybe we should define it once and re-use it everywhere?

Kenzo-Wada:
That is true in the actual implementation, but I thought that since this PR only changes Docs, there is no need to define constants. what do you think?

brillout:
Ok makes sense. I didn’t see it was about the docs.

Merged 👌 Feel free to keep PRs coming if you want. Looking forward to them 👀

Issue #621, opened 2023-02-01
Workaround wrongly published packages


brillout:
### Description

Examples:
 - https://github.com/vitejs/vite/issues/11299
 - https://github.com/aws-amplify/amplify-ui/issues/3155
 - https://github.com/cyco130/vite-plugin-cjs-interop

brillout:
- https://github.com/brillout/vite-plugin-ssr/discussions/635

brillout:
https://github.com/vitejs/vite/issues/11299

brillout:
https://github.com/brillout/vite-plugin-ssr/discussions/682

brillout:
https://github.com/brillout/vite-plugin-ssr/discussions/700#discussioncomment-5283034

brillout:
- Move all user land to ESM in order to avoid:
    ```
    Error [ERR_REQUIRE_ESM]: require() of ES Module
   /home/rom/code/vite-plugin-ssr/node_modules/.pnpm/node-fetch@3.3.1/node_modules/node-fetch/src/index.js
   from
   /home/rom/code/vite-plugin-ssr/examples/graphql-apollo-react/server/index.js
   not supported.
   ```

brillout:
CSS imports: https://github.com/brillout/vite-plugin-ssr/discussions/791.

brillout:
> CSS imports: #791.

Maybe, instead of automatically setting `ssr.noExternal`: and try-catch at `ssrLoadModule()` and tell user to add library to `ssr.noExternal`.

brillout:
> > CSS imports: #791.
> 
> Maybe, instead of automatically setting `ssr.noExternal`: add try-catch at `ssrLoadModule()` and tell user to add library to `ssr.noExternal`.

Contribution welcome.

brillout:
https://github.com/brillout/vite-plugin-ssr/discussions/830

brillout:
https://github.com/brillout/vite-plugin-ssr/discussions/901

brillout:
https://github.com/brillout/vite-plugin-ssr/discussions/934

brillout:
Also `ReferenceError: window is not defined`

brillout:
https://github.com/brillout/vite-plugin-ssr/discussions/571#discussioncomment-6137618 (including repro: https://github.com/antun/vite-ssr-redux-react-live-runner-example).

brillout:
> add try-catch 

[...]

PR #1358, opened 2023-12-15
show import errors in dev

nitedani:
Related: https://github.com/vikejs/vike/issues/621

Show helper message in dev.


nitedani:
I need some help with showing the error message, snippet, and the original error below them.
https://github.com/vikejs/vike/pull/1358/files#diff-52c661a478f000a8c948b903364849179409ab4986372eadd14f8f0c76012631R20-R28

brillout:
Neat 💯

How about we wrap [`viteDevServer.ssrLoadModule(virtualFileIdImportUserCodeServer)`](https://github.com/vikejs/vike/blob/9a3ac5afc806701af3aee467674c74cd5e0acb89/vike/node/runtime/page-files/setup.ts#L13) instead? I guess that would catch the same exceptions?

As for production, we can catch exceptions thrown by [`await loadServerBuild(outDir)`](https://github.com/vikejs/vike/blob/9a3ac5afc806701af3aee467674c74cd5e0acb89/vike/node/runtime/globalContext/loadImportBuild.ts#L23).

We can also catch exceptions thrown by [`await hookFnCaller()`](https://github.com/vikejs/vike/blob/9a3ac5afc806701af3aee467674c74cd5e0acb89/vike/shared/hooks/executeHook.ts#L69) such as React's infamous `Warning: React.jsx: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: undefined.`. (I wrote a React patch a while ago to make debugging this React error much easier; I'll see if I can find it again and polish it.)

nitedani:
[viteDevServer.ssrLoadModule](https://github.com/vikejs/vike/blob/9a3ac5afc806701af3aee467674c74cd5e0acb89/vike/node/runtime/page-files/setup.ts#L13)
doesn't throw any errors

I've found no other way to catch these errors.

brillout:
In principle, the error must/should be catched somewhere by Vike, otherwise
Vike wouldn’t be able to show the error page.

In other words: if the error leads to the error page to be rendered then we
have, de facto, the opportunity to catch the error in Vike Land. (Although
make sure the error page exists, otherwise Vike has no error page to
render.)

brillout:
Notable changes:
 - https://github.com/vikejs/vike/pull/1358/commits/df9423bebcbbf625dd25b7ce3559f187212e3878
 - https://github.com/vikejs/vike/pull/1358/commits/52a11ca2ea3aa16e267b9f8ab6d6c4a02fb956cb

The rest is mostly just some minor refactoring.

Let me know if you object with any of the changes, otherwise I'll merge 🎉

nitedani:
onRuntimeError is not called, it stops here: https://github.com/vikejs/vike/pull/1358/files#diff-7fb721f8b206cd2db4fc64bbf0c4e005aac41836d50d66133acc17b019255d1cR114

brillout:
I actually realized I forgot to try my changes 😅

> onRuntimeError is not called, it stops here: https://github.com/vikejs/vike/pull/1358/files#diff-7fb721f8b206cd2db4fc64bbf0c4e005aac41836d50d66133acc17b019255d1cR114

Let me have a look.

brillout:
I just had a quick look. Vike treats the error as transpilation error and thus doesn't trigger `onRuntimeError()`. Let me fix this.

brillout:
@nitedani Alright, I think it's good now. Shall we merge?

(I'll double check if we missed something from the list later today. Other than that, the PR LGTM now.)

Issue #1339, opened 2023-12-09
Remove `?extractAssets` workaround in favor of `build.ssrEmitAssets`


brillout:
### Description

The [code for the `?extractAssets` workaround](https://github.com/vikejs/vike/blob/main/vike/node/plugin/plugins/extractAssetsPlugin.ts) is non-trivial and brittle. (See https://vike.dev/includeAssetsImportedByServer for why the workaround is needed.)

It would be more robust to use [`build.ssrEmitAssets`](https://vitejs.dev/config/build-options.html#build-ssremitassets) instead. (There is actually [an issues with a Babel plugin](https://gist.github.com/brillout/e03c8485da06bbcd338c623a23f9d347).)

See also:
 - https://github.com/vitejs/vite/pull/11430
 - https://github.com/vitejs/vite/discussions/13808

nitedani:
I've tried but I'm stuck:
https://github.com/vikejs/vike/blob/nitedani/extract-assets/vike/node/plugin/plugins/ssrEmitAssets.ts

I don't know why, but the ssr manifest doesn't contain any of the css imports.
I tested tried it with the [path-aliases-v1](https://github.com/vikejs/vike/tree/nitedani/extract-assets/examples/path-aliases-v1) eample.

It works just fine here: https://stackblitz.com/edit/vitejs-vite-svc9fw?file=dist%2Fssr-manifest.json&terminal=dev
Here, the css is included in the ssr manifest.

brillout:
Implemented by https://github.com/vikejs/vike/pull/1364.

PR #1364, opened 2023-12-17
Remove ?extractAssets workaround in favor of build.ssrEmitAssets

nitedani:
Implements https://github.com/vikejs/vike/issues/1339.


nitedani:
Is this the right direction?

brillout:
I think it's easier if we inject the server manifest into the client manifest [here](https://github.com/vikejs/vike/blob/86d6558eb0ea0716b0f349af4cf6c35f7c3e2acd/vike/node/plugin/plugins/buildConfig.ts#L69-L78). That way, the rest of Vike's code will work as it does today, e.g. no need to modify `node/runtime/globalContext.ts` then. Atually, if we keep `dist/assets.json` semantically equivalent, then I think we don't need to modify the server runtime at all.

brillout:
I'm reading `ssrEmitAssets.ts` right now, and I'm seeing logic to merge manifests. So I guess you already tried to merge the server manifest into `dist/assets.json` and have things "just work".

Still, I don't think we need to adapt the server runtime (nor the pre-rendering code) if we keep `dist/assets.json` semantically equiavelent. (More precisely speaking, all we need to do is to remove the `?extractAssets` handling from the server runtime, which you seem to have already done.)

brillout:
AFAICT, if we do a clever merging instead of just a "dumb"  `{ ...clientManifestParsed, ...serverManifestParsed }`, then we can keep `assets.json` semantically equivalent and, therefore, alleviate ourselfes the need to adapt the server runtime.

So I guess the question is this: can we make the merging so clever that evertyhing else just works as before? A lot of Vite frameworks use `ssrEmitAssets`; a GitHub search for "ssrEmitAssets" would point to a lot of code to copy from.

brillout:
LGTM, let me know if for you as well then I'll merge.

Looking forward to further resolve all blockers and merge the worker server PR ✨

nitedani:
Looks good, works for me.

PR #1365, opened 2023-12-17
docs: add /data

AurelienLourot:
Follow-up on #1267. I intend to add the `/useData` documentation in another follow-up PR.


brillout:
Quite neat that `data()` is now the default showcased hook :ok_hand:

- How about we completely move the section `/data-fetching > ## onBeforeRender()` to `/onBeforeRender`? Instead, we could add a note with a link to `/onBeforeRender` at the end of `/data-fetching > ## data()`, and create a section `/data-fetching > ## See also` and add a link to `/onBeforeRender` in there as well.
- While `## Custom settings` is a neat trick, I'm tempted to move and document it only in `/onBeforeRender`. The rationale being to make `/data-fetching` trivial for juniors, while `/onBeforeRender` is for expert users. In general, if possible, it would be neat if juniors aren't confrontend too much with `meta` in guides.
 - How about we add a note at the beginning of `/onBeforeRender` saying "The `onBeforeRender()` hook is for expert Vike users. If you're new to Vike, we recommend using the `data()` hook, and/or using Vike extensions for deep integrations with GraphQL, RPC, Tanstack Query, etc.".
- Also to make guides increasingly junior friendly, I think we can move `/data-fetching > ## Client Routing` to `/data`. (It also neatly fits the idea that the guide `Data Fetching` is a high-level tutorial, while `API > data()` is exhaustive.)
- I think we can remove the following comment from `/data-fetching`. I guess users don't need to know about `passToClient` when they use `data()`.
  ```js
  // Because pageContext.data is part of `passToClient` by default, it is
  // available here in the browser.
  ```
- After we implement the `useData()` hook, we can stop showcasing `onRenderHtml()` and `onRenderClient()` in `/data-fetching`. Further making things trivial for juniors :ok_hand:

Let me know if something's unclear.

AurelienLourot:
Done :)

brillout:
(Didn't finish proofreading my changes, I will do it tentativelly tomorrow - if my flight preparation is going to be as quick as expected.)

brillout:
(Started using `+data.js` on `/examples/react-full`, it's quite a lot of fun.)

brillout:
This now LGTM. WDYT?

AurelienLourot:
LGTM :)

brillout:
Very neat. Can't wait for `useData()` 🙂 👀 (but no rush).

PR #1372, opened 2023-12-22
feat: cacheControl setting

### LINKED ISSUE #1275: Unexpected pageContext behaviour when navigating to page via the URL bar and back

AurelienLourot:
Spin-off of #1282. Fixes #1275

### DIFF
--- a/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/configDefinitionsBuiltIn.ts
+++ b/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/configDefinitionsBuiltIn.ts
@@ -150,6 +150,9 @@ const configDefinitionsBuiltIn: ConfigDefinitionsBuiltIn = {
   },
   hooksTimeout: {
     env: { server: true, client: true }
+  },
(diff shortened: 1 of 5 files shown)


AurelienLourot:
I'd do https://vike.dev/cacheControl in a separate PR

brillout:
Very neat 💯

I did some polishing, let me know if you object with anything otherwise I'll merge 🎉

AurelienLourot:
LGTM :100:

PR #1377, opened 2023-12-23
docs: improve /react

brillout:
@AurelienLourot @luisfloat

The resources about island architecture are neat, it would be nice if they are listed somewhere in https://github.com/luisfloat/vike-island-example.


AurelienLourot:
One of the main points of the islands example (hopefully soon to become a Vike extension) is to mix React and Vue in the same page. Why placing it now under https://vike.dev/react ? Why not https://vike.dev/vue or any other UI framework?

I wouldn't mind just making https://vike.dev/islands detached, if this is what you are after. What do you think?

brillout:
> hopefully soon to become a Vike extension

That'd be nice indeed, we could list it under `## Vike Extensions` then.

> One of the main points of the islands example [...] is to mix React and Vue in the same page.

I've udpated the tagline, but let me know if you think it doesn't do it justice (also feel free to push a changes to my PR and/or use GitHub's change suggestion feature).

> Why not https://vike.dev/vue 

Done.

> I wouldn't mind just making https://vike.dev/islands detached, if this is what you are after. What do you think?

I'm not against keeping `/islands`, but I'm not sure where we would link it from.


I think, with my PR, the example is actually going to get more attention. Even more so in the future if it becomes a Vike extension.

AurelienLourot (approved):
LGTM :pray:

brillout:
@luisfloat I'm seing you upvoted my last comment, so I'll take is a PR approval. Merging, but feel free to disagree with anything.

Looking forward to a `vike-react-islands`/`vike-vue-islands`/`vike-islands` 🚀

PR #1384, opened 2023-12-25
add vike-angular types

nitedani:
Can't wait for the XOR6 type 🥹


brillout:
> Can't wait for the XOR6 type 🥹

😅

PR #1386, opened 2023-12-26
feat: generic `<Data>` type for `onBeforePrerenderStart()`

AurelienLourot:
Needed by https://github.com/vikejs/vike-vue/pull/52

@brillout it works :) I have updated the `react-full` example accordingly, which makes the `react-data-fetching` example obsolete, by the way. If you like this DX I will:

1. adapt `vue-full` in the same way
2. delete `react-data-fetching`
3. document this at https://vike.dev/onBeforePrerenderStart

### DIFF
--- a/examples/react-full/pages/hello/+Page.tsx
+++ b/examples/react-full/pages/hello/+Page.tsx
@@ -1,8 +1,11 @@
 export default Page
 
 import React from 'react'
+import { useData } from '../../renderer/useData'
+import type { Data } from './+data'
 
-function Page({ name }: { name: string }) {
+function Page() {
+  const { name } = useData<Data>()
   return (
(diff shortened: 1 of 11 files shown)


brillout:
💯 LGTM. Feel free to merge.

brillout:
If you want I can add you to https://npmjs.com/package/vike so you can push new versions.

AurelienLourot:
> If you want I can add you to https://npmjs.com/package/vike so you can push new versions.

Yes, why not. I won't release any new versions without asking you first, at least for now, because I feel like I don't always know if it's safe to release. But I don't mind doing this one. Thanks!

brillout:
👍 Done.

PR #1387, opened 2023-12-26
examples: use useData() in `vue-full`

AurelienLourot:
Similar to
* https://github.com/vikejs/vike/pull/1386
* https://github.com/vikejs/vike-vue/pull/52


examples/vue-full/pages/hello/+Page.vue
```diff
@@ -19,5 +19,7 @@
 </template>
 
 <script lang="ts" setup>
-defineProps(['name'])
+import type { Data } from './+data'
+import { useData } from '../../renderer/useData'
+const { name } = useData<Data>()
```
AurelienLourot, line 24: @brillout this is not reactive when I client-navigate from `/hello/eli` to `/hello/jon`. I believe this has to do with `pageContextReactive` in `app.ts`. But somehow I can't spot what's missing. I can dig more but I thought maybe you can already spot it? Thanks

Everything else works.

brillout: I also don't see why it doesn't work.

I would have assumed it works since the next `pageContext.data` is a completely different object than the previous one.

Also, Vue's `reactive()` is recursive. So I'd assume this part of the story to also work. I'd `console.log(name)` to see if it's indeed wrapped and I'd also try using `data.name` / `pageContext.data.name` in the template to see if it works then.

AurelienLourot: Fixed with https://github.com/vikejs/vike/pull/1387/commits/6e34e3fec3 - Let me know if you don't like this DX but I think it's nice

brillout: I'm not familiar with Vue, but I'm thinking maybe Vue users are accustomed to use `someReactiveObject.name` instead of directly using `name`. I guess we can go for that solution then 👍

Btw. about the `vike-{react,vue,solid}` docs, I was thinking we could add a readme for each of them that simply lists features with, for each feature, a link to the docs e.g. https://vike.dev/ClientOnly.

PR #1395, opened 2023-12-30
internal refactor

brillout:
@phonzammi In case you're curious, this is how I did it: a *lot* of micro refactoring commits culminating to a substantial refactor. The live stream would have taken a little too long I guess :)


phonzammi:
Yeah, I saw it every time you pushed a commit,

> The live stream would have taken a little too long I guess

Absolutely 🤦‍♂️

Thanks, very awesome 💯

PR #1409, opened 2024-01-03
Fix criptic → cryptic typo in faq.page.server.mdx

redbar0n:


brillout:
Thanks Magne

brillout:
And thank you for spreading the word about Vike. I think you'll like a lot what Vike will become in the coming months and years.

PR #1424, opened 2024-01-10
add docs for vike-vue/ClientOnly

4350pChris:


brillout:
I did a little bit of polishing (semi-related to your PR). As always, feel free to object with any my changes.

I'm not that familiar with Vue so I don't know whether it's obvious, but maybe we should document that all props are passed through?

4350pChris:
Good point. It's not that obvious actually.
At some point I wanted to take a deeper look at that component again to improve DX because right now tsc can't infer the prop types on `ClientOnly` which should become the ones from the component `load` resolves to... But that's for the future. For now it works even though you don't get any hints by your IDE.

PR #1428, opened 2024-01-11
Update routing.page.server.mdx

Perlover:
I'm not sure, but I think the example implied a different path in route


brillout:
I think the docs are correct here.

PR #1432, opened 2024-01-12
add docs for properties in `config.meta`

### LINKED ISSUE #1300: documentation: `env` and `effect`
### Description

#1206 covers `env: {client: ..., server: ...}` but other `env` values like `{config: true}` (formerly `client-only`, see #578) should be documented.

Same for `cumulative` and `effect`. A detached page is fine as this is an advanced topic.

4350pChris:
Close #1300 

I tried my hand at this. I'm not sure if the description for `env.config` is right, I tried figuring out what it does from looking at the code.


brillout:
Yes, that was actuall all accurate.

Merging, but let me know if you object with any of my changes.

PR #1433, opened 2024-01-12
Document `useData()`, and misc improvements

brillout:


brillout:
@AurelienLourot WDYT? On a tangent: as you can see, I've started using the term "component hook" instead of "hook" for `usePageContext()` and `useData()`. (I've thought a bit more about it and I came to the conclusion that calling these functions "hook" isn't *that* much of a misnomer. E.g. `useState()` is a function to "hook state onto a component". Same for `useEffect()`: it's a function to "hook an effect into a component".)

AurelienLourot (commented):
LGTM, massive work, thank you!

PR #1434, opened 2024-01-12
feat: Vike CLI

nitedani:


nitedani:
A few cases for the new prerender behavior:
examples/react-full:

vite.config.ts -> vike({prerender: true}) -> pnpm vike build -> prerenders
vite.config.ts -> vike({prerender: false}) -> pnpm vike build -> no prerender
vite.config.ts -> vike({prerender: false}) -> pnpm vike build -c={prerender:true} -> prerenders
vite.config.ts -> vike({prerender: true}) -> pnpm vike prerender-> prerenders
vite.config.ts -> vike({prerender: false}) -> pnpm vike prerender-> prerenders with a warning

vite.config.ts -> vike is not included in plugins -> pnpm vike build -> no prerender
vite.config.ts -> vike is not included in plugins -> pnpm vike build -c={prerender:true} -> prerenders
vite.config.ts -> vike is not included in plugins -> pnpm vike prerender -> prerenders
vite.config.ts -> vike is not included in plugins -> pnpm vike prerender -c={prerender:false} -> prerenders with a warning

brillout:
👍 Sounds good.

brillout:
Just had an idea, how about we support this:

```bash
VIKE_CONFIG="{prerender: false}" vike build
```

I would even go as far as to deprecate the `--config` option in favor of the `VIKE_CONFIG` env. The bottom line being that the user is supposed to only change the config over the CLI in the context of different (CI/production/staging) environements.

Likewise, the user can change Vite's config:

```bash
VITE_CONFIG="{base: '/some-base/'}" vike build
```

WDYT?

nitedani:
Ok, let's do VIKE_CONFIG and VITE_CONFIG

PR #1439, opened 2024-01-15
docs docs

brillout:


brillout:
FYI @AurelienLourot I've started to stop using the term SPA 💯; you're right it's much cleaner to avoid it. Also the docs now increasingly reference `boilerplates/` as example, and less `examples/{vue,react}-full`. I think the boilerplates can now be seen as flagship examples of using Vike without any Vike extension.

AurelienLourot (commented):
Ack, thanks for the heads-up :)

PR #1441, opened 2024-01-16
docs: minor improvements

AurelienLourot:


brillout:
👌

brillout:
Btw. I'm fine if you push directly to `main` if the confidence is only >90% that I would approve it. I can always revert/change/discuss if I object with anything in the aftermath.

You can also create a PR and merge it right after opening it, in order to ping me, but it's very much optional as I always see when commits from others are pushed to `main`.

PR #1447, opened 2024-01-18
docs: contributors and consulting

AurelienLourot:


brillout:
It's live. I like it a lot 👌

PR #1455, opened 2024-01-24
docs: `/vike-{react,vue,solid}`

AurelienLourot:


brillout:
I really like the detailed description of what `vike-react` does. That's also very neat for all the users that use Vike without Vike extension, so they can much easier piggy back on `vike-react`.

One thing I'm a little bit reluncant about is showing such details to juniors, which would happen if a junior goes to `/extensions` then cilicks on `/vike-react`. I can see that it would it scare them.

I'm thinking a neat place for such implementation details could actually be the `readme.md`, since it's then neatly collocated next to the code.

Another thing is that I think we should have a clear getting started path. It's actually already quite scattered (it's very much on my radar to fix that). I'll think about that.

AurelienLourot:
Yes I agree - let me move that to the `vike-{react,vue,solid}` repos then

brillout:
> document how Vue plugins are to be added to a vike-vue project, that's why I need to create https://vike.dev/vike-vue so that I have then the space to document that

I see. It's a good question. It's a very common use case, so indeed it needs to be documented fairly early. Couple of ideas:
 - As usual creating a page per tool `/pinia` => mention `onCreateApp()`
 - A high-level guide "Integrate tools" with subsections "React / Vue / Solid"

> A high-level guide "Integrate tools" with subsections "React / Vue / Solid"

It's probably best I work on this.

AurelienLourot:
Adding this documentation to `vike-vue` in https://github.com/vikejs/vike-vue/pull/75

AurelienLourot:
Adding this documentation to `vike-solid` in https://github.com/vikejs/vike-solid/pull/56

PR #1463, opened 2024-01-30
fix: sort styles in dev

nitedani:
The issue:
Virtual style modules are loaded too early, causing them to be overridden by other global styles.
For example, when using `vite-plugin-compiled-react` to override component level styles of Mantine:

```
    <head>
        <link rel="stylesheet" type="text/css" href="/@id/__x00__virtual:vite-plugin-compiled-react:b440be9.css?direct">
        <link rel="stylesheet" type="text/css" href="/@id/__x00__virtual:vite-plugin-compiled-react:78d174f.css?direct">
        <link rel="stylesheet" type="text/css" href="/@id/__x00__virtual:vite-plugin-compiled-react:dbd3697.css?direct">
        <link rel="stylesheet" type="text/css" href="/@id/__x00__virtual:vite-plugin-compiled-react:42b6e80.css?direct">
        <link rel="stylesheet" type="text/css" href="/@id/__x00__virtual:vite-plugin-compiled-react:6aac012.css?direct">
        <link rel="stylesheet" type="text/css" href="/@id/__x00__virtual:vite-plugin-compiled-react:65fa5c2.css?direct">
        <link rel="stylesheet" type="text/css" href="/@fs/src/node_modules/.pnpm/@mantine+core@7.4.1_@mantine+hooks@7.4.1_@types+react@18.2.47_react-dom@18.2.0_react@18.2.0/node_modules/@mantine/core/styles.css?direct">
    </head>
```

### DIFF
--- a/vike/node/runtime/renderPage/getPageAssets/retrieveAssetsDev.ts
+++ b/vike/node/runtime/renderPage/getPageAssets/retrieveAssetsDev.ts
@@ -27,7 +27,7 @@ async function retrieveAssetsDev(clientDependencies: ClientDependency[], viteDev
       collectCss(mod, assetUrls, visitedModules)
     })
   )
-  return Array.from(assetUrls)
+  return sortStyleUrls(Array.from(assetUrls))
 }
 
 // Collect the CSS to be injected to the HTML to avoid FLOUC
@@ -69,6 +69,18 @@ function isStyle(mod: ModuleNode) {
     return true
   }
   return false
+}
+
+function sortStyleUrls(styleUrls: string[]) {
+  styleUrls.sort((a, b) => {
+    const aIsVirtual = a.startsWith('/@id/')
+    const bIsVirtual = b.startsWith('/@id/')
+    if (aIsVirtual && !bIsVirtual) return 1
+    if (bIsVirtual && !aIsVirtual) return -1
+    return 0
+  })
+
+  return styleUrls
 }
 
 /*


nitedani:
Actually, this fix is not the whole picture:

It fixes the order in the initial html response, but the loading order is still incorrect when the js loads on the client:
```js
import { onRenderClient as import_0 } from "/node_modules/.vite/deps/vike-react_renderer_onRenderClient.js?v=a2a23753";
import * as import_1 from '/src/pages/cases/@id/+Page.tsx'; <<--this should go after the wrapper.tsx import
import import_2 from "/node_modules/.vite/deps/vike-react-query_renderer_FallbackErrorBoundary.js?v=a2a23753";
import import_3 from "/node_modules/.vite/deps/vike-react-query_renderer_VikeReactQueryWrapper.js?v=a2a23753";
import { Wrapper as import_4 } from '/src/pages/wrapper.tsx';  <<-- this should come before the +Page.tsx import
export const configValuesImported = [
  {
    configName: 'onRenderClient',
    importPath: 'vike-react/renderer/onRenderClient',
    isValueFile: false,
    exportValue: import_0,
    exportName: "onRenderClient",
  },
  {
    configName: 'Page',
    importPath: '/src/pages/cases/@id/+Page.tsx',
    isValueFile: true,
    exportValues: import_1,
  },
  {
    configName: 'FallbackErrorBoundary',
    importPath: 'vike-react-query/renderer/FallbackErrorBoundary',
    isValueFile: false,
    exportValue: import_2,
    exportName: "default",
  },
  {
    configName: 'VikeReactQueryWrapper',
    importPath: 'vike-react-query/renderer/VikeReactQueryWrapper',
    isValueFile: false,
    exportValue: import_3,
    exportName: "default",
  },
  {
    configName: 'Wrapper',
    importPath: '/src/pages/wrapper.tsx',
    isValueFile: false,
    exportValue: import_4,
    exportName: "Wrapper",
  },
];
...
```
How to make Wrapper load before +Page?

brillout:
> The wrapper css imports would override the page css imports, which is unexpected.

Makes sense.

There are currently no guarentees regarding CSS order. The usual way to workaround the lack of order is to make sure each CSS rule is scoped.

We can start improving the ordering and see if we can eventually converge to something that works in most situations. I'm not sure a perfect solution can exist.

What I'm thinking so far is this:
 1. Sort import order by config inheritance: least specifc configs come first. In this example, Page could come last (given Wrapper was set globally).
 2. Virtual modules come first. (So that user `.css` files override them.)
 3. `node_modules/` comes first. (So that user `.css` files override them.)

I don't what the precendence between these 3 orders should be.

The order of `1.` is determined by `sortAfterInheritanceOrder()` (also see `resolveConfigValueSources()` which sets `pageConfig.configValueSources` which is used by `getVirtualFilePageConfigValuesAll()`).

nitedani:
Continued in https://github.com/vikejs/vike/pull/1467

PR #1467, opened 2024-01-30
fix: sort config value sources

nitedani:
This fixes the issues for me, discussed in https://github.com/vikejs/vike/pull/1463
Not sure this is the right place to do it though -  but I tried, can't do it in `sortAfterInheritanceOrder`

### DIFF
--- a/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
+++ b/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig.ts
@@ -22,7 +22,9 @@ import {
   getOutDirs,
   assertKeys,
   objectKeys,
-  objectFromEntries
+  objectFromEntries,
+  higherFirst,
+  makeFirst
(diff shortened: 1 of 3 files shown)


brillout:
Yea, it's better we consolidate the ordering logic into one place. I'll have a look at it. (Let me know if it's urgent / a blocker for you.)

brillout:
I just rebased and forced-push. I hope it's ok.

brillout:
Result:
```js
                   import { onRenderHtml as import_0 } from 'vike-react/renderer/onRenderHtml';
                   import * as import_1 from '/pages/+Head.ts';
                   import * as import_2 from '/pages/+Layout.ts';
                   import * as import_3 from '/pages/index/+Page.tsx';
```

There is still potential to improve the ordering by setting a precedence between configs, but let's see if it's needed.

brillout:
Released as `0.4.160-commit-30d535e`.

PR #1494, opened 2024-02-14
chore: Documentation update about routing

Christopher2K:
Hi folks!

I'd like to submit an addition to the doc, an idea directly coming from a personal frustration. I was trying to find out why I wasn't getting the `routeParams` on client side and then, I realized that I actually need to use the `passToClient` option. 

I think it would save tons of time to colocate this piece of information with the routing doc.

What do y'all think?

Thanks for your time!

### DIFF
--- a/docs/pages/route-string/+Page.mdx
+++ b/docs/pages/route-string/+Page.mdx
@@ -17,6 +17,9 @@ export default '/film/@filmId'
 
 The value of `filmId` is available at `pageContext.routeParams.filmId`.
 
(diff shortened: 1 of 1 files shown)


brillout:
Thanks for the PR, and thanks for rising the issue.

I think https://github.com/vikejs/vike/commit/9365fa5e429831c7daf639a0edb374810e8e74a1 (this commit means that a warning will always be shown for Server Routing users) is a better way to handle this: adding `routeParams` to `passToClient` is only needed for Server Routing, while the vast majority of users (will) use Client Routing. The neat thing of this appraoch is that it keeps the docs simple for Client Routing users.

PR #1507, opened 2024-02-21
Update +Page.mdx: Optional SSR

redbar0n:
Update NextJS from 'No' to 'Limited' with references to the relevant issues.


redbar0n:
I think we'd do well to make the table as accurate as possible. Otherwise, when people discover that a 'No' doesn't actually mean it's _not possible_ (which you'd assume from a 'No'), but that more or less common workarounds exist, then the credibility of the table (and other claims) as a whole would suffer.

Besides, my impression is that actually seeing the intricacies of the workarounds is a pretty good argument for Vike in themselves.

I think providing a few footnote-ish links strikes a good balance between backing the claims made and being succinct. It was the most succinct I could get it, while still achieving the goal of being helpful.

brillout:
Ok makes sense 👍

PR #1508, opened 2024-02-21
Update +Page.mdx: Correct RSC info + Server-first section

redbar0n:
The section on React Server Components gave the incorrect picture of how they work. Also made other improvements to the section on 'Server-first orientation'.


brillout:
Hi @redbar0n, thank you for your PRs. I was recently thinking that the content of `vike.dev/nextjs` is getting quite big. I'm thinking of making it more succinct but, on the other hand, your content is very helpful for those who want to learn more. How about we externalize (some of) the content? Would be hosting it on your side be something you'd be up for? The `vike.dev/nextjs` page would then prominently link it. I feel like this would be a good solution, also because you could then quickly write content without awaiting my review (I'm currently focusing on shipping the `v1.0.0` release).

redbar0n:
Yeah, I think what most people would be after initially is the comparison table.
Then, maybe Axx0n's list (which btw should be updated to be accurate for 2024, since it dates to Jan 2021).
Then dive down more, for the reasoning.

I think simply reorganizing a bit and putting the longer content towards the bottom of the page would go a long way.

Another fix could be to make the `[i]` info notes be expandable.

I don't think externalizing the content is such a good idea, since it's all relevant for people evaluating NextJS vs Vike, and it might then get missed. I don't have a host for it either.  The length it's not really too bad, imho. As someone frequently researching options, I usually prefer a bit too much info than a bit too little. Tersifying would take more time, too. I suggest putting that on hold until you start getting feedback about it.

Luckily I don't have more content to contribute on the matter currently, so I shouldn't take up much more of your time with this page in the foreseeable future. :-) Though I can submit a PR with the aforementioned reorg. of the content, so you can judge if it helps.

brillout (approved):

brillout:
Actually, I've just read your PR and it's an excellent sum up. (I really do like your writing style. I wish it's something I'd be good at; I'm inherently not particularly good at it 🙂.)

As for your proposal to reorganize the page, that sounds like a good plan and I agree moving in-depth content to the bottom does the trick.

How about this: I can give you ownership of the `vike.dev/nextjs` page and you can create PRs as you see fit. In other words, I will only push back if I see something that is crucial to be corrected. So you can expect that I will usually just wave through your PRs.

This PR LGMT. How about we merge it and re-organize on a follow up PR? Looking forward to it.

redbar0n:
Sounds great! Thanks for the kind words <3

Your writing isn't that bad! It's always much harder to write about something when one's down in the weeds of it (i.e. in the code). Since I'm more focused on philosophy than coding these days, I have the advantage of viewing these concepts (and tech decisions) more from a few steps back, like a newbie would.

I mostly just try to write chronologically, have one main point for each sentence, and try to keep them below 22 words. I tend to re-read what I write quite a few times, and I often need to split up long trains of thought, to make them more digestible. It's not so much hocus pocus, really. :-)

brillout:
Neat, looking forward to it :-)

Makes sense and, yes, a lot of it is training. Although I think there is a style inherent to the individual's character.

PR #1519, opened 2024-02-23
`vike.config.js`

brillout:


brillout:
Moved to https://github.com/vikejs/vike/pull/1771. (Same PR but a different branch name.)

PR #1524, opened 2024-02-26
(WIP) add skipInject

iacore:
I tried to fix #1521 and failed at it.

- [ ] `pageContext.skipInject` (mostly done. still need whitelisting)
- [ ] custom URL (dist/client/abc.ext)


brillout:
I'm not sure (yet) we should go with that idea.

I'm not against it, but we should elaborate a holistic rationale before implementing the idea.

If you want, you can create a new feature request so that we can discuss the idea. I'm closing this PR in the meantime, but let's re-open and resume the PR after we discussed the idea.

PR #1535, opened 2024-03-03
Update Route type in +Page.mdx

royalswe:
Changed from Route -> RouteSync


brillout:
Thank you @royalswe. (Btw. any news on sponsoring? Definitely let me know if there is something we can do to unblock.)

Issue #695, opened 2023-03-10
[HTML placeholders for asset injection] Improve DX, performance, and docs


brillout:
### Description

`vite-plugin-ssr@0.4.92` released placeholders to control where assets are injected in the HTML:

```html
<html>
  <head>
     <meta charset="UTF-8">
     __VITE_PLUGIN_SSR__ASSETS_FIRST__
  </head>
  <body>
     <!-- ... -->
     __VITE_PLUGIN_SSR__ASSETS_LAST__
  </body>
</html>
```

**TODO 1 - Improve DX**

```js
import { html } from 'vite-plugin-ssr'

export function onRenderHtml(pageContext) {
  const pageHtml = /* ... */
  return html`<!DOCTYPE html>
    <html>
      <head>
         <meta charset="UTF-8">
         ${html.assetsFirst}
      </head>
      <body>
        <div id="page-root">${html.dangerouslySkipEscape(pageHtml)}</div>
        ${html.assetsLast}
      </body>
    </html>`
}
```

**TODO 2 - Mention this in the docs**

**TODO 3 - Improve performance**

We can and should inject assets without touching `pageHtml` which can be potentially very big.

chen6516:
> ### Description
> `vite-plugin-ssr@0.4.92` released placeholders to control where assets are injected in the HTML:
> 
> ```
> <html>
>   <head>
>      <meta charset="UTF-8">
>      __VITE_PLUGIN_SSR__ASSETS_FIRST__
>   </head>
>   <body>
>      
>      __VITE_PLUGIN_SSR__ASSETS_LAST__
>   </body>
> </html>
> ```
> 
> **TODO 1 - Improve DX**
> 
> ```js
> import { html } from 'vite-plugin-ssr/server'
> 
> export function onRenderHtml(pageContext) {
>   const pageHtml = /* ... */
>   return html`<!DOCTYPE html>
>     <html>
>       <head>
>          <meta charset="UTF-8">
>          ${html.assetsFirst}
>       </head>
>       <body>
>         <div id="page-root">${html.dangerouslySkipEscape(pageHtml)}</div>
>         ${html.assetsLast}
>       </body>
>     </html>`
> }
> ```
> 
> **TODO 2 - Mention this in the docs**
> 
> **TODO 3 - Improve performance**
> 
> We can and should inject assets without touching `pageHtml` which can be potentially very big.

as
 `__VITE_PLUGIN_SSR__ASSETS_LAST__` 

to be
[...]

PR #1539, opened 2024-03-05
injectBreakLines Regex Speedup

dajinchu:
Using node --inspect & chrome debugger, I found that [this regex line](https://github.com/vikejs/vike/blob/74f8fadfd169b22f931072db1e9d65e2b702bc35/vike/node/runtime/html/injectAssets/injectHtmlTags.ts#L121) was taking ~20ms which is significant overhead.

I don't know the context of this function and if this is a sign that I don't have Vike setup correctly (injecting too many things?), but either way this Regex can be faster if we remove the lookahead and use `replace` instead of creating and joining an array.

Testing shows this PR brings `injectBreakLines` down to 0.1ms for my use case.

### DIFF
--- a/vike/node/runtime/html/injectAssets/injectHtmlTags.ts
+++ b/vike/node/runtime/html/injectAssets/injectHtmlTags.ts
@@ -118,7 +118,7 @@ function injectBreakLines(htmlFragment: string, before: string, after: string) {
   const whitespace = `${paddingParent}${whitespaceExtra}`
   const padding = `\n${whitespace}`
 
-  htmlFragment = htmlFragment.split(/<(?=[^\/])/).join(`${padding}<`)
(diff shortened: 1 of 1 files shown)


brillout:
:+1: It seems equivalent so LGTM.

>  is a sign that I don't have Vike setup correctly

Not that I can tell.

FYI the plan is to eventually implement this: https://github.com/vikejs/vike/issues/695. This should solve the underlying issue once and for all.

brillout:
Merged by https://github.com/vikejs/vike/commit/b3cc3cb82439a42e2d14ef5b34d70131f7c8e152 and pre-released as `0.4.165-commit-b3cc3cb`.

Thanks for the PR and, as always, thanks for sponsoring :green_heart:. Let me know if there is any other issue.

Issue #1333, opened 2023-12-08
docs: normalize example lists (e.g. prefix all example entries with a timestamp)


brillout:
### Description

1. Prefix all example list entries with `<Timestamp>`, i.e. apply https://github.com/vikejs/vike/commit/d9529f641f4602c01151c7abf47cfafd24f690a1#diff-ba9c034a3139ccb209778aede367dda968133e5f52675451202a56034a9b5dfb to all example lists.
2. Always use the `<Github>` component, i.e. apply 9be47fc51a222d5bd4d495f40b0d0a07604675cd to all example list.

PR #1556, opened 2024-03-16
docs: normalize example list for `deploy` section

Kenzo-Wada:
related: https://github.com/vikejs/vike/issues/1333

I will continue to work on this Issue in parallel with the creation of the Japanese documentation so that conflicts occur as little as possible:)


brillout:
Thank you & sounds good!

Issue #1557, opened 2024-03-17
Electron and file:// protocol


rustytsuki:
### Description

When I build vike app as ssg pages. Using taruti to load these pages worked fine. But after use electron to load the static page through file:// protocol, I received an exception in source code:
vike/utils/parseUrl.ts
```
const PROTOCOLS = ['http://', 'https://', 'tauri://']

function isParsable(url: string): boolean {
  // `parseUrl()` works with these URLs
  return (
    PROTOCOLS.some((p) => url.startsWith(p)) ||
    url.startsWith('/') ||
    url.startsWith('.') ||
    url.startsWith('?') ||
    url.startsWith('#') ||
    url === ''
  )
}
```
With my due respect, I think it would be a good idea to support file::// protocol.
Or, any suggestion for me to use vike generated static pages in electron?

lencil:
Here is an ugly workround, Because vike has supported tauri protocol, so implementing a custom protocol with name of "tauri://" of electron that has the same effect as the file://.
https://www.electronjs.org/docs/latest/api/protocol

However, I also think supportting "file://" protocol would be better.

brillout:
Does adding `'file://'` to `const PROTOCOLS = ['http://', 'https://', 'tauri://']` fix the issue?

PR welcome, but let me know if it isn't that trivial of a change and I'll implement it.

rustytsuki:
> Does adding `'file://'` to `const PROTOCOLS = ['http://', 'https://', 'tauri://']` fix the issue?
> 
> PR welcome, but let me know if it isn't that trivial of a change and I'll implement it.

It does work. Here is PR: https://github.com/vikejs/vike/pull/1559

brillout:
:+1: Fix pre-released as `vike@0.4.165-commit-8eba585`.

PR #1559, opened 2024-03-18
support Electron by adding 'file://' protocol

rustytsuki:
I've tested it in Electron, and work fine with 'file://' protocol.


rustytsuki:
https://github.com/vikejs/vike/issues/1557

Issue #1350, opened 2023-12-12
Support Vike


brillout:
If your company is using Vike to build a revenue-generating product please consider sponsoring, as it makes a tremendous impact.

- [Vike sponsoring](https://github.com/sponsors/vikejs)
- [Rom sponsoring](https://github.com/sponsors/brillout) (sponsor the Vike organization instead if possible)

> [!NOTE]
> When sponsoring you get:
> - Priority bump (support, bug fixes, feature requests)
> - Extra help from the Vike team
>
> It's a unique opportunity for tight-knit collaboration, such as implementing requirements that aren't possible to implement with other frameworks.

royalswe:
What alternatives are there if the company does not want to enter card numbers on Github?
I saw that there were invoice alternatives, but the requirements is very high to use it.
Requirements: https://docs.github.com/en/enterprise-cloud@latest/sponsors/sponsoring-open-source-contributors/paying-for-github-sponsors-by-invoice#requirements

I mentioned virtual cards and maybe they'll pick up on that.
Our company has an enterprise Github account.

brillout:
@royalswe Thanks, we hugely appreciated it 💚 We can setup a paypal / bank wire / invoicing directly between Vike and your company. Which one would you prefer?

AurelienLourot:
@royalswe Thank you for considering sponsoring. You can also find us on [discord](https://discord.com/invite/hfHhnJyVg8) if you'd like to discuss details privately via direct messages

royalswe:
Thanks! I will tell my manager. If I have further questions I will contact via discord.

brillout:
Closing, as we'll focus on the [Open Source Pricing](https://vike.dev/pricing). That said, sponsoring welcome for tight-knit collaborations.

PR #1568, opened 2024-03-20
i18n doc: pass query params to 'urlLogical'

lebretont:
Pass the query parameters to 'urlLogical' in the i18n doc.
So the pages can access the parameters.

### DIFF
--- a/docs/pages/i18n/+Page.mdx
+++ b/docs/pages/i18n/+Page.mdx
@@ -8,22 +8,20 @@ You can internationalize (i18n) your app by using the <Link text={<><code>onBefo
 export { onBeforeRoute }
 
 function onBeforeRoute(pageContext) {
-  const { urlWithoutLocale, locale } = extractLocale(pageContext.urlPathname)
-
-  const searchOriginal = pageContext.urlParsed.searchOriginal ?? "";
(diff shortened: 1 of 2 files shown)


brillout:
Isn't it implicit that `extractLocale()` would preserve it?

I'm inclined to not add a note about this because as it would dilute the core message of the docs.

Closing but feel free to continue discussing this.

lebretont:
> Isn't it implicit that `extractLocale()` would preserve it?
> 
> I'm inclined to not add a note about this because as it would dilute the core message of the docs.
> 
> Closing but feel free to continue discussing this.

`extractLocale()` doesn't preserve it because we are using `urlPathname`.
Only `urlOriginal` has the query params

brillout:
Indeed. How

brillout:
How about this?

brillout:
FYI `pageContext.urlParsed` is going to be renamed `pageContext.url` is the next major release.

lebretont:
Cleaner than mine!

brillout:
Merged :tada: Thank you for the PR and pushing through despite my initial push back.

(In case it's an option for your company: we're looking for [sponsors](https://github.com/vikejs/vike/issues/1350).)

PR #1575, opened 2024-03-24
Add the "dynamicAssetImportFilter" callback to allow correctly injecting css for async vue components

MichaelRoosz:
This pr adds the "dynamicAssetImportFilter" callback to allow correctly injecting css assets for async vue components.

As discussed in https://github.com/vikejs/vike/discussions/1324, when vue components are defined async with `defineAsyncComponent()` the current situation in Vike is:

- In **dev** mode:
  - once an async component has been used (loaded) by any request, its assets will be injected for all following requests
  - after some time, given a larger amount of async components, this may result in dozens of css assets being injected despite them not being used at all on the current page
  
 - in **prod** (build) mode:
   - because `retrieveAssetsProd()`currently ignores `dynamicImports` in the manifest entry, assets of async loaded components do not get injected at all resulting in FOUC (Flash of unstyled content)
   
 The idea of  of the `dynamicAssetImportFilter` callback is to allow user code to keep track of which async components have been used for the current request and then filter the dynamically imported assets for those components.
 
 
 Initially I was thinking about extending the existing `injectFilter` callback, but I discovered some problems with this idea:
 - it is not being used in dev mode currently
 - it is not being used to filter the "early hints" http "Link" headers currently, which results in the browser still loading the assets that have `inject` set to `false`.
 - `InjectFilterEntry` would have to be extended by something like `importedBy: [ { importer: string, type: "static"|"dynamic" }, .. ]` which could result in a lot of overhead data and needs some more complicated checks to generate properly
 - performance wise I think it is best to ignore unneeded assets/modules in the first place, so I went with the `dynamicAssetImportFilter` idea
 
 
 Please regard this implementation as a proposal, I am willing to modify and/or extend it in any way to meet your requirements and wishes.
 
 The "real world" use case for this is: The pages I need to render are fully dynamic - I get a list of components to render along with their properties from an api. I cannot know which components I need to render for a page and thus I want to load them all async (`defineAsyncComponent()`) and dynamic with the vue `<component :is="component" />` syntax.
 
 Vike a really awesome framework and I think the only one where this is currently possible.
 
 Thank you for your feedback! I am grateful for any chance getting this merged and will refine this in any way you want.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1703,6 +1703,27 @@ importers:
         specifier: ^5.0.10
         version: 5.0.10(@types/node@20.10.5)
 
+  test/dynamicAssetImportFilter-vue:
+    dependencies:
+      '@vitejs/plugin-vue':
+        specifier: ^5.0.4
+        version: 5.0.4(vite@5.2.5)(vue@3.4.21)
+      '@vue/compiler-sfc':
+        specifier: ^3.4.21
+        version: 3.4.21
+      '@vue/server-renderer':
+        specifier: ^3.4.21
+        version: 3.4.21(vue@3.4.21)
+      vike:
+        specifier: link:../../vike
+        version: link:../../vike
+      vite:
+        specifier: ^5.2.5
+        version: 5.2.5
+      vue:
+        specifier: ^3.4.21
+        version: 3.4.21(typescript@4.8.4)
+
   test/env:
     dependencies:
       '@vitejs/plugin-react':
@@ -2168,6 +2189,14 @@ packages:
 
   /@babel/parser@7.23.5:
     resolution: {integrity: sha512-hOOqoiNXrmGdFbhgCzu6GiURxUgM27Xwd/aPuu8RfHEZPBzL1Z54okAHAQjXfcQNwvrlkAmAp4SlRTZ45vlthQ==}
+    engines: {node: '>=6.0.0'}
+    hasBin: true
+    dependencies:
+      '@babel/types': 7.23.5
+    dev: false
+
+  /@babel/parser@7.24.1:
+    resolution: {integrity: sha512-Zo9c7N3xdOIQrNip7Lc9wvRPzlRtovHVE4lkz8WEDr7uYh/GMQhSiIgFxGIArRHYdJE5kxtZjAf8rT0xhdLCzg==}
     engines: {node: '>=6.0.0'}
     hasBin: true
     dependencies:
@@ -4990,6 +5019,17 @@ packages:
       vue: 3.3.10(typescript@4.8.4)
     dev: false
 
+  /@vitejs/plugin-vue@5.0.4(vite@5.2.5)(vue@3.4.21):
+    resolution: {integrity: sha512-WS3hevEszI6CEVEx28F8RjTX97k3KsrcY6kvTg7+Whm5y3oYvcqzVeGCU3hxSAn4uY2CLCkeokkGKpoctccilQ==}
+    engines: {node: ^18.0.0 || >=20.0.0}
+    peerDependencies:
+      vite: ^5.0.0
+      vue: ^3.2.25
+    dependencies:
+      vite: 5.2.5
+      vue: 3.4.21(typescript@4.8.4)
+    dev: false
+
   /@vitest/expect@1.3.1:
     resolution: {integrity: sha512-xofQFwIzfdmLLlHa6ag0dPV8YsnKOCP1KdAeVVh34vSjN2dcUiXYCD9htu/9eM7t8Xln4v03U9HLxLpPlsXdZw==}
     dependencies:
@@ -5077,6 +5117,16 @@ packages:
       source-map-js: 1.0.2
     dev: false
 
+  /@vue/compiler-core@3.4.21:
+    resolution: {integrity: sha512-MjXawxZf2SbZszLPYxaFCjxfibYrzr3eYbKxwpLR9EQN+oaziSu3qKVbwBERj1IFIB8OLUewxB5m/BFzi613og==}
+    dependencies:
+      '@babel/parser': 7.24.1
+      '@vue/shared': 3.4.21
+      entities: 4.5.0
+      estree-walker: 2.0.2
+      source-map-js: 1.2.0
+    dev: false
+
   /@vue/compiler-dom@3.2.33:
     resolution: {integrity: sha512-GhiG1C8X98Xz9QUX/RlA6/kgPBWJkjq0Rq6//5XTAGSYrTMBgcLpP9+CnlUg1TFxnnCVughAG+KZl28XJqw8uQ==}
     dependencies:
@@ -5096,6 +5146,13 @@ packages:
     dependencies:
       '@vue/compiler-core': 3.3.8
       '@vue/shared': 3.3.8
+    dev: false
+
+  /@vue/compiler-dom@3.4.21:
+    resolution: {integrity: sha512-IZC6FKowtT1sl0CR5DpXSiEB5ayw75oT2bma1BEhV7RRR1+cfwLrxc2Z8Zq/RGFzJ8w5r9QtCOvTjQgdn0IKmA==}
+    dependencies:
+      '@vue/compiler-core': 3.4.21
+      '@vue/shared': 3.4.21
     dev: false
 
   /@vue/compiler-sfc@3.2.33:
@@ -5143,6 +5200,20 @@ packages:
       source-map-js: 1.0.2
     dev: false
 
+  /@vue/compiler-sfc@3.4.21:
+    resolution: {integrity: sha512-me7epoTxYlY+2CUM7hy9PCDdpMPfIwrOvAXud2Upk10g4YLv9UBW7kL798TvMeDhPthkZ0CONNrK2GoeI1ODiQ==}
+    dependencies:
+      '@babel/parser': 7.24.1
+      '@vue/compiler-core': 3.4.21
+      '@vue/compiler-dom': 3.4.21
+      '@vue/compiler-ssr': 3.4.21
+      '@vue/shared': 3.4.21
+      estree-walker: 2.0.2
+      magic-string: 0.30.8
+      postcss: 8.4.37
+      source-map-js: 1.2.0
+    dev: false
+
   /@vue/compiler-ssr@3.2.33:
     resolution: {integrity: sha512-XQh1Xdk3VquDpXsnoCd7JnMoWec9CfAzQDQsaMcSU79OrrO2PNR0ErlIjm/mGq3GmBfkQjzZACV+7GhfRB8xMQ==}
     dependencies:
@@ -5162,6 +5233,13 @@ packages:
     dependencies:
       '@vue/compiler-dom': 3.3.8
       '@vue/shared': 3.3.8
+    dev: false
+
+  /@vue/compiler-ssr@3.4.21:
+    resolution: {integrity: sha512-M5+9nI2lPpAsgXOGQobnIueVqc9sisBFexh5yMIMRAPYLa7+5wEJs8iqOZc1WAa9WQbx9GR2twgznU8LTIiZ4Q==}
+    dependencies:
+      '@vue/compiler-dom': 3.4.21
+      '@vue/shared': 3.4.21
     dev: false
 
   /@vue/devtools-api@6.5.1:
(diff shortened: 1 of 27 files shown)


brillout:
Hi Michael, thank your for your PR.

Indeed, being able to determine whether and what dynamic imports should be preloaded would be nice.

How about making it a config?

```js
// /pages/+config.js

export default {
  preloadDynamicImports: true,
  // Or regex:
  preloadDynamicImports: /\.png$/i,
  // Or function:
  preloadDynamicImports: (url) => url.endsWith('.png'),
}
```

If we add `pageContext` to `preloadDynamicAssets(url, pageContext)` then I guess that would be strictly more powerful.

I'm just brainstorming; we can first discuss before taking actions.

MichaelRoosz:
Hello @brillout ,
a config setting could also work, but for my use case I would need to know who (which module) is importing something. I do not care about the url of the asset itself. So it would be more something like `preloadDynamicAssets(importer, pageContext)`, or if you prefer `preloadDynamicAssets(url, importer, pageContext)` - but then this would need to get called for every asset and we would need to go through the whole module tree, while I would like to avoid this for performance reasons (currently my code is skipping the whole module path if `preloadDynamicAssets(importer)`return false

brillout:
> a config setting could also work

I think using a config instead of a `pageContext` prop is a better approach, and it's an orthogonal question to other considerations such as whether to pass `importer`. The `injectFilter()` feature is an exception for legacy reasons and it has issues: I'm actually looking forward to make it a config instead.

>  for my use case I would need to know who (which module) is importing something

I see. Yes, it would indeed mean that the function is called not only for every node in the module graph but also for every edge, which I can see to be a confusing DX.

> this would need to get called for every asset and we would need to go through the whole module tree, while I would like to avoid this for performance reasons (currently my code is skipping the whole module path if `dynamicAssetImportFilter(importer)` returns false.

Hm, it has to be tested I think. Also, I presume not many users would need this for `node_modules/`, so I guess we can skip `noExternal` packages.


> Also always walking the full module tree would create an issue - a dynamically loaded module may have "normally" (statically) loaded modules - if we want to ignore it, but still travel down its children, we would then add the assets from the statically loaded child-modules, which we likely do not want.

> Option A

I ain't sure I see the issue it's trying to solve. Not sure if it's about performance or about your use case.

In general, the more I know about your use case the better.

> Option B

For DX reasons, I'm a bit reluctant about this as I can see it to be potentially confusing for users, especially juniors.

<hr>

Thanks for your double-indie sponsoring, I definitely appreciate it. Speaking of which, is sponsoring something your employer would be up for? In case it's also Munich based, I'm happy to spontaneously pass by. Although the expectation on my side is that your employer is open for building trust.

One thing to consider about this PR is the fact that there are a couple of things I want to change about `injectFilter()` and that I'd be pushing for having a single function to cover a maximum amount of uses cases in an elegant way. I understand this means more work for the PR author, but I can take over the implementation if that's something you'd prefer.

All-in-all, this PR seems to be about a fairly specific use case and solving it seems fairly complex, thus I'm labeling it as `low-prio`.  See the [issue labels](https://github.com/vikejs/vike/labels) in case you're curious to see where I come from. (Sponsoring would significantly bump prioritization.)

MichaelRoosz:
I see why you want to go with just one filter function, but consider a bigger project with many components (maybe a few hundred). If all of these are loaded async like in my case, then for a single page request the filter function would need to be several hundred times if each component has at least one asset. And most pages will likely only use a few components, resulting in only maybe 10 out of 200 filter calls to be needed.

My use case is a very important one: supporting css injection for async components. Without support for this, it will be impossible to load components async without having the page flicker.

Here is a comprise we maybe can both agree on:
- we make `injectFilter()` a config
- we also call it for dev mode (this a a hard requirement to support async components)
- we make it also filter early hints aka the "Link" http headers
- instead of first collecting all the assets and then calling  `injectFilter()`, we call it while collecting the assets.
- we extend `InjectFilterEntry ` like this:
```
type InjectFilterEntry = {
  inject: false | 'HTML_BEGIN' | 'HTML_END' // Whether and where to inject
  src: string // Asset's URL
  assetType: "image" | "script" | "font" | "style" | "audio" | "video" | "document" | "fetch" | "track" | "worker" | "embed" | "object" | null
  mediaType: string // MIME type
  isEntry: boolean // true  ⇒ <script> or <link rel="stylesheet" type="text/css">
                   // false ⇒ preload tag, e.g. <link rel="preload" as="font">
  earlyHint: boolean // whether to early hint it via a http Link header
  importer: string // the module that imported it
  importType: "static"|"dynamic" // in dev mode type is always "dynamic", we could also use the value "dynamic_dev" here
  isImporter: boolean
}
````
- we add another config value (default: false), to have injectFilter also called for importers:
```
callInjectFilterForImporters: boolean
```

About sponsoring, yes my company has offices in Munich, and maybe we ill be sponsoring vike in the future. But it would need to be useful to us. Right now we are in a proof-of-concept phase, trying to figure out which framework works for us. Without support for dynamic components we will have to choose a different one.

**Comment by MichaelRoosz** [comment]
a slight optimization:

we only call `injectFilter()` during collection for the importers (when callInjectFilterForImporters = true). for "normal" assets we call it like now after collection.

in this case we would put 
````
 importer: string // the module that imported it
 importType: "static"|"dynamic" // in dev mode type is always "dynamic", we could also use the value "dynamic_dev" here
```
into an array to list all importers

brillout:
To be transparent, I didn't think everything through and only skimmed over your PR. So I may very well be out-of-touch with certain aspects of this. I'm open to revisit prioritization in which case I'll dig and think everything through.

> for a single page request the filter function would need to be several hundred times

> * instead of first collecting all the assets and then calling  `injectFilter()`, we call it while collecting the assets.

I don't see (yet) why this is needed. I don't expect going through the entire module graph to be a performance issue.

So, instead, I'm leaning more towards calling `injectFilter()` only once per request while passing more information to it.

In principe, I presume that if the user has access to the entire module graph then the user has enough information to know what dynamic import should be preloaded. I ain't saying it's a good DX (let's see if we can find one), but I'm saying that I assume that we can achieve the aforementioned use cases while calling `injectFilter()` only once.

> we also call it for dev mode (this a a hard requirement to support async components)

Makes sense, and it's actually a major thing I'm unhappy about `injectFilter()`.

> * we extend `InjectFilterEntry ` like this:

Sounds good.

> My use case is a very important one: supporting css injection for async components. Without support for this, it will be impossible to load components async without having the page flicker.

I see.

> About sponsoring, yes my company has offices in Munich, and maybe we will be sponsoring vike in the future. But it would need to be useful to us. Right now we are in a proof-of-concept phase, trying to figure out which framework works for us. Without support for dynamic components we will have to choose a different one.

Sounds good.

Depending on the sponsoring tier your company is open to consider, I'm open to label this as `high-prio` and maybe even `highest-prio` with an implementation ETA of 1-2 weeks. (It would probably be faster if I take over the implementation then.)

Beyond this PR, do you foresee any other potential blockers for using Vike?

Given this is a proof-of-concept phase, I understand the situation may be ambiguous; meeting in person may (soon or eventually) make sense.

MichaelRoosz:
okay, I can understand your thoughts.

then I think we would agree on:

- we make injectFilter() a config, it is called once per request
- we also call it for dev mode
- we make it also filter early hints aka the "Link" http headers
- we extend InjectFilterEntry like this:
```
type InjectFilterEntry = {
  inject: false | 'HTML_BEGIN' | 'HTML_END' // Whether and where to inject
  src: string // Asset's URL
  assetType: "image" | "script" | "font" | "style" | "audio" | "video" | "document" | "fetch" | "track" | "worker" | "embed" | "object" | null
  mediaType: string // MIME type
  isEntry: boolean // true  ⇒ <script> or <link rel="stylesheet" type="text/css">
                   // false ⇒ preload tag, e.g. <link rel="preload" as="font">
  //
  // new props start here:
  //
  earlyHint: boolean // whether to early hint it via a http Link header
  importedBy: Array{
    importer: string // the module that imported it
    type: "static"|"dynamic" // in dev mode type is always "dynamic"
  },
}
```

thank you for the offer with the the priority changes. but to be honest I would like to first establish a good relation with you and the vike project without money on the table.

I can offer you to do all the work for this pr until you can happily merge it. then, very likely, my team will use vike long-term for the websites my team manages which will give me some good arguments to persuade management why we should sponsor you. Still I do not want to get your hopes up too high, I think we would land around bronze / $200 level.

but in any case we for sure can meet in person and share our ideas if you are interested. I would invite you to our office mid-april.

brillout:
> we make injectFilter() a config

We don't have to do this in this PR. If anything it's probably better we postpone this because this PR would introduce a breaking change otherwise.

> is this something you would like?

Sounds good and it seems like this will get the PR closer to mergeability.

> I think we would land around bronze / $200 level.

It's the lowest company sponsoring tier so it gives only a minor priority bump.

> will give me some good arguments to persuade management why we should sponsor you

What's also critical is the fact that it's (very?) likely that your company may never end up sponsoring: persuading management to give out money without anything in exchange is notoriously hard and close to impossible in many kinds of companies.

At the end of the day, it boils down to why should Vike prioritize the needs of a single user over the needs of many other users. The priorities listed in https://github.com/vikejs/vike/issues / https://github.com/vikejs/vike/labels show that there are many other issues with much higher priorities to work on.

> if yes, I would update my pr until tomorrow evening.

You're welcome to continue the PR but note that I will have only limited capacity to review and think through the design, given I'm still seeing this ticket as `low-prio`.

MichaelRoosz:
I get that you want to earn money with this project, but to be honest, I do not like this money milking mentality.
I was willing to do all the work for this issue, exactly to your specification, you would only have to check if the code (which will only be a few lines) meets your wishes. If you value your time and work so much, is my work and time nothing worth to you?
Yes , I see that you have open issues from other people, but did those create pull requests including some tests even?
Is this the open source mentality? 
I was about to promote your framework to my friends and colleagues at work, but I do not think I will peruse this further.
I will put a "low prio" label on your framework and go back to the scratch board, maybe I can work it with Nuxt....

brillout:
> likely that your company may never end up sponsoring

Alternatively, we can have a collaboration between your company and Vike on a contractual basis and only focus on resolving your current blockers. Feel free to PM me for details.

brillout:
Why should Vike focus on your very specific need and not on needs that would affect and be beneficial to a lot more users?

As for the "money milking" part, I clearly wouldn't do Open Source if I'd be interested in personal wealth. Like everyone, I need to pay my bills.

> I was willing to do all the work for this issue, exactly to your specification, you would only have to check if the code (which will only be a few lines) meets your wishes. If you value your time and work so much, is my work and time nothing worth to you?

I do appreciate your work. And, again, thank you for opening the PR. The issue is that taking design decisions is the most time and energy consuming part of implementing new features. And it's unfortunately work that cannot be delegated to others.

If Nuxt works better for you, then I understand that you'll be migrating to Nuxt. If your migration to Nuxt doesn't work out then my offer to have contractual work between your company and Vike still stands.

Let's close this in the meantime. But let me know if you still want to give this PR a shot, I can re-open it. (Although, again, I will have only limited capacity to review and take design decisions.)

> I was about to promote your framework to my friends and colleagues at work, but I do not think I will persue this further.

Thank you I appreciate it and I'm sorry you don't feel like this anymore. I really wish the monetary aspect of Open Source wouldn't be as tricky as it is.

brillout:
I'm thinking, one thing we could do is that we merge the new feature of this PR as an internal/experimental/undocumented feature. I think / I'm hopefully we can then merge this fairly quickly without too much effort. Re-opening, let me know what you think about this.

MichaelRoosz:
This is a kind offer which I would have taken yesterday, but we already moved on.
We now use the [ssr-vue](https://github.com/vitejs/vite-plugin-vue/tree/main/playground/ssr-vue) example from `vite-plugin-vue` as a foundation, added some parts like https://github.com/vitejs/vite/issues/2282#issuecomment-1079659649 for asset preloading for dev mode, and wrote the missing parts ourselves.

It also turned out, that vue.js actually already supplies [a list of module IDs](https://vitejs.dev/guide/ssr#generating-preload-directives) that have been used during rendering:
```
const ctx = {}
const html = await renderToString(app, ctx)
// ctx.modules is now a Set of module IDs that were used during the render
```
This allows us now to directly look up the assets for those module IDs.

I will not annoy you again with my proposals, but I would assume than an option to directly forward this module id list to vike could be useful.

Anyway, even though I think we are two very different people and this pull request was very stressful to me, I do not regret sending you the $100 - I wish you all the best for your project - live long and prosper!

brillout:
Thanks for circling back on this.

Indeed, directly using Vite's SSR API is a highly flexible solution. Although it may mean that you'll end up reinventing the wheel, your mileage may vary.

Too bad we failed at building trust. (I did some extensive introspection and I won't repeat a couple of mistakes I made.)

Thank you for your donation, I appreciate it and I'll re-invest it into Open Source.

Shall Vike, one day, sound appealing to you again, the door is open.

All the best & kind regards.

PR #1582, opened 2024-03-28
Fix PushState MonkeyPatch

dajinchu:
## Problem

When Vike monkey patches `history.pushState`, it saves the "original" pushState to call during Vike's routing. However, 3rd party libraries also monkey patch `pushState` in order to be informed when the page changes (in my case, [Amplitude](https://github.com/amplitude/Amplitude-TypeScript)). These libraries will not be informed when Vike navigates.

## Solution

This PR changes Vike's navigation to go through `history.pushState` so that any monkey patches that happen _after_ Vike's will still run. It detects whether Vike's own navigation is happening to no-op the monkey patch.

### DIFF
--- a/vike/client/client-routing-runtime/history.ts
+++ b/vike/client/client-routing-runtime/history.ts
@@ -7,10 +7,7 @@ export {
   monkeyPatchHistoryPushState
 }
 
-import { assert, assertUsage, getGlobalObject, hasProp, isObject } from './utils.js'
-const globalObject = getGlobalObject<{
-  pushStateOriginal?: PushStateOriginal
-}>('history.ts', {})
+import { assert, assertUsage, hasProp, isObject } from './utils.js'
 
 // No way found to add TypeScript types to `history.state`: https://github.com/microsoft/TypeScript/issues/36178
 type HistoryState = {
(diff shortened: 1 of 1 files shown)


vike/client/client-routing-runtime/history.ts
```diff
@@ -97,27 +100,27 @@ function replaceHistoryState(state: HistoryState, url?: string) {
...
-      // Don't allow user to overwrite triggeredBy as it would break Vike's handling of the 'popstate' event
-      triggeredBy: 'user'
+  window.history.pushState = new Proxy(window.history.pushState, {
+    apply: (target, thisArg, [stateFromUser, unused, url]) => {
+      assertUsage(
+        null === stateFromUser || undefined === stateFromUser || isObject(stateFromUser),
+        'history.pushState(state) argument state must be an object'
+      )
+      const state: HistoryState = isVikeHistoryState(stateFromUser)
+        ? stateFromUser
```
dajinchu, line 115: so technically, in userland I could call `history.pushState({timestamp: 1, scrollPosition: 2, triggeredBy: "me"})` and we would trust it, so this is not the most robust system.

We could create a `Symbol` in this scope if we really want to be sure that no userland dev can accidentally bypass this check.

brillout:
I see.

I wonder whether `new Proxy()` is needed; I just pushed [a minimal solution](https://github.com/vikejs/vike/pull/1582/commits/13302ca509be2f8883fbc478309eb32a2a1d8d82), let me know what you think.

dajinchu:
Minimal solution works for me.

brillout:
Pre-released as `0.4.167-commit-cfadd0a`.

dajinchu:
Awesome thanks!

(I used Proxy because I ran into the `this` binding issue as well)

brillout:
Sure thing, let me know if there is anything else!

PR #1588, opened 2024-04-01
make .server.js and .client.js change meta.env

nitedani:


brillout:
Neat. One thing I was thinking about is that we could have a third extension e.g. `.shared.` (or some other word but so far I like `.shared.` most). WDYT?

As for `assertUsage()`, how about we set the env props to `true` instead of `assertUsage()`?

brillout:
I guess we can update the tests?

brillout:
LGTM :tada:

It'd be good if we update the docs at https://vike.dev/file-env. Up for doing that or should I do it? Whichever you prefer.

brillout:
(I'll fix the new bug later today.)

brillout:
Merged :sparkles:

As for the docs let me know if you're up for it (new PR welcome), ohterwise not a problem I can also do it.

Thanks for the PR, looking forward to more (if you want) :slightly_smiling_face:

PR #1601, opened 2024-04-05
don't use options api in examples

### LINKED ISSUE #1581: Examples + vike-vue: use composition API to support composition-only build of vue
### Description

Something I stumbled upon some time ago:
The examples and vike-vue use the options API to define the root component & integrate it with routing: https://github.com/vikejs/vike/blob/main/examples/vue-full/renderer/app.ts

Because of this, the app breaks when you set [`_VUE_OPTIONS_API_: false`](https://vuejs.org/api/compile-time-flags.html#VUE_OPTIONS_API), a compile-time flag to exclude code related to the options API and reduce bundle size. Using the composition API is just a tiny change (using a ref instead of `data()` for `this.Page`), happy to submit a PR :)

jonaskuske:
close #1581 

Also see https://github.com/vikejs/vike-vue/pull/81

### DIFF
--- a/boilerplates/boilerplate-vue-ts/renderer/app.ts
+++ b/boilerplates/boilerplate-vue-ts/renderer/app.ts
@@ -1,32 +1,17 @@
 export { createApp }
 
-import { createSSRApp, defineComponent, h, markRaw, reactive } from 'vue'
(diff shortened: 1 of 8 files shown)


brillout:
Very nice. Thanks @jonaskuske for the PR.

PR #1602, opened 2024-04-08
Update ClientOnly documentation for vike-solid

magne4000:


brillout:
:+1:

I reverted the `.gitignore` change but let me know if you believe we should keep the `/.idea/` entry.

PR #1605, opened 2024-04-10
add docs for config env

nitedani:


brillout:
I think it's quite a neat way to change the `meta.env`, so I've also added a couple of mentions.

PR #1617, opened 2024-04-22
feat: [experimental] prefetch page context (#246)

### LINKED ISSUE #246: Link Prefetching: also prefetch `pageContext`
Currently, there's no way to do any user-defined additional preloading jobs after finishing to preload static assets. So it's impossible to prefetch data needed for routes. Having an `onAfterPreload` hook (tried to match the naming convention) will solve the problem and it'll be very helpful to deeply tune the page performance.

usk94:
resolve #246

This PR introduces a feature to prefetch `pageContext` (specifically focusing on `data`) when users hover over a link.

To ensure efficient network usage, the prefetch operation includes a cooldown period to prevent it from being re-triggered too soon after the initial action.
And if the user hovers over multiple links, `pageContext` fetched for previous links can be used instead of being discarded.

Users can configure this feature through `+config.ts` or by setting `data-prefetch-page-context` attributes directly on anchor tags.


usk94:
@brillout 
I would appreciate some guidance on how to determine when the prefetch of `pageContext` is complete. 
While I have managed to fetch the context containing `config`, `exportsAll`, etc., when hovering over a link, I am unsure how to carry this data over to the rendering of the next page after navigation. 

Any advice on this would be greatly helpful. Thank you in advance for your assistance.

usk94:
@brillout 
Thank you for your response; I await your further insights.

brillout:
@usk94 Alright, let's do this!

Have a look at `renderPageClientSide.ts`, that's where the whole client-side rendering orchestration happens. This file has a `globalObject` so I think we can save the prefetched `pageContext` there. As for the prefetching itself, how about we use `getPageContextFromHooks_isNotHydration()` to do the pre-fetching?

Open question: do you think viewport prefetching makes sense for `pageContext` pre-fetching? This seems quite aggressive, maybe too aggressive. I'd be inclined to not even offer this setting :thinking:

usk94:
@brillout 
> Open question: do you think viewport prefetching makes sense for pageContext pre-fetching? This seems quite aggressive, maybe too aggressive. I'd be inclined to not even offer this setting 🤔

Are you concerned that, for example, it might be too aggressive in terms of saturating the network with too many API requests?

For instance, in frameworks like Astro, when viewport prefetching is enabled, the priority of these requests is managed to be lower to mitigate potential network congestion.
https://docs.astro.build/en/guides/prefetch/

But even with that consideration, I agree that it might be too aggressive. I think starting with hover prefetching would be sufficient.

brillout:
>  todo: getPageContextFromHooks_isNotHydration return does not contains "urlOriginal", "Page".
>   // maybe I need to merge with pageContext(props).

Yes, that's expected. It only fetches the `pageContext` from hooks, most notably `onBeforeRender()` and `data()` (but *not* `onRenderClient()`).

I was thinking whether we should also call `onRenderClient()` upon prefetching but that isn't trivial (it would need to render the next page in some kind of virtual DOM). So I'm thinking that only using `getPageContextFromHooks_isNotHydration()` is fine for now.

Let me know if you think `getPageContextFromHooks_isNotHydration()` doesn't fit, I can do some refactoring.

> Are you concerned that, for example, it might be too aggressive in terms of saturating the network with too many API requests?

Yes, and also overloading the server (its database and logs).

> But even with that consideration, I agree that it might be too aggressive. I think starting with hover prefetching would be sufficient.

:+1: We could also, eventually, add support for viewport prefetching for individual links that opt in using the `data-` attribute but I think we can skip that for now.

> Astro

Indeed Astro seems to be using [clever heuristics](https://docs.astro.build/en/guides/prefetch/#prefetch-strategies:~:text=Each%20strategy%20is%20fine%2Dtuned%20to%20only%20prefetch%20when%20needed%20and%20save%20your%20users%E2%80%99%20bandwidth.%20For%20example%3A). If you want we can improve Vike's prefetching in a follow up PR after this one. I also have a couple of ideas for improving static asset prefetching.

brillout:
@Boeing787 (sponsor who requested this feature) Do you have any special wishes? Your use case is that you want data to be prefetched when the user hovers a on link, correct?

Boeing787:
@brillout correct, to load the pageContext

usk94:
@brillout 
I think the PR is now ready for review. Could you please take a look? 

Also, the CI test has been failing since [the second to last commit](https://github.com/vikejs/vike/pull/1617/commits/077d865e3ac16f29f0d3b177535690c84927094a), and I'm unable to pinpoint the issue. 
If possible, could you provide some insight into what might be going wrong?

brillout:
Hi Yusaku, thank you.

Looks good at first sight, let me have a closer look.

> some CI tests has been failing since [the second to last commit](https://github.com/vikejs/vike/pull/1617/commits/077d865e3ac16f29f0d3b177535690c84927094a)

That seems surprising indeed. Let me dig.

vike/client/client-routing-runtime/renderPageClientSide.ts
```diff
@@ -203,9 +207,18 @@ async function renderPageClientSide(renderArgs: RenderArgs): Promise<void> {
       // Render page view
       await renderPageView(pageContext)
     } else {
-      let res: Awaited<ReturnType<typeof getPageContextFromHooks_isNotHydration>>
+      let res: Awaited<ReturnType<typeof getPageContextFromHooks_isNotHydration>> | PrefetchedPageContext
       try {
-        res = await getPageContextFromHooks_isNotHydration(pageContext, false)
+        const prefetchedPageContext = globalObjectForPrefetchedPageContext?.prefetchedPageContext
+        if (
```
brillout, line 213: We need to check whether it expired.

vike/client/client-routing-runtime/prefetch.ts
```diff
@@ -138,3 +215,24 @@ async function prefetchIfPossible(url: string): Promise<void> {
...
+async function prefetchContextIfPossible(url: string, expire: number | undefined): Promise<void> {
+  const now = Date.now()
+  const lastPrefetch = globalObject?.lastPrefetchTime?.get(url)
+  if (lastPrefetch && expire && now - lastPrefetch < expire) {
+    return
+  }
+  const pageContext = await createPageContext(url)
+  let pageContextFromRoute: PageContextFromRoute
+  try {
+    pageContextFromRoute = await route(pageContext)
```
brillout, line 228: `route()` is being called twice: by `prefetchAssetsIfPossible()` and `prefetchContextIfPossible()`. Can we run it only once instead? It would be nice to avoid calling the user's Route Functions too much.

vike/client/client-routing-runtime/prefetch.ts
```diff
@@ -41,6 +68,42 @@ async function prefetchAssets(pageId: string, pageContext: PageContextUserFiles)
...
+    )
+    objectAssign(pageContext, {
+      isHydration: false as const,
+      isBackwardNavigation: null,
+      _hasPageContextFromServer: false as const,
+      _hasPageContextFromClient: true as const,
+      _pageId: pageId,
+      _pageConfigs: pageContext._pageConfigs
+    })
+    const res = await getPageContextFromHooks_isNotHydration(pageContext, false)
```
brillout, line 100: Actually, how about we use `fetchPageContextFromServer()` instead of `getPageContextFromHooks_isNotHydration()`? I think it would be more precise (avoids calling client-side hooks), and also maybe we can then skip the need to call `loadUserFilesClientSide()`.

vike/client/client-routing-runtime/prefetch/getPrefetchSettings.ts
```diff
@@ -1,29 +1,39 @@
...
-  prefetchStaticAssets: PrefetchStaticAssets
+  prefetchStaticAssets: PrefetchWhen
+  prefetchPageContext: PrefetchPageContext
 }
 
 function getPrefetchSettings(pageContext: PageContextPrefetch, linkTag: HTMLElement): PrefetchSettings {
   let prefetchStaticAssets = getPrefetchStaticAssets(pageContext, linkTag)
-  if (prefetchStaticAssets === 'viewport' && import.meta.env.DEV) {
+  let prefetchPageContext = getPrefetchPageContext(pageContext, linkTag)
+  if (prefetchStaticAssets === 'viewport' && process.env.NODE_ENV === 'development') {
```
brillout, line 24: Why did you change it to `process.env.NODE_ENV`? I'm not sure Vite supports `process.env.NODE_ENV`.

vike/client/client-routing-runtime/prefetch/getPrefetchSettings.ts
```diff
@@ -109,3 +152,44 @@ function getPrefetchAttribute(linkTag: HTMLElement): PrefetchStaticAssets | null
 
   assert(false)
 }
+
+function getPrefetchPageContextAttribute(linkTag: HTMLElement): PrefetchPageContext | null {
+  const whenAttr = linkTag.getAttribute('data-prefetch-page-context-when')
```
brillout, line 157: How about `data-prefetch-page-context` instead of `data-prefetch-page-context-when`?

vike/client/client-routing-runtime/prefetch.ts
```diff
@@ -138,3 +215,24 @@ async function prefetchIfPossible(url: string): Promise<void> {
   if (!(await isClientSideRoutable(pageContextFromRoute._pageId, pageContext))) return
   await prefetchAssets(pageContextFromRoute._pageId, pageContext)
 }
+
+async function prefetchContextIfPossible(url: string, expire: number | undefined): Promise<void> {
+  const now = Date.now()
+  const lastPrefetch = globalObject?.lastPrefetchTime?.get(url)
+  if (lastPrefetch && expire && now - lastPrefetch < expire) {
```
brillout, line 222: - How about some default `expire` value? E.g. 5 seconds or maybe 30 seconds? I'm not sure.
- I wonder whether `expire` should also apply to `prefetch()` :thinking: WDYT?

vike/client/client-routing-runtime/prefetch/getPrefetchSettings.ts
```diff
@@ -109,3 +152,44 @@ function getPrefetchAttribute(linkTag: HTMLElement): PrefetchStaticAssets | null
...
+
+    assertUsage(false, `data-prefetch-page-context has value "${whenAttr}" but it should instead be "false", "hover"`)
+  }
+
+  if (expireAttr !== null) {
+    if (!Number.isNaN(parseInt(expireAttr, 10))) {
+      return { when: 'hover', expire: parseInt(expireAttr, 10) }
+    }
+
+    assertUsage(false, `data-prefetch-page-context-expire has value "${expireAttr}" but it should instead be number`)
```
brillout, line 191: There is some code repetition here. See code above which is a duplicate. This is client-side code so it's important to keep it compact.

vike/client/client-routing-runtime/prefetch/getPrefetchSettings.ts
```diff
@@ -67,7 +77,40 @@ function getPrefetchStaticAssets(pageContext: PageContextPrefetch, linkTag: HTML
...
+function getPrefetchPageContext(pageContext: PageContextPrefetch, linkTag: HTMLElement): PrefetchPageContext {
+  {
+    const prefetchAttribute = getPrefetchPageContextAttribute(linkTag)
+    if (prefetchAttribute !== null) return prefetchAttribute
+  }
+
+  if ('prefetchPageContext' in pageContext.exports) {
+    const { prefetchPageContext } = pageContext.exports
+
+    const wrongUsageMsg = `prefetchPageContext should be an object with 'when' and 'expire' properties. 'when' should be false, 'hover', and 'expire' should be a number`
```
brillout, line 89: Actually, thinking about it: how about we make the setting simply `boolean | number` instead?

brillout:
As for the failing CI, I think we can have a look at it later.

vike/client/client-routing-runtime/prefetch/getPrefetchSettings.ts
usk94: Even though I didn't add `import.meta.env.DEV`, If I had kept that part, the CI test would have failed.
https://github.com/vikejs/vike/actions/runs/9136157470/job/25124574955

brillout: I see. I suggest we have a look at it again after we're happy with the rest. Marking as resolved in the meantime.

vike/client/client-routing-runtime/prefetch.ts
usk94: > I wonder whether expire should also apply to prefetch() 🤔 WDYT?

I think it would be reasonable to exclude the expire logic, as it is expected to be triggered less frequently than hover.

vike/client/client-routing-runtime/renderPageClientSide.ts
brillout: If the user hovered a link 10 minutes ago and now navigates to the url (without hovering), then I think we should not use that 10 minutes old pageContext?

vike/client/client-routing-runtime/prefetch.ts
brillout: Ok 👍

vike/client/client-routing-runtime/renderPageClientSide.ts
usk94: I understand. Thanks👍

vike/client/client-routing-runtime/prefetch.ts
usk94: > How about some default expire value? E.g. 5 seconds or maybe 30 seconds? I'm not sure.

For now, I set it up in 5 seconds.

vike/client/client-routing-runtime/renderPageClientSide.ts
```diff
@@ -203,9 +207,18 @@ async function renderPageClientSide(renderArgs: RenderArgs): Promise<void> {
...
+          prefetchedPageContext?.pageContextFromHooks &&
+          '_pageId' in prefetchedPageContext.pageContextFromHooks &&
+          prefetchedPageContext.pageContextFromHooks._pageId === pageContext._pageId
```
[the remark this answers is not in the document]
usk94: Sorry, I'm not sure I understand your point. May I ask for more details?

usk94:
@brillout 
Thank you for your review!
Except where I responded to your comments, I have updated based on your feedback.
Could you please check it again?

vike/client/client-routing-runtime/renderPageClientSide.ts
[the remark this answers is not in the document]
brillout: I think we can save multiple `pageContext` objects, one per `pageId`. In other words `globalObject.prefetchedPageContext` would be a `Record<string, PrefetchedPageContext>` where `string` is the `pageId`. The idea is that if the user hovers over multiple links, the `pageContext` fetched for previous links can be used instead of being discarded.

vike/client/client-routing-runtime/prefetch.ts
brillout: You're right, it requires a little bit of refactoring. But I think it's worth it: I think it's more accurate/correct to avoid calling client-side hooks upon prefetching. Maybe the solution is to split `getPageContextFromHooks_isNotHydration()` in two parts?

brillout:
> @brillout Thank you for your review! Except where I responded to your comments, I have updated based on your feedback. Could you please check it again?

:+1: I can re-review after we resolved the two pending conversations, but let me know if you want me to re-review before that. Looking forward to it!

vike/client/client-routing-runtime/prefetch.ts
usk94: All right👍
I split `getPageContextFromHooks_isNotHydration()` into two parts.

vike/client/client-routing-runtime/renderPageClientSide.ts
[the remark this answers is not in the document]
usk94: I understood, thank you!
> The idea is that if the user hovers over multiple links, the pageContext fetched for previous links can be used instead of being discarded.

Then, I have implemented this with ` { pageId: string; prefetchedPageContext: PrefetchedPageContext }[]` rather than `Record<string, PrefetchedPageContext>`(because I struggled to fix type errors in `getPageContextFromHooks.ts`)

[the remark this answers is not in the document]
brillout: :+1: We can polish such small details later.

usk94:
@brillout 
I have updated on all your feedback. May I ask for a re-review?

brillout:
@usk94 Neat. I will, tentatively today.

brillout:
https://github.com/vikejs/vike/pull/1617/commits/b0f2579739c279d1ef41f85ebaaecf1c0b5f339b fixes the CI.

One (/ the only?) reason is that we need to use the URL instead of `_pageId` as cache key, because of parameterized routes.

phonzammi:
@brillout, Could the issue be that `lastPrefetch` and`'expire` might be undefined?

brillout:
@phonzammi Maybe, I didn’t dig into that part yet 👀

usk94:
@brillout 
> One (/ the only?) reason is that we need to use the URL instead of _pageId as cache key, because of parameterized routes.

Thanks! CI passes now.

usk94:
And when I switched from `process.env.NODE_ENV` back to `import.meta.env.DEV` at this time, there were no errors successfully.

brillout:
👍 I’ll review again today/tomorrow.

vike/client/client-routing-runtime/prefetch.ts
```diff
@@ -95,17 +170,26 @@ function addLinkPrefetchHandlers(pageContext: { exports: Record<string, unknown>
...
 
-    const { prefetchStaticAssets } = getPrefetchSettings(pageContext, linkTag)
-    if (!prefetchStaticAssets) return
+    const { prefetchStaticAssets, prefetchPageContext } = getPrefetchSettings(pageContextBeforeRenderClient, linkTag)
+    if (!prefetchStaticAssets && !prefetchPageContext) return
+
+    const pageContext = await createPageContext(url)
+    let pageContextFromRoute: PageContextFromRoute
+    try {
+      pageContextFromRoute = await route(pageContext)
```
brillout, line 148: I don't think we should route here: I'm inclined to call the users's Route Functions not eagerly and only when strictly needed.

We can share the routing result by merging the event listeners.

vike/client/client-routing-runtime/prefetch.ts
```diff
@@ -115,26 +199,50 @@ function addLinkPrefetchHandlers(pageContext: { exports: Record<string, unknown>
...
+  console.log('globalobject', globalObject.lastPrefetchTime)
+  console.log('lastprefetch', lastPrefetch)
+  if (lastPrefetch && expire && now - lastPrefetch < expire) {
     return
   }
-  if (!pageContextFromRoute?._pageId) return
-  if (!(await isClientSideRoutable(pageContextFromRoute._pageId, pageContext))) return
-  await prefetchAssets(pageContextFromRoute._pageId, pageContext)
+  await prefetchPageContext(pageId, pageContext)
+  globalObject.lastPrefetchTime?.set(pageContext.urlOriginal, now)
```
brillout, line 216: I think it would be better to move that line inside `prefetchPageContext()`?

vike/client/client-routing-runtime/prefetch.ts
```diff
@@ -41,6 +68,19 @@ async function prefetchAssets(pageId: string, pageContext: PageContextUserFiles)
...
 }
 
+async function prefetchPageContext(pageId: string, pageContext: PageContextForPrefetch): Promise<void> {
+  try {
+    objectAssign(pageContext, { _pageId: pageId })
+    const res = await getPageContextFromServerHooks(pageContext, false)
+    const pageContextWithDuplicateKey = globalObject.prefetchedPageContexts.find((pc) => pc.pageId === pageId)
+    if (!pageContextWithDuplicateKey) {
+      globalObject.prefetchedPageContexts.push({ pageId, prefetchedPageContext: res })
+    }
```
brillout, line 78: I think the `else` case is missing?

brillout:
I added some comments. I'll resuming reviewing after the comments are resolved.

vike/client/client-routing-runtime/prefetch.ts
usk94: > We can share the routing result by merging the event listeners.

I'm not sure I understand, does this mean merging `route()` into the logic of `addLinkPrefetchHandlers()`?

usk94:
@brillout 
I fixed except for the point to which I replied. 
However, the CI is down again, could you provide some insight into what might be going wrong?

usk94:
@brillout 
You are right. I'll do some more research.

vike/client/client-routing-runtime/prefetch.ts
brillout: I think we can merge `prefetchContextIfPossible()` and `prefetchAssetsIfPossible()` into one function and call `route()` in there. That way we call `route()` only if necessary / when we need to, and only once.

brillout:
> @brillout You are right. I'll do some more research.

But feel free to let me know if after having digged you think it's unrelated to your changes.

usk94:
memo: 
- A part of `objectAssign` was restored so that all but `Unit tests E2E` could pass.
- It fails because `timestamp2` and `timestamp6` in `test/misc/testRun.ts#history.pushState()`

usk94:
@brillout 
Sorry, it seems that my changes were the cause. 
Adding lastPrefetchTime as an item in prefetchedPageContexts is causing the CI(Unit tests E2E) to fail. 

I made a slight adjustments to handle whether `lastPrefetchedPageContext.lastPrefetchTime`(the result of find on prefetchedPageContexts) is undefined or not, but the CI is still failing.
I'm not sure what is preventing the success of E2E testing, and I apologize for that, but would it be okay to leave this part as it is for now?

vike/client/client-routing-runtime/prefetch.ts
usk94: I might not fully understand your point yet. `The route()` is not called every time a hover event occurs, but only when the page is loaded. 
Are you suggesting that you want to reduce the number of times route() is called? 
If we merge `prefetchContextIfPossible()` and `prefetchAssetsIfPossible()` into a single function and include the `route()` function, it would result in calling it every time a hover event occurs, which seems to be the scenario you are concerned about.

brillout: > `The route()` is not called every time a hover event occurs, but only when the page is loaded.

That's probablematic in dev, as it's essentially a cache that breaks HMR.

> it would result in calling it every time a hover event occurs

I think it's fine, that's how it was working before this PR and no one complained.

> I might not fully understand your point yet.

It's just about calling `route()` one time instead of two times per hover.

brillout:
There are also other tests that are failing, have you looked at these ones? The `history.pushState()` test is a fairly complex test; instead you may want to have a look at why the boilerplate test is failing which is a much easier test file.

Keep me updated. I can also have a look at why the CI is failing, just let me know.

usk94:
@brillout
> There are also other tests that are failing, have you looked at these ones?

Yes, and with this commit, all tests except for `Unit tests E2E` are passing.
[ad621e4](https://github.com/vikejs/vike/pull/1617/commits/ad621e4d19a6a0fdf1ec63eedb22b8206679ecd0)

> I made a slight adjustments to handle whether lastPrefetchedPageContext.lastPrefetchTime(the result of find on prefetchedPageContexts) is undefined or not

Therefore, I attempted this to pass the remaining the `history.pushState()` test, but it didn't make any significant difference, so I reverted the problematic commit for now.

vike/client/client-routing-runtime/prefetch.ts
usk94: > That's probablematic in dev, as it's essentially a cache that breaks HMR.

I understand now that it's not correct for `route()` to be called only when the page is loaded. Thank you for the clarification.

> I think we can merge prefetchContextIfPossible() and prefetchAssetsIfPossible() into one function and call route() in there.

I've implemented it now.

brillout:
I re-applied that commit and the CI is still green. Anyways, I reverted it nevertheless because I don't think we need it.

Let me do another round of reviewing.

brillout:
@usk94 I found the culprit: https://github.com/vikejs/vike/pull/1617/commits/bb6aac33477dc70fe1699a626f67a30f851bd17b.

usk94:
@brillout 
Oh, I was using _pageId there as well... Thank you for finding that, and I'm sorry for the trouble.

brillout:
> I'm sorry for the trouble.

No worries. Although, FYI, I always double, triple, quadruple check every line of code I make before pushing changes. E.g. I've re-read the changes made to `renderPageClient.ts` and `getPrefetchSettings.ts` like 10 times by now.

In my experience, accuracy is a must-have for OSS tools. And the sooner I focus on accuracy the faster I develop in the long haul.

Now, of course, it isn't possible for you to be accurate around code that you didn't author. That said, I think there is potential to increase accuracy about the code you author.

I've added a couple of `TODO` comments which I'll implement after we merge this (in order to keep this PR minimal).

usk94:
@brillout 
Thank you for the feedback. I acknowledge that I significantly lacked the careful and professional approach you mentioned.

If I have the opportunity to contribute in the future, I'll make sure to double-check my code more thoroughly. I appreciate your guidance and support.

brillout:
No issues, we're all learning :slightly_smiling_face:

brillout:
@usk94 I'll finish this a little later once I'm done with the new head management design. I hope it's ok! :slightly_smiling_face:

usk94:
@brillout 
That's perfectly fine! Thank you for letting me know.

brillout:
@usk94 Thank you for starting this! Apologies for having wrongly labeled this as https://github.com/vikejs/vike/labels/easy-contribution%20%F0%9F%8C%B1 Let me know if you're up for contributing some more and I'll update the https://github.com/vikejs/vike/labels/easy-contribution%20%F0%9F%8C%B1 list while double, triple, quadruple checking before I label something :grinning:

PR #1620, opened 2024-04-25
Update nextjs comparison

redbar0n:
Add SSG section, clarify SPA-only mode and add links, reorder some content, and expand some subtitles.


brillout:
:100:

PR #1621, opened 2024-04-26
Update nextjs comparison some more

redbar0n:
Clarify some headlines, add a link in the table, and move UI framework section down and improve it a bit.


brillout:
Nice. Thank you :green_heart:

PR #1624, opened 2024-04-30
feat: add cookie to pageContext for dev mode

mohamadkhani:
In many cases we need to access cookie in SSR codes. You described correctly how we can pass it to `pageContext` in `server/index.ts` file. 
But it is just effective for production mode. I realized in dev mode, passing header items using Vike Vite plugin was applied for 'user-agent' only, but not for cookie.
I passed `cookie` to `pageContext` like `user-agent` in Vike plugin.

### DIFF
--- a/vike/node/plugin/shared/addSsrMiddleware.ts
+++ b/vike/node/plugin/shared/addSsrMiddleware.ts
@@ -13,7 +13,8 @@ function addSsrMiddleware(middlewares: ConnectServer) {
     const userAgent = headers['user-agent']
     const pageContextInit = {
       urlOriginal: url,
-      userAgent
+      userAgent,
+      headers
     }
     let pageContext: Awaited<ReturnType<typeof renderPage>>
     try {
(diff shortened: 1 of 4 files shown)


brillout:
How about we add the entire `headers` object?

mohamadkhani:
Great :+1:  
I will add it to the next commit in the pull request.

mohamadkhani:
Since current developers are using `userAgent` property we cannot remove it. So it can mark as deprecate if have a plan to remove it. But I'll keep this in the commit.

mohamadkhani:
By the way, I guess passing the entire header object might make it harder to specify which header items should be passed to the client.

brillout:
> By the way, I guess passing the entire header object might make it harder to specify which header items should be passed to the client.

Indeed, do you think that's an issue?

I just squased your commits and forced-pushed.

brillout:
Pre-released as `0.4.171-commit-2e4e1ce`.

mohamadkhani:
Thank you. 
Great project, awesome developer :100:

brillout:
I'm glad you're enjoying Vike. In case that's something your company would be up for you, Vike is looking for [sponsors](#1350).

PR #1632, opened 2024-05-02
missing trailing slash in page context url

### LINKED ISSUE #1629: `urlParsed` removes trailing slash
### Description

When the `urlOriginal` is `/en/index.pageContext.json?asd=asd` and in the address bar is displayed `/en/?asd=asd` the `urlPathname` is `/en`.  When the `urlOriginal` is `/en/?asd=asd` the `urlPathname` is `/en/`.

This behavior is not influenced by `disableUrlNormalization: true`.

In my opinion the `urlPathname` should be `/en/` when the `urlOriginal` is `/en/index.pageContext.json?asd=asd`.

openscript:
fixes #1629

### DIFF
--- a/vike/shared/getPageContextRequestUrl.ts
+++ b/vike/shared/getPageContextRequestUrl.ts
@@ -11,6 +11,6 @@ const doNotCreateExtraDirectory = false
 
 // See node/renderPage/handlePageContextRequestUrl.ts
 function getPageContextRequestUrl(url: string): string {
-  let pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory, true)
+  let pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
   return pageContextRequestUrl
 }
--- a/vike/shared/getPageContextRequestUrl.ts
+++ b/vike/shared/getPageContextRequestUrl.ts
@@ -11,6 +11,6 @@ const doNotCreateExtraDirectory = false
 
 // See node/renderPage/handlePageContextRequestUrl.ts
 function getPageContextRequestUrl(url: string): string {
-  let pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
+  const pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
   return pageContextRequestUrl
 }
--- a/vike/shared/getPageContextRequestUrl.ts
+++ b/vike/shared/getPageContextRequestUrl.ts
@@ -11,6 +11,7 @@ const doNotCreateExtraDirectory = false
 
 // See node/renderPage/handlePageContextRequestUrl.ts
 function getPageContextRequestUrl(url: string): string {
-  const pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
+  let pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
+  if(url.endsWith('/')) pageContextRequestUrl += '/'
   return pageContextRequestUrl
 }
--- a/vike/shared/getPageContextRequestUrl.ts
+++ b/vike/shared/getPageContextRequestUrl.ts
@@ -11,7 +11,6 @@ const doNotCreateExtraDirectory = false
 
 // See node/renderPage/handlePageContextRequestUrl.ts
 function getPageContextRequestUrl(url: string): string {
-  let pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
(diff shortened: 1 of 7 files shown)


brillout:
What's the issue? Can you rebase on top of `git fetch; git rebase -i
origin/main`?

vike/shared/getPageContextRequestUrl.ts
```diff
@@ -11,6 +11,7 @@ const doNotCreateExtraDirectory = false
 
 // See node/renderPage/handlePageContextRequestUrl.ts
 function getPageContextRequestUrl(url: string): string {
-  const pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
+  let pageContextRequestUrl = urlToFile(url, pageContextJsonFileExtension, doNotCreateExtraDirectory)
+  if(url.endsWith('/')) pageContextRequestUrl += '/'
```
openscript, line 15: I'm not sure if this is the right way to do it. Other parts depend on `getPageContextRequestUrl()` and will fail now, because requests to `/test/index.pageContext.json/` cannot be resolved. How to proceed with this?

brillout: See the modules that import `export { pageContextJsonFileExtension }`. (Feel free to let me know if this is getting too complex; I can take over the PR.)

openscript: If you have the patience with me to iterate a few times, until I got it right, I'm happy to work on it. So I can get to know Vike better, contribute a little bit back and learn a bunch of things.

brillout: If it's fun and/or interesting for you, and you don't mind minimal answers from my side, then let's continue!

openscript: Note: Adding a query param like `trailingSlash` is less intrusive, but has the downside that `trailingSlash` becomes a reserved param in Vike.

brillout: Or maybe `/en//index.pageContext.json` after all.

openscript: Note: Having a trailing slash `/en/index.pageContext.json/` doesn't really work out. To reach the goal the request must be made to `/en/index.pageContext.json/`, that not resolves to a file without changing the webserver routing, whereas `/en//index.pageContext.json` just works. Even though this is a quirk, because [RFC2396](www.ietf.org/rfc/rfc2396.txt) defines path separator as a *single* slash.

openscript: > Or maybe /en//index.pageContext.json after all.

I will follow that path and then we can have a look at the implementation.

brillout: Good to know :+1: 

My concern about `pageContext.json?trailingSlash` is that the URL doesn't end with `.json` anymore. I don't know which of the two is the lesser evil :thinking:

An alternative would be to add a property `pageContext._urlHasTrailingSlash` that is included in `pageContext.json` (i.e. in the default `passToClient` values see `PASS_TO_CLIENT`). That would be the cleanest I guess but maybe a little bit more cumbersome to implement.

openscript:
So far:

- `/en/index.pageContext.json/` breaks webserver routing
- `/en//index.pageContext.json` write to file paths with double slash breaks

> An alternative would be to add a property pageContext._urlHasTrailingSlash that is included in pageContext.json (i.e. in the default passToClient values see PASS_TO_CLIENT). That would be the cleanest I guess but maybe a little bit more cumbersome to implement.

I will follow that lead now.

brillout:
@openscript Makes sense :+1:

openscript:
Where would be a good place to update the `url*` contexts after the pageContext was fetched?

brillout:
Actually, what we need is the other way around: we need the client to tell the server that the URL has a trailing slash. So I'm not sure `pageContext._urlHasTrailingSlash` would be of any help here.

brillout:
How about this then:

- `/en` => `/en/index.pageContext.json`
- `/en/` => `/en/index.slash.pageContext.json`

brillout:
We can then use `pageContext._urlHasTrailingSlash` on the server-side only (no need to add it to `passToClient`).

openscript:
Is my thinking about the solution wrong? `pageContext._urlHasTrailingSlash` would be saved in the `pageContext.json` file. When the client fetches, it reads `pageContext._urlHasTrailingSlash` from the fetched `pageContext.json`, then it adjusts the `urlOriginal` etc.

Maybe it's better if you take over this issue. I feel a bit lost with it.

brillout:
The client-side is the source of truth. So the bottom line here is that the information needs to be communicated from the client to the server.

I thought more about it and I'm afraid there doesn't seem to a solution. The issue is that pre-rendering requires the `pageContext.json` file to be written at a unique path and, thus, served at a unique URL. So there isn't a way to determine whether there is a trailing a slash. I went a head and closed, but let me know if you know a framework that supports pre-rendering and trailing slashes and that doesn't have that issue, maybe I missed a solution.

PR #1634, opened 2024-05-08
fix: improve addSsrMiddleware() condition

brillout:
@phonzammi This doesn't seem to work, see CI result.


phonzammi:
I think it's failing because when we use Vite without a server framework (e.g., Express), then `server.httpServer` will not be `null`.

brillout:
I see. Closing but we can re-open if we find another solution.

brillout:
FYI, after Dani and myself are done with the Vike CLI and the server transpilation PRs, we'll have a lot more opportunities to workaround the issue.

PR #1645, opened 2024-05-15
Add `nonce` property to script tags

### LINKED ISSUE #1554: CSP nonce support
### Description

It would be good if the `renderPage` was able to apply CSP nonce to the relevent tags. We are using helmet with express in our application, and i have managed to get it to work by doing something linke this

```JS
const pageContext = await renderPage(pageContextInit);
const { httpResponse } = pageContext;

if (!httpResponse) {
    return next();
  } else {
    const { body, statusCode, headers, earlyHints } = httpResponse;

    if (res.writeEarlyHints) {
      const links: string[] = earlyHints.map((e) => e.earlyHintLink);
      res.writeEarlyHints({ link: links });
    }

    headers.forEach(([name, value]) => res.setHeader(name, value));

    res.status(statusCode);
    res.send(applyCSPToHTML(body, res.locals.cspNonce));
  }
  ```

```js
import { parse, serialize } from "parse5";

export const applyCSPToHTML = (html: string, nonce: string): string => {
  const document = parse(html);

  // Function to apply nonce attribute to script tags
  const applyNonceToScripts = (node: any) => {
    if (node.tagName === "script") {
      const nonceAttribute = node.attrs.find(
        (attr: any) => attr.name === "nonce",
      );
      if (!nonceAttribute) {
        node.attrs.push({ name: "nonce", value: nonce });
      }
    }
    if (node.childNodes) {
      node.childNodes.forEach(applyNonceToScripts);
    }
  };

  // Apply nonce attribute to script tags
  applyNonceToScripts(document);

  // Serialize the modified document back to HTML
  const modifiedHTML = serialize(document);
  return modifiedHTML;
};
```

Im aware, that I would be able to include a meta tag in the HTML easy enough, but applying a nonce to tags generated by Vike is a little more awkward currently.

branberry:
Resolves #1554

### DIFF
--- a/vike/node/runtime/html/injectAssets.ts
+++ b/vike/node/runtime/html/injectAssets.ts
@@ -26,7 +26,6 @@ type PageContextInjectAssets = {
   _baseServer: string
   _pageConfigs: PageConfigRuntime[]
   is404: null | boolean
-  nonce?: boolean
 } & PageContextSerialization
 
 async function injectHtmlTagsToString(
--- a/vike/node/runtime/html/injectAssets.ts
+++ b/vike/node/runtime/html/injectAssets.ts
@@ -26,6 +26,7 @@ type PageContextInjectAssets = {
   _baseServer: string
   _pageConfigs: PageConfigRuntime[]
   is404: null | boolean
+  nonce?: boolean
 } & PageContextSerialization
 
 async function injectHtmlTagsToString(
--- a/vike/node/runtime/html/injectAssets/getHtmlTags.ts
+++ b/vike/node/runtime/html/injectAssets/getHtmlTags.ts
@@ -190,7 +190,7 @@ async function mergeScriptEntries(pageAssets: PageAsset[]): Promise<null | strin
 function getPageContextJsonScriptTag(pageContext: PageContextSerialization, nonce?: string): string {
   const pageContextSerialized = sanitizeJson(serializePageContextClientSide(pageContext))
 
-  const htmlTag = `<script id="vike_pageContext" type="application/json" ${nonce && `nonce="${nonce}"`}>${pageContextSerialized}</script>`
+  const htmlTag = `<script id="vike_pageContext" type="application/json" ${nonce ? `nonce="${nonce}"` : undefined}>${pageContextSerialized}</script>`
 
   // Used by contra.com https://github.com/gajus
   // @ts-expect-error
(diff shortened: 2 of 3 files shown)


brillout:
:+1: 

- The `crypto` module should only be loaded if needed, and by using [`@brillout/import`](https://github.com/brillout/import) to avoid `crypto` from being bundled by bundlers.
- What's wrong with using `Math.random()` instead? I wonder how `Math.random()` can be a security hazard.

branberry:
Thanks for the feedback! I'll update the import to not include `crypto` as a part of the bundle. Based on the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#examples), that the `nonce` value needs to be cryptographically secure:

```
Allowing all inline scripts is considered a security risk, so it's recommended to use a nonce-source or a hash-source instead. To allow inline scripts and styles with a nonce-source, you need to generate a random nonce value (using a cryptographically secure random token generator) and include it in the policy. It is important to note, this nonce value needs to be dynamically generated as it has to be unique for each HTTP request:
```

Based on the [MDN docs for Math.random](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random), it does not generate cryptographically secure values.

brillout:
- :+1: Good to know, how about adding a comment about this in the code?
 - I think we need to add the `nonce` to all other `<script>` tags?
 - How about we set the generated random `nonce` value to `pageContext.nonce`? Is there a use case for that? E.g. if the user wants to inject a script to the HTML?
 - Maybe we can also let the user set a `pageContext.none` value (Vike wouldn't use `crypto` then but take that user-defined value instead), but I don't think there is a use case for that?
 - How about we load the `crypto` module only if *strictly* necessary , e.g. only if`pageContext.nonce` is set?
 - Up for adding docs? Maybe creating a new page `vike.dev/CSP`?

branberry:
Thank you for the feedback! I don't have answers for you at the moment about whether or not there is value in generating a nonce by default. I'll also work on adding the nonce to the other script tags as well.

umjames:
What about nonce attributes on style and link tags?

brillout:
@branberry Are you still up for this? Let's re-open if you are (or someone else). Looking forward to it!

JustFly1984:
@brillout  For SSG case I have to provide a dummy nonce value, which gets replaced in runtime by coloudflare csp worker. It uses node:crypto under the hood. The framework itself should not generate nonce if it is SSG rendered code, cos in this case it will  be random on each build, and it will be impossible to replace it in html.

this is example of cloudflare worker: https://github.com/moveyourdigital/cloudflare-worker-csp-nonce/blob/master/nonce-worker.js

brillout:
Also: https://github.com/vikejs/vike/discussions/1919#discussioncomment-11042864

PR #1669, opened 2024-06-03
Apply Vite config server headers in dev and preview

IoriBranford:
The only advice I could find on adding response headers was #697. Pulling from Vite config will be more intuitive and developer-friendly and not require writing your own server.

Any other ignored Vite server settings can also be applied in this way.


brillout:
Neat PR! 

> Any other ignored Vite server settings can also be applied in this way.

Further PRs welcome :green_heart:

brillout:
Pre-released as `0.4.172-commit-605b4a5`.

PR #1678, opened 2024-06-07
run transformIndexHtml in prerender

nitedani:
I'm not sure we should do it like this, what do you think?
Maybe it's not ok to call plugin.transformIndexHtml outside of vite.. But if the vite build and prerender runs in the same process, maybe we can bind the function to the vite instance.

### DIFF
--- a/examples/react-full/vite.config.ts
+++ b/examples/react-full/vite.config.ts
@@ -9,6 +9,12 @@ export default {
       prerender: true
     }),
     mdx(),
-    react()
+    react(),
+    {
+      name: 'html-minify',
+      transformIndexHtml(html) {
+        return html.replaceAll('\n', '')
+      }
+    }
   ]
 } as UserConfig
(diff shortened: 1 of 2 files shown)


brillout:
I actually wonder: what's the official way of applying `transformIndexHtml()` at bulid-time? How do other frameworks do it? Little utility to grep into all Vite-based frameworks: https://github.com/brillout/vite-frameworks.

Btw. let's move the test from `/examples/react-full/` to `/test/misc/`: I'm trying to minimize the amount of E2E tests and I'm using `/test/misc/` to test all kinds of things.

nitedani:
I've checked and they don't seem to support this. They don't call transformIndexHtml at all.
To make it work, there is a custom implementation for vite-pwa for every supported framework:
https://github.com/vite-pwa

brillout:
Does it work with vite-plugin-pwa? If it does then I think it's worth it to do that hack and continue working on this PR.

> So, maybe Vike should have one too?

Yes, we could eventually have an extension `vike-pwa` that modifies the HTML. We can create a new Vike built-in hook for it. But let's do it in a separate PR and later as I don't think it's a priority for now (especially if we do end up merging this PR).

brillout:
With `+middleware` and the ability for Vike extensions to add Vite plugins, we can create `vike-pwa` instead of doing this. Let's close in favor of creating `vike-pwa`.

PR #1679, opened 2024-06-08
feat: on boot draft

OlaviSau:


vike/client/client-routing-runtime/createPageContext.ts
```diff
@@ -35,6 +37,18 @@ async function createPageContext(urlOriginal: string) {
     _pageRoutes: pageRoutes,
     _onBeforeRouteHook: onBeforeRouteHook
   }
+
+  if (!globalObject.onBootResult && onBootHook) {
+    globalObject.onBootResult = await onBootHook.hookFn(pageContext) as Record<string, unknown>
+  }
+
+  if (isObject(globalObject.onBootResult)) {
+    pageContext = {
```
brillout, line 46: I think we can use `Object.assign()` instead of re-creating a new object.

vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/configDefinitionsBuiltIn.ts
```diff
@@ -206,6 +207,9 @@ const configDefinitionsBuiltInGlobal: Record<ConfigNameGlobal, ConfigDefinitionI
   onBeforeRoute: {
     env: { server: true, client: 'if-client-routing', eager: true }
   },
+  onBoot: {
+    env: { server: true, client: true, eager: true }
```
brillout, line 211: ~~I'm not sure, but I don't think we need `eager`.~~

Actually, we need `eager` if we make `onBoot()` non-global and cumulative (see my sibling comment).

vike/shared/page-configs/Config.ts
```diff
@@ -131,6 +133,18 @@ type OnBeforeRenderAsync = (
  *  https://vike.dev/onBeforeRender
  */
 type OnBeforeRenderSync = (pageContext: PageContextServer) => { pageContext: Partial<Vike.PageContext> } | void
+/** Hook called when Vike is first initialized.
+ *
+ *  https://vike.dev/onBoot
+ */
+type OnBoot = (
```
brillout, line 140: Let's name it `OnBootAsync` to keep it consistent with the other type names.

vike/shared/route/loadPageRoutes.ts
```diff
@@ -175,12 +175,14 @@ function getGlobalHooks(
...
+  onBootHook: null | Hook
   filesystemRoots: null | FilesystemRoot[]
 } {
   // V1 Design
   if (pageConfigs.length > 0) {
-    const hook = getHookFromPageConfigGlobal(pageConfigGlobal, 'onBeforeRoute')
-    return { onBeforeRouteHook: hook, filesystemRoots: null }
+    const onBootHook = getHookFromPageConfigGlobal(pageConfigGlobal, 'onBoot')
+    const onBeforeRouteHook = getHookFromPageConfigGlobal(pageConfigGlobal, 'onBeforeRoute')
+    return { onBootHook, onBeforeRouteHook, filesystemRoots: null }
```
brillout, line 185: It's a good question whether `onBoot()` should be global.

Maybe we can make `onBoot()` non-global and cumulative. In other words: the result of the various `onBoot()` hooks are cached and only the results of the relevant `onBoot()` hooks are applied each time Vike creates a new `pageContext` object.

~~But this means that `onBoot()` can be called later upon navigating to a new page and `onBoot()` is a misnomer then.~~ Ah, no, actually if make `onBoot()` eager then we can call *all* `onBoot()` hooks at boot time.

vike/node/runtime/renderPage.ts
```diff
@@ -149,6 +149,14 @@ async function renderPageAndPrepare(
     // From now on, renderContext.pageConfigs contains all the configuration data; getVikeConfig() isn't called anymore for this request
   }
 
+  if (renderContext.onBootHook) {
+    const onBootResult = await renderContext.onBootHook?.hookFn(pageContextInit) as Record<string, unknown>
```
brillout, line 153: I don't think we should call `onBoot()` at request-time? My thinking was to call `onBoot()` exactly once and as soon as possible whenever its code is loaded and then cache its result.

brillout:
Hi Olavi, thank you for opening this PR! It goes in the right direction.

I've left a couple of comments.

I'm a bit busy with two `highest-priority` tickets (`pageContext` prefetching and implementing the new head management design for `vike-{react,vue,solid}`), so I may be a bit slow with replying.

Can't wait to land `onBoot()` hooks. I've actually beeing thinking of bumping its priority to `highest-priority`.

vike/shared/route/loadPageRoutes.ts
brillout: I'm actually leaning more towards making `onBoot()` global as you suggested.

The thing is that, if we don't make it global, then we have to resolve routing before calling `onBoot()`, which is somewhat contradictory as an idea of `onBoot()` is to call it as early as possible and before any request is made to the server.

If a user wants a per-page hook, then he can use `onBeforeRender()`. (So far I don't think we need to introduce a second hook `onRequest()`.)

WDYT?

brillout:
Done: http://vike.dev/onCreateGlobalContext.

Thank you @OlaviSau for your initial work. Door is definitely open for more contributions. Kind regards, Rom.

PR #1688, opened 2024-06-11
Add symlinks support

iserdmi:
Also there is example, which works fine with symlinks with this PR changes: https://github.com/iserdmi/vike-link-example/tree/master

Symlinks should be relative and also not be gitignored. Because if them gitignored it is not possible to get their mode with `$ git ls-file`

### DIFF
--- a/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/crawlPlusFiles.ts
+++ b/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/crawlPlusFiles.ts
@@ -179,8 +179,10 @@ async function gitLsFiles(
     // ```
     // 100644 f6928073402b241b468b199893ff6f4aed0b7195 0 pages/index/+Page.tsx
     // ```
-    const [mode, _, __, filePath, ...rest] = resultLine.split(/\s/)
-    assert(mode && filePath && rest.length === 0)
+    const [info, filePath, ...rest1] = resultLine.split(/\t/)
+    assert(info && filePath && rest1.length === 0)
+    const [mode, _, __, ...rest2] = info.split(' ')
+    assert(mode && _ && __ && rest2.length === 0)
 
     // Deleted?
     if (filesDeleted.includes(filePath)) continue
--- a/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/crawlPlusFiles.ts
+++ b/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/crawlPlusFiles.ts
@@ -50,15 +50,15 @@ async function crawlPlusFiles(
 
   // Crawl
   let files: string[] = []
-  const symlinksExists = await isSymlinksExists(userRootDir, outDirRelativeFromUserRootDir)
-  const res =
-    crawlWithGit !== false && !symlinksExists && (await gitLsFiles(userRootDir, outDirRelativeFromUserRootDir))
+  const res = crawlWithGit !== false && (await gitLsFiles(userRootDir, outDirRelativeFromUserRootDir))
   if (
     res &&
+    // We cannot find files inside symlinks with `git ls-files` because it doesn't follow symlinks
+    res.symlinks.length === 0 &&
     // Fallback to fast-glob for users that dynamically generate plus files. (Assuming all (generetad) plus files to be skipped because users usually included them in `.gitignore`.)
-    res.length > 0
+    res.files.length > 0
   ) {
-    files = res
+    files = res.files
   } else {
     files = await fastGlob(userRootDir, outDirRelativeFromUserRootDir)
   }
@@ -98,7 +98,13 @@ async function crawlPlusFiles(
 }
 
 // Same as fastGlob() but using `$ git ls-files`
-async function gitLsFiles(userRootDir: string, outDirRelativeFromUserRootDir: string | null): Promise<string[] | null> {
+async function gitLsFiles(
+  userRootDir: string,
+  outDirRelativeFromUserRootDir: string | null
+): Promise<{
+  files: string[]
+  symlinks: string[]
+} | null> {
   if (gitIsNotUsable) return null
 
   // Preserve UTF-8 file paths.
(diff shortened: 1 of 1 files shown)


brillout:
So there is no way to use `git ls-files` to get the content of symlink directories?

The PR makes the crawling 2x slower, let's re-use the already existing `git ls-files` command instead.

iserdmi:
> So there is no way to use `git ls-files` to get the content of symlink directories?
I can not find a way to do it

> Let's re-use the already existing git ls-files command instead.
Ok

iserdmi:
@brillout I have reused git ls-files command output. But now it worse. Previously there was filter by "+*{ext}", now I can not use this filter, becouse it does not match symlink dirs, which has not "+" or any extension

I think it will be allow users by config choose if they want to use fast-glob or git ls-files

brillout:
> we cannot use this pattern, becouse it cannot found symlink dirs

That's actually a bummer. How big is the performance loss when skipping that?

iserdmi:
@brillout I do not know how to check it. I think it is suitable to each project. Sometimes you have a lot of files, sometimes not. Sometimes a lot of them with "+", sometimes no.

Or you have perfomance tests somewhere inside packages?

brillout:
I actually suspect that Git uses some kind of cache and that filtering doesn't add any noticeable performance gain.

> Or you have perfomance tests somewhere inside packages?

How about https://github.com/torvalds/linux or some other big project. Let's see how `$ git ls-files` performance there with and without filtering.

iserdmi:
@brillout  Filters make command slower, not faster. But I do not know if it is faster to filter with git ls-files or with our js code.

Here is measures from vike package
```
time git ls-files "**/*.md" > /dev/null
0.01s user 0.01s system 44% cpu 0.032 total
```
```
time git ls-files > /dev/null
0.01s user 0.00s system 62% cpu 0.019 total
```

Linuxs too large, it will be cloned 20 hours I think.

iserdmi:
I have added some optimizations. Now we do not search again with fast-glob if symlinks found. We just search with fastglob inside symlinks, other results will be got from git ls-files as usual

```
  // Crawl
  let files: string[] = []
  const res = crawlWithGit !== false && (await gitLsFiles(userRootDir, outDirRelativeFromUserRootDir))
  if (!res || res.files.length === 0) {
    // Fallback to fast-glob for users that dynamically generate plus files. (Assuming all (generetad) plus files to be skipped because users usually included them in `.gitignore`.)
    files = await fastGlob(userRootDir, outDirRelativeFromUserRootDir)
  } else if (res.symlinks.length > 0) {
    // We cannot find files inside symlinks with `git ls-files` because it doesn't follow symlinks
    const filesInSymlinks = (
      await Promise.all(
        res.symlinks.map(async (symlink) => {
          return (await fastGlob(path.posix.join(userRootDir, symlink), outDirRelativeFromUserRootDir)).map((file) =>
            path.posix.join(symlink, file)
          )
        })
      )
    ).flatMap((filesInSymlink) => filesInSymlink)
    files = res.files.concat(...filesInSymlinks)
  } else {
    // Use result of `git ls-files` as is, becouse no symlinks found
    files = res.files
  }
  ```

**Comment by iserdmi** [comment]
@brillout How do you think, should I fix something else?

**Comment by iserdmi** [comment]
I have added support for symlinks that is not a dir, but a file. Also add them to example: https://github.com/iserdmi/vike-link-example/tree/master

**Comment by iserdmi** [comment]
Measures for linux  https://github.com/torvalds/linux 
```
time git ls-files "**/+*.js" "**/+*.ts" "**/+*.cjs" "**/+*.cts" "**/+*.mjs" "**/+*.mts" "**/+*.jsx" "**/+*.tsx" "**/+*.cjsx" "**/+*.ctsx" "**/+*.mjsx" "**/+*.mtsx" "**/+*.vue" "**/+*.svelte" "**/+*.marko" "**/+*.md" "**/+*.mdx" > /dev/null
0.22s user 0.01s system 93% cpu 0.249 total
```
```
time git ls-files > /dev/null
0.05s user 0.01s system 79% cpu 0.081 total
```

**Comment by brillout** [comment]
Very neat. Let me have a closer look and review later today. Looking forward to it.

**Comment by brillout** [comment]
I can confirm that using the filter argument of `$ git ls-files` doesn't bring any significant performance gain, but using `--exclude` does make a difference.

The following are tests against [Linux's Git repository](https://github.com/torvalds/linux) (`84940` files, 5GB).

<br/>

## Filter argument

```
time git ls-files --others --exclude-standard --cached --stage | wc
  84956  339827 7577701

real	0m0.114s
user	0m0.062s
sys 	0m0.068s
```

```
time git ls-files --others --exclude-standard --cached --stage "**/*.sh" | wc
  893    3572   88664

real	0m0.094s
user	0m0.033s
sys 	0m0.063s
```

I picked the biggest performance difference afer running the same commands like five times.

<br/>

## `--exclude`

That said, for untracked files, using `--exclude` does make a difference.

```
mkdir untracked && mv * untracked
```

```
time git ls-files --others --exclude-standard | wc
  84940   84943 4178851

real	0m0.238s
user	0m0.127s
sys 	0m0.124s
```

```
time git ls-files --others --exclude-standard --exclude="**o**" | wc
  26143   26143 1113819

real	0m0.097s
user	0m0.047s
sys 	0m0.057s
```

brillout:
@iserdmi It doesn't work when the symlink directory isn't tracked: try with a newly created symlink directory you didn't check into your repository yet and you'll see that `git ls-files --others --exclude-standard --stage` is missing the mode for the symlink directory. Ideas?

iserdmi:
@brillout Hello! I think if mode missing in staged result, we can retrieve it by fs module manually

brillout:
There can be a lot of untracked files, so maybe that would be too slow?

iserdmi:
In general I see next option:
1. Parse mode manually with fast
2. Fallback to fast-glob for all results (forget results of git)
3. Combined. If there is more then N results without mode fallback to fats-glob, else parse mode manually with sd

brillout:
Number 3 sounds good! Up for it?

iserdmi:
@brillout Please, take a look to my last commit. Number 3 is there

brillout:
@iserdmi Did some refactoring and increased the threshold to `5`. Let me know what you think.

iserdmi:
@brillout I think it is ok. Lets do new release? I really want to use it now in my project 🙂

brillout:
Pre-released as `0.4.177-commit-ff3d6cd`.

Thank you for your contributions :green_heart:

iserdmi:
@brillout Thank you too! It is my first contribution to not mine opensource project. I was happy to work together!

brillout:
I'm glad Vike was your first :) It was fun!

brillout:
I'm removing this, because it doesn't work.

See https://github.com/vikejs/vike/pull/2086 and https://github.com/vikejs/vike/pull/2087.

Use the upcoming setting `VIKE_CRAWL={git:false}` instead. (ETA today or next week.)

PR #1703, opened 2024-06-17
improve docs

brillout:


brillout:
As always: feel free to disagree.

PR #1709, opened 2024-06-21
Update why page: architectural freedom headline

redbar0n:
To make the benefit clearer from the headline, for those that simply skim.


brillout:
Funny, I actually had the same idea but reverted. If you also stumbled upon that idea, then I guess we should it. I've went ahead and developed the other headlines as well.

Merging, but feel free to disagree with my changes :eyes:

PR #1711, opened 2024-06-21
Update FAQ: add sponsoring as individual

redbar0n:


brillout:
It's mostly company sponsoring that makes a difference though. I'll be removing that section as soon as the new landing page lands anyways.

Thanks though!

PR #1716, opened 2024-06-26
docs: Add vike-cloudflare

magne4000:


brillout:
:100:

FYI the plan is to enable Vike extensions to be able to add Vite plugins (Dani's `vike-react-zustand` PR needs it). We'll then be able to make `vike-cloudflare` a Vike extension, also for giving it the capability to extend Vike which we'll need for integrating Vite Runtime.

brillout (approved):

brillout:
(LGTM => feel free to merge.)

PR #1736, opened 2024-07-09
chore: Re-add clean script

phonzammi:
@brillout 

I just found out that the `pnpm dev:setup` script is still using the `pnpm run clean` command, but there is no such script or command available.
Here, I've put that script back. 

Feel free to merge or modify this PR as you see fit.


brillout:
Thanks for the hint :broom:

Let me know if you disagree with my commit.

PR #1741, opened 2024-07-11
docs: add `+Loading` docs

nitedani:
I think this is not complete, do you have ideas how to improve it?


brillout:
WDYT? Mergeable?

nitedani:
yes

PR #1740, opened 2024-07-11
fix: support `injectScriptsAt: 'STREAM'`

phonzammi:
Hi @brillout, What do you think of this PR?

PR #1743, opened 2024-07-13
feat: new setting `injectScriptsAt`

phonzammi:
Alternative to #1740 

@brillout, Please take a look and review this PR

### DIFF
--- a/docs/headingsDetached.tsx
+++ b/docs/headingsDetached.tsx
@@ -586,5 +586,9 @@ const headingsDetached = [
   {
     title: '`bodyHtmlEnd`',
     url: '/bodyHtmlEnd'
+  },
+  {
+    title: '`injectScriptsAt`',
+    url: '/injectScriptsAt'
   }
 ] satisfies HeadingDetachedDefinition[]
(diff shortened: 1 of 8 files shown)


brillout:
WDYT? Should we merge?

phonzammi:
> WDYT? Should we merge?

I love it. I just leave one review, if it's fine then LGTM :rocket:

brillout:
> https://github.com/vikejs/vike/commit/6bdbe7cc26cd81b45af22a1d3a6426617a5c4470

How about we allow the user to do a mistake but add a warning instead of blocking him?

phonzammi:
Should we? I don't know, I think I'm fine either way.

brillout:
My concern is that the user wants to change the scripts position (otherwise he wouldn't use the setting in the first place) but we break his intent. I'd rather show a warning instead of blocking his intent.

phonzammi:
Right, it's just that if the user returns a pageContextPromise from onRenderHtml(), then they wouldn't have it.

brillout:
Shall we merge?

brillout:
> https://github.com/vikejs/vike/pull/1743/commits/8001eb19edba1fa3796e2963b85c35cd4c2fdb51

Good catch :ok_hand:

phonzammi:
LGTM :rocket:

brillout:
Merged :rocket:

brillout:
I suggest we give #1740 another try and then release a new version. (The docs will be describing a feature that isn't released yet for a little while, but I think it's ok if we manage to merge/decide on 1740 sooner rather than later.)

phonzammi:
Thank you very much :+1:

PR #1750, opened 2024-07-16
fix: make build deterministic

uifox:
To easily verify this change you can take latest vike and modify `dist/esm/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/crawlPlusFiles.js` as I did in the PR.

### DIFF
--- a/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/crawlPlusFiles.ts
+++ b/vike/node/plugin/plugins/importUserCode/v1-design/getVikeConfig/crawlPlusFiles.ts
@@ -201,7 +201,7 @@ async function fastGlob(userRootDir: string, outDirRelativeFromUserRootDir: stri
     cwd: userRootDir,
     dot: false
   })
-  return files
+  return files?.sort()
 }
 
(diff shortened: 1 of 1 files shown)


brillout:
Why?

uifox:
> Why?

fast-glob does not guarantee the same order of file paths and this is causing a different hash in the filename of `entry-client-routing.${hash}.js`. We run the build several times per pods and even though we build from the same source, we get different hashes. This can result in situations where a browser receives a page that references a client routing file with a hash that does not exist on the pod that will serve the request and the app is broken.

We need deterministic builds and sorting the paths fixes our issue.

brillout:
Pre-released as `0.4.180-commit-648cd01`.

Btw. would your company be up for sponsoring (https://github.com/vikejs/vike/issues/1350)? It has a huge impact for the project.

uifox:
@brillout I truely believe vike deserves sponsorship. Unfortunately I am not the decision maker in this regard but I advocated your case toward management a few times already and I got positive feedback, so I hope in the near future it will take place. Thanks a lot for your strong commitment to vike. And keep up with the great work.

brillout:
Thank you, it's much appreciated :+1:

PR #1758, opened 2024-07-22
docs: minor fix

pelly-ryu:
There's no `export` in the TypeScript example code.


pelly-ryu:
I just realized there's no `export` in the other hooks' examples also.

This may have been intentional. I'll close this PR first. Sorry for interrupting you.

brillout:
I think it makes sense. Other PRs welcome to make the docs more consistent!

PR #1764, opened 2024-07-24
fix: PUBLIC variables now work on windows

pafnuty:
In Windows OS, the word `PUBLIC` is reserved in environment variables.
If I use `PUBLIC_ENV__xx` variables (e.g. in +client.ts), the project build will terminate with an error like this:

```bash
error during build:
[vike:envVars] [vike][Wrong Usage] import.meta.env.PUBLIC is used in client-side file /src/routes/+client.ts which means that the environment variable PUBLIC will be included in client-side bundles and, therefore, PUBLIC will be publicly exposed which can be a security leak. Use import.meta.env.PUBLIC only in server-side files, or rename PUBLIC to PUBLIC_ENV__PUBLIC, see https://vike.dev/env
```

This minor fix eliminates this error.
I did a check for windows only and only exact match with the word `PUBLIC`.

### DIFF
--- a/vike/node/plugin/plugins/envVars.ts
+++ b/vike/node/plugin/plugins/envVars.ts
@@ -22,11 +22,10 @@ const PUBLIC_ENV_WHITELIST = [
 
 function isEnvNameInWhiteList(envName: string): boolean {
   // OS Windows reserved PUBLIC env variable
-  
-  if(process.platform === 'win32' && envName === 'PUBLIC') {
-    return true;
+  if (process.platform === 'win32' && envName === 'PUBLIC') {
+    return true
   }
-  return PUBLIC_ENV_WHITELIST.includes(envName);
+  return PUBLIC_ENV_WHITELIST.includes(envName)
 }
 
 function envVarsPlugin(): Plugin {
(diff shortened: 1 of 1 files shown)


brillout:
I ain't sure I understand the issue this is solving? What's the reason you can't rename the environment variable to something else that starts with `PUBLIC_ENV__`?

pafnuty:
@brillout 
Unfortunately, that's not possible.
This is a system variable, introduced in windows vista

https://gist.github.com/pkfrom/f6eb82316b725a51f357#windows-vista-and-later
by default `PUBLIC` var can be uses as `%PUBLIC%` in explorer and will take the user to the folder `C:\Users\Public`

pafnuty:
The thing is, that's what I'm trying to do with the client-side variable. 
I created a file +client.js, in which I placed the following code:
```js
console.log(`PUBLIC_ENV__TEST: ${import.meta.env.PUBLIC_ENV__TEST}`);
```
dev server startup result:
image

build result (added console.log to envVars.js for clarity):
image

It turns out that vike searches through all environment variables, comes across `PUBLIC` and stops building with an error, while dev works with warnings

brillout:
Fix pre-released as `0.4.181-commit-ee50efa`.

PR #1799, opened 2024-08-11
fix: streamReadableWebToString encoding

nitedani:


vike/node/runtime/html/stream.ts
```diff
@@ -115,15 +115,17 @@ async function streamReadableNodeToString(readableNode: StreamReadableNode): Pro
...
     const { done, value } = await reader.read()
-    if (done) {
-      break
-    }
-    str += value
+    if (done) break
+    str += decoder.decode(value, { stream: true })
   }
+
+  str += decoder.decode()
```
brillout, line 128: Curious: why this line?

nitedani: image

nitedani: The final decoder.decode() is actually only necessary in a very specific scenario:
If the very last chunk of the stream ends with the beginning of a multi-byte character, that incomplete character data will be held by the decoder and not output. The final decoder.decode() call will then output a replacement character (usually �) for this incomplete sequence.

brillout:
Released in `0.4.184`.

PR #1814, opened 2024-08-17
docs: Add effector to documentation 

sergeysova:
Hello.
Adding example to documentation about integrating effector with vike react.

Example added to effector org to simplify support for effector team.


brillout:
Neat!

PR #1821, opened 2024-08-21
fix: web stream encoding

nitedani:


brillout:
Let me finish https://github.com/vikejs/vike/pull/1823 and then I'll release a new version.

PR #1829, opened 2024-08-27
monorepo, vike-node-dev, vike-node

nitedani:


brillout:
WDYT? Should we proceed with this? We can further discuss this if your intuition tells you we shouldn't do this, I ain't strongly opinionated here.

Btw. I was thinking, for users that use Vike as SSG it could be nice to have `vike-node-dev` be included in Vike core. So that these users don't have to install any extensions beyond Vike core and `vike-{react,vue,solid}`. But let's see if we eventually want to do this or not.

It seems like there are a couple of examples missing, is that the reason why you labeled the PR as `draft`?

PR #1859, opened 2024-09-07
fix: rename STREAM to HTML_STREAM

brillout:
@phonzammi I made it a hard breaking change, as I guess `vike-solid` is the only user using `injectScriptsAt: 'STREAM'`. Make to sure to update the `+require` value of `vike-solid` to the Vike version I'm about to release.


phonzammi:
Got it 👍

phonzammi:
WDYT https://github.com/vikejs/vike-solid/pull/117/commits/dc83aeb1619b64baa100606235fe86a13bed1796?

brillout:
That would work, although enforcing 195 only for vike-solid-query would be a little cleaner. But it's fine either way.

PR #1872, opened 2024-09-10
fix: broken English

roschaefer:
Interesting, that I come across this only one day after #1867.

On my machine I can see still see the `Uncaught SyntaxError: missing ) after argument list`.


brillout:
Thank you!

PR #1879, opened 2024-09-16
fix: prevent page reload breaking in Chrome when DevTools is open (fix #1875)

### LINKED ISSUE #1875: Chrome does not reload in dev mode with DevTools open
### Description

Setup:
- vike app in dev mode
- Chrome (tested on linux)
- DevTools open

Test:
- reload the page (click on reload button) once => it should reload
- reload the page (click on reload button) again (or 2 times at most) => it should not reload
- now click in URL bar and press ENTER => it should reload

The problem would be just a nuisance but after reload I can no longer inspect the response in the Network tab of DevTools.

I traced it to the `onPageHide` hook that tries to execute tasks in synchronous mode on reload.
https://github.com/vikejs/vike/blob/93017bedc0e5f88ff40a9b3a9b379c607576fd03/vike/utils/onPageVisibilityChange.ts#L7
A solution would be to call the listener in a `requestAnimationFrame`.

I know it's a Chrome problem :)

pdanpdan:


brillout:
:+1: How about we do this only for Chrome and we add a comment liking to #1875?

pdanpdan:
Can we also detect if devtools is open? :)
I'm afraid it would generate problems because it would work differently in different browsers. 
At least this way if there is a problem introduced it would show everywhere.

brillout:
> At least this way if there is a problem introduced it would show everywhere.

I agree in principle, but I feel like using `requestAnimationFrame()` upon page termination is asking for trouble. So I'm leaning more towards being conservative on that end.

> Can we also detect if devtools is open? :)

Good idea :grin: (if we go down that path).

pdanpdan:
Before going further, the original question: can you also reproduce the problem?

brillout:
> can you also reproduce the problem?

Do you mean the problem you described at https://github.com/vikejs/vike/issues/1875#issue-2520486323? With my Chrome version?

pdanpdan:
Yes

brillout:
I cannot reproduce, using the following:

```bash
git clone git@github.com:brillout/vike-reprod-1875 && cd my-app/ && pnpm install && pnpm dev
```

Then going to http://localhost:3000/, open the browser's dev tools, then reload multiple times. It always reloads successfully without any hanging.

Chrome version: `128.0.6613.119 (Official Build) (64-bit)` on Linux.

pdanpdan:
Then it means it's only a bug in Chrome Beta (129).
To be honest it's not the only broken thing.
So it does not make sense to put a fix in Vike for the moment.
I'll close the PR and we can resurrect it if next stable chrome is still broken.

brillout:
Makes sense, thanks!

PR #1889, opened 2024-09-23
chore: replace fast-glob with tinyglobby

rtritto:
Replace `fast-glob` with [tinyglobby](https://github.com/SuperchupuDev/tinyglobby)


brillout:
Seems like some tests are failing.

brillout:
> Seems like some tests are failing.

It's a test for the old design. Let's close and re-open once the old design is removed. Remind me when that's the case, I may forget about this PR.

Thank you!

PR #1898, opened 2024-09-25
fix: node_modules in .gitignore

rtritto:
~~`/node_modules/` works only in Unix OS.~~
~~`node_modules` works on all OSs.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-/node_modules/
+node_modules


brillout:
Being able to develop Vike on windows is a non-goal.

rtritto:
With `/node_modules/`, the files in `/test/misc/node_modules` aren't ignored.

brillout:
Yea.. so please be careful before creating a PR. Thank you.

PR #1899, opened 2024-09-25
chore: vike path in pnpm-lock .yaml from absolute to relative

rtritto:


brillout:
Curious: is this PR solving a problem?

rtritto:
When I change a dependency, all path links are updated with my absolute path.

Next, `/home/rom/tmp` in [/vike/vike/node/runtime/renderPage/logErrorHint\getErrorHint.spec.ts](https://github.com/vikejs/vike/blob/main/vike/node/runtime/renderPage/logErrorHint/getErrorHint.spec.ts) needs to be fixed.

PR #1928, opened 2024-10-17
feat: onPopState() 

brillout:


brillout:
@dajinchu This PR makes a couple of fixes around how Vike handles `history.pushState()` which I just released as `0.4.200`. Let me know if hit any regression.

PR #1968, opened 2024-11-18
docs: minor

phonzammi:


brillout:
Good catch! Please self merge.

Issue #1943, opened 2024-11-01
Add spellchecker to CI


brillout:
### Description

Maybe we an use the tool that was used at https://github.com/vikejs/vike/pull/1835; IIRC the tool has an npm package.

IanWorley:
I would like to take a shot with this issue. Does not seem to difficult at all.

brillout:
@IanWorley Ok! So far I think the best would be to integrate the spell check similar to how the formatting check is integrated: https://github.com/vikejs/vike/blob/main/.github/workflows/formatting.yml.

IanWorley:
I will submit a draft pr but I still have some issue at this time for right now I have config only detecting typos from mdx files in the docs directory do we want this running on the entire project and what is the scope? Anyway I will also reach out in the discussion of typos as it seems there examples of typos are not being pick up in the mdx file.

brillout:
> Do we want this running on the entire project and what is the scope?

Not for now. (Eventually yes once we don't have any pending open PRs.)

For now we can run against the whole `docs/` directory, include `.tsx` files. (No pending PR is modifying it.)

> I will also reach out in the discussion of typos as it seems there examples of typos are not being pick up in the mdx file.

:+1:

IanWorley:
> @IanWorley 👍 Btw. how big is their npm package? It's fine if it's a bit heavy, but it shouldn't be too heavy.
Sorry I did not mean to close the pr I just drop my comments. To response to this I do not think typos is a npm package. I will look into using cspell at this moment my only problem is that there might be dictionary file if that is fine with you. By the way size of cspell is 203kb

brillout:
I found https://github.com/dalisoft/typos-rs-npm so maybe we can use that? We can also ask the typos team if there is an official npm package and if not whether it's on their radar.

> I will look into using cspell at this moment my only problem is that there might be dictionary file if that is fine with you.

How would that be a problem? Can we implement [this workflow](https://github.com/vikejs/vike/pull/1971#issuecomment-2489159526) with it?

> By the way size of cspell is 203kb

That's totally fine.

IanWorley:
Yeah, at first I was going to use cspell, but the amount of manual dictionary entries seemed tedious at best. That’s why I chose typos-rs-npm instead—it seems far more maintainable than managing a massive custom dictionary with cspell.

cr7pt0gr4ph7:
[...]

PR #1971, opened 2024-11-20
CI/add spell checker 

IanWorley:
#1943


brillout:
The reason I was mentioning using their npm package is because it would allow us to correct spelling mistakes by simply running an npm script.

That's what we're doing with formatting:

- `pnpm run format:check` => throws an error if formatting is incorrect, used by the CI
- `pnpm run format` => fixes formatting issues, used by the user

I wonder if we can reproduce the same workflow for spelling?

IanWorley:
@brillout my bad I will use the npm package

brillout:
@IanWorley :+1: Btw. how big is their npm package? It's fine if it's a bit heavy, but it shouldn't be too heavy.

PR #1981, opened 2024-11-23
make getCacheControl return defaultValue on 5xx server errors

nonamich:
Some cdn providers cache 5xx responses, so it's better to avoid it


brillout:
Pre-released as `0.4.204-commit-7f41ba2`.

PR #1996, opened 2024-12-03
docs: replace deprecated Effect type with ConfigEffect

cr7pt0gr4ph7:
The `Effect` type is deprecated according to the code comments, so the examples in the docs should be updated accordingly.

https://github.com/vikejs/vike/blob/4dcddbc6b875a9da196747e939b9c8a6de70f5a0/vike/types/index.ts#L67-L74


brillout:
:pray:

PR #1997, opened 2024-12-03
fix: add `.pnpm-store` to `.gitignore`

cr7pt0gr4ph7:
## Summary

Depending on your checkout method and setup (for example, within a VS Code DevContainer), `pnpm` creates a `.pnpm-store` folder at the root of your workspace that should be ignored by `git`.

Not ignoring it can cause quite a headache and >10,000 untracked changes due to files in `.pnpm-store`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /node_modules/
+/.pnpm-store/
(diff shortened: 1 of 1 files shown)


brillout:
:+1:

Issue #1692, opened 2024-06-12
New settings `clear` and `default` for controlling inheritance of cumulative configs


brillout:
Add two new settings `default: boolean` and `clear: boolean` for controlling inheritance of cumulative values:

```js
// pages/+config.js

import { Layout } from './Layout'

export default {
  Layout: {
    value: Layout,
    // If a child defines a layout it will replace this layout
    default: true
  }
}
```

```js
// pages/admin-panel/+config.js

import { Layout } from './Layout'

export default {
  // Overrides all previously defined layouts (i.e. all layouts defined by ancestors)
  Layout: {
    value: Layout,
    clear: true
  }
}
```

### Groups

```js
// pages/+config.js

import HeadDefaultOpenGraph from './HeadDefaultOpenGraph'
import HeadGlobal from './HeadGlobal'

export default {
  Head: [
    {
      value: HeadGlobal
    },
    {
      value: HeadDefaultOpenGraph,
      group: 'open-graph' // Can be any string (it's used as a unique identifying key)
    }
  ]
}
```

```js
// pages/some-page/+config.js

import HeadOpenGraph from './HeadOpenGraph'

// This page inherits HeadGlobal but not HeadDefaultOpenGraph

export default {
  Head: {
    value: HeadOpenGraph,
    group: 'open-graph',
    // Alternatively we can add `default: true` to the HeadDefaultOpenGraph value above
    clear: true
  }
}
```

bighorik:
I would like to be able to flexibly configure this inheritance.
In my practice, a situation often happens when, for example, there are three pages:
applications/listing
applications/favorite
applications/create

For the first two, you need to use a common complex Layout, for the latter, a simpler layout. Accordingly, I would like to be able to set my own Layout for /create, overwriting the original one.

However, there are also reverse situations when you need to specify the Layout for 1 of the many pages, for example:
personalAccount/profile
personalAccount/settings
personalAccount/companies/portfolio
personalAccount/companies/stats
personalAccount/companies/about

In this case, I have added additional navigation for pages /companies - the ability to inherit previous Layouts is great

brillout:
@bighorik Yeah, currently the only way is to do this:

```
applications/(complex-layout)/listing
applications/(complex-layout)/favorite
applications/(simple-layout)/create
```

Let me know whether it's a showstopper for you and I'll bump the priority of this ticket.

AgentEnder:
[...]

PR #2005, opened 2024-12-06
Add E2E tests for the behavior of `config.meta.<setting>.env`, `.cumulative` and `.effect`

cr7pt0gr4ph7:
## Rationale
I couldn't find any tests that specifically tested the configuration inheritance mechanism as well as the ordering of inherited cumulative settings, so I've taken the liberty to implement those myself, in preparation for possible further improvements (see e.g. #1692 as well as some type inference improvements related to `config.meta` I intend to contribute).

I've chosen to go with an E2E test here to be able to exercise the build-time logic configuration merging & serialization logic as well. There might be prettier ways to do this, but it gets the job done and covers the user-observable behavior.


brillout:
It's actually something I've been wondering: whether it's worth it to test it.

brillout:
So far I'm inclined to say it isn't worth it. But I'll have a closer look at your tests.

cr7pt0gr4ph7:
> It's actually something I've been wondering: whether it's worth it to test it.

As a user of Vike in one of my projects, I'd say definitely yes. It's such a fundamental part of Vike which has its own complexities (especially when something like `inherit: false` or cumulative settings with groups are going to be implemented) and is possibly security-relevant (when an `env: { server: true }` setting that should not be exposed to clients is unintentionally exposed, which you won't notice in applications because it wouldn't break anything) that I was surprised to not see any tests of it.

brillout:
I'm actually going to work on code related to config handling this week. So it's a good time to add inheritance tests. I'll have a closer look at this PR today/tomorrow (looks good at first sight).

brillout:
The `test/playground` is getting quite big; some refactoring will help maintain it.
 - Move the tests inside `pages/**/*` so that `.testRun.ts` only imports a bunch of functions and executes these functions. In other words: no test logic is directly defined inside `.testRun.ts`.
 - Move `serializeSettings()` inside a new directory `utils/` at the root of the app.
 - Make `serializeSettings()` generic e.g. `serializePageContextProps(pageContext, props: string[])`.
 - Define the the serializing logic `serializeSettings()` and it's parsing counterpart inside the same file.
 - Move `isBrowser.ts` inside `utils/`.

Looking forward to it. It's going to make test/playground a neat testbed.

cr7pt0gr4ph7:
> The `test/playground` is getting quite big; some refactoring will help maintain it.
> [...]

Done :+1: A bit of care has to be taken to not import `@brillout/test-e2e` in files like `serializeSettings.ts` that are shared between the test code and the code-under-test. `@brillout/test-e2e` will try to parse the command line options as soon as it's loaded - which causes it to throw an error if one tries to run the code-under-test separately with e.g. `pnpm run dev --host`.

cr7pt0gr4ph7:
@brillout: The last CI failures where due to the screenshot tests, which this PR should not even have touched upon. I tried to reproduce the issues locally, but couldn't find expected screenshots that the CI compares against...

brillout:
Awesome, collocating is a major improvement. Reading your new tests is now a breeze :sparkles:. I left a final polishing review, let's merge after that.

brillout:
Thank you @cr7pt0gr4ph7 for your contribution :green_heart:

PR #2014, opened 2024-12-09
whitelist Vue DevTools `pageContext` access

brillout:


brillout:
Not needed, see https://github.com/vuejs/devtools/issues/542.

PR #2044, opened 2025-01-02
Fix typo/path in getVirtualFileImportUserCode.ts

f0x52:
Noticed a typo in the generated comment, then realized the path itself seemed to need an update too.
Might be clearer to prefix it with vike/node/plugin/.. instead, so it's even clearer where it's from?


brillout:
:+1: We can also replace the path with a GitHub link.

PR #2048, opened 2025-01-07
Fix typo

Hugo-Thebault:


brillout:
`GuardAsync` already sets the type of `pageContext`

brillout:
Thank you!

PR #2050, opened 2025-01-07
+middleware example with @universal-middleware/router

magne4000:
WIP around how we can move some logic used in `+middleware.ts` process to `universal-middleware`.

@brillout This is mostly to discuss what we should port to `universal-middleware` and create a proper API.

`universal-middleware` in progress PR: https://github.com/magne4000/universal-middleware/pull/95

### DIFF
--- a/test/middleware-config/express-entry.ts
+++ b/test/middleware-config/express-entry.ts
@@ -1,10 +1,10 @@
 import { dirname } from 'node:path'
 import { fileURLToPath } from 'node:url'
 
-import { createMiddleware } from '@universal-middleware/express'
+import type { EnhancedMiddleware } from '@universal-middleware/core'
+import { apply } from '@universal-middleware/express'
 import express from 'express'
 import { getMiddlewares } from 'vike/__internal'
-import type { Middleware } from './pages/Middleware'
 import { createDevMiddleware } from 'vike/server'
 
 const __filename = fileURLToPath(import.meta.url)
@@ -24,22 +24,13 @@ async function startServer() {
     app.use(devMiddleware)
   }
 
-  const middlewares = (await getMiddlewares()) as Middleware[]
-  middlewares.sort((m1, m2) => getOrder(m1.order) - getOrder(m2.order))
-  middlewares.forEach((middlewareSpec) => {
-    const middleware = middlewareSpec.value
-    app.all('*', createMiddleware(() => middleware)())
-  })
+  const middlewares = (await getMiddlewares()) as EnhancedMiddleware[]
+  // @ts-ignore Typescript seems to think that apply requires express 5
+  apply(app, middlewares)
 
   app.listen(port, () => {
     console.log(`Server listening on http://localhost:${port}`)
   })
 
   return app
-}
-
-function getOrder(order: Middleware['order']): number {
-  if (order === 'post') return 100
-  if (order === 'pre') return -100
-  return order
 }
--- a/test/middleware-config/pages/middlewareTelefunc.ts
+++ b/test/middleware-config/pages/middlewareTelefunc.ts
@@ -1,14 +1,10 @@
 export { middlewareTelefunc }
 
+import { type UniversalMiddleware, enhance } from '@universal-middleware/core'
 import { telefunc } from 'telefunc'
-import type { UniversalMiddleware } from '@universal-middleware/core'
-import type { Middleware } from './Middleware'
 
 const telefuncUniversalMiddleware: UniversalMiddleware = async (request, context, runtime) => {
   const url = request.url.toString()
-  const urlParsed = new URL(url)
-
-  if (urlParsed.pathname !== '/_telefunc') return
 
   const httpResponse = await telefunc({
     url,
@@ -28,10 +24,8 @@ const telefuncUniversalMiddleware: UniversalMiddleware = async (request, context
   })
 }
 
-const middlewareTelefunc: Middleware[] = [
-  {
-    name: 'telefunc',
-    order: 'pre',
-    value: telefuncUniversalMiddleware
-  }
-]
+const middlewareTelefunc = enhance(telefuncUniversalMiddleware, {
+  name: 'telefunc',
+  method: 'POST',
+  path: '/_telefunc'
+})
--- a/test/middleware-config/pages/middlewareVike.ts
+++ b/test/middleware-config/pages/middlewareVike.ts
@@ -1,9 +1,8 @@
 export { middlewareVike }
 
-import { renderPage } from 'vike/server'
 // TODO: stop using universal-middleware and directly integrate server middlewares instead. (Bati generates boilerplates that use universal-middleware https://github.com/magne4000/universal-middleware to make Bati's internal logic easier. This is temporary and will be removed soon.)
-import type { UniversalMiddleware } from '@universal-middleware/core'
-import type { Middleware } from './Middleware'
+import { type UniversalMiddleware, enhance } from '@universal-middleware/core'
+import { renderPage } from 'vike/server'
 
 const vikeUniversalMiddleware: UniversalMiddleware = async (request, context, runtime) => {
   const pageContextInit = { ...context, ...runtime, urlOriginal: request.url, headersOriginal: request.headers }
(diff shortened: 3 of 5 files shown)


test/middleware-config/express-entry.ts
```diff
@@ -33,22 +34,19 @@ async function startServer() {
...
-  })
+  const middlewares = (await getMiddlewares()) as (RouteDefinition | MiddlewareDefinition)[]
+  const router = new UniversalRouter()
+  apply(router, middlewares)
+  app.all('*', createMiddleware(() => router[universalSymbol])())
+  // TODO replace with UniversalExpressRouter once done.
+  //      It should look like this:
+  // const router = new UniversalExpressRouter(app)
+  // apply(router, middlewares)
+  //      So no need for manually calling app.*, and rely on express routing
```
magne4000, line 45: This could be moved to `vike-node`, which would be responsible to create the proper `Universal*Router`.

brillout: :+1: And eventually `vike-{express,hono,...}` so that one Vike extension = one tool (e.g. to fit the copy next to the slot machine on the landing page).

test/middleware-config/express-entry.ts
```diff
@@ -33,22 +34,19 @@ async function startServer() {
...
-    app.all('*', createMiddleware(() => middleware)())
-  })
+  const middlewares = (await getMiddlewares()) as (RouteDefinition | MiddlewareDefinition)[]
+  const router = new UniversalRouter()
+  apply(router, middlewares)
+  app.all('*', createMiddleware(() => router[universalSymbol])())
+  // TODO replace with UniversalExpressRouter once done.
+  //      It should look like this:
+  // const router = new UniversalExpressRouter(app)
+  // apply(router, middlewares)
```
brillout, line 44: I wonder if it could be further simplified to this:

```js
import { apply } from '@universal-middleware/express'
apply(app, middlewares)
```

brillout: That would be the only `universal-middleware` code in this file (and inside `vike-express`).

magne4000: Indeed, that's even better

brillout:
I wonder whether/how we can add a name to a universal middleware, so that users can selectively remove middlewares.

Use case: user is using `vike-react-authjs` but wants a custom server integration. The answer could be that he has to fully eject `vike-react-authjs` or he replaces the universal middleware with a custom middleware. I guess both are okay, although if it's easy to implement then it would be nice to have some kind of name or other means to select a universal middleware.

test/middleware-config/express-entry.ts
magne4000: It should be simple enough, and I like the idea.
I can see that in Bati we would then just add a comment somewhere or a paragraph in the README to explain that it can be ejected.
We could even allow the user to retrieve the built-in express like this:
```ts
import { app } from "vike-express";
```

brillout: :+1:

> We could even allow the user to retrieve the built-in express like this:

Neat idea :sparkles:

magne4000:
> I wonder whether/how we can add a name to a universal middleware, so that users can selectively remove middlewares.
> 
> Use case: user is using `vike-react-authjs` but wants a custom server integration. The answer could be that he has to fully eject `vike-react-authjs` or he replaces the universal middleware with a custom middleware. I guess both are okay, although if it's easy to implement then it would be nice to have some kind of name or other means to select a universal middleware.

I was thinking about that with the same use case in mind. IMO the best solution would be to be able to just eject the +middleware, but I think we're not there yet.
So adding a `name` onto them could do the trick for now (and it doesn't hurt).

brillout:
Yea, I think a `name` is always a good thing to have. I'm just wondering how it could fit with the current DX with `withRoute()` and `withOrder()`. Maybe merging these into a single `withMeta()` could do the trick. Actually, I feel like naming that it `withRoute()` instead of `withMeta()` could be a better name. We could then require a `name` when using `withRoute()` (and iff. using `withRoute()` because in the user land setting a `name` isn't that useful I guess).

test/middleware-config/express-entry.ts
magne4000: https://github.com/vikejs/vike/pull/2050/files#diff-6a1963a7349ecacbba17183fc73dce759419880a402305d558aa187bc3ebeb03R35-R37 :sparkles:

test/middleware-config/pages/middlewareTelefunc.ts
```diff
@@ -28,10 +30,7 @@ const telefuncUniversalMiddleware: UniversalMiddleware = async (request, context
...
-  {
-    name: 'telefunc',
-    order: 'pre',
-    value: telefuncUniversalMiddleware
-  }
-]
+const middlewareTelefunc = decorate(telefuncUniversalMiddleware, {
+  method: 'POST',
+  path: '/_telefunc'
+})
```
magne4000, line 36: `withRoute` replaced by a more generic `decorate` helper, to make things more explicit, and simplify adding props in the future

brillout: Good idea to keep it generic. (Name idea: `enhance()` instead of `decorate()`, but I'm fine with `decorate()` if you prefer it.)

magne4000: `enhance` was my first name idea, but I preferred `decorate`. I'll let that macerate a little and see how I feel about it when we're closer to a finished implementation.

test/middleware-config/pages/middlewareVike.ts
```diff
@@ -19,10 +20,7 @@ const vikeUniversalMiddleware: UniversalMiddleware = async (request, context, ru
...
-const middlewareVike: Middleware[] = [
-  {
-    name: 'vike',
-    order: 'post',
-    value: vikeUniversalMiddleware
-  }
-]
+const middlewareVike = decorate(vikeUniversalMiddleware, {
+  method: 'GET',
+  path: '/**'
```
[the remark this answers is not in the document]
brillout: Unless I'm missing something, we could make it that internally all universal middlewares are catch-all middlewares but we wrap a `if (!matches(url, path)) return` condition. This would allow us to use any path syntax we want.

test/middleware-config/pages/middlewareTelefunc.ts
brillout: :+1: The thing I prefer about `enhance()` is that it's more visual and concrete, whereas `decorate()` is more abstract and vague (or maybe it's subjective), thus more difficult to remember.

Bonus: in the docs we can mention "enhanced middlewares", for example:
 - > Enhanced middlewares are middlewares that define a route
 - > Enhanced middlewares must have a name

brillout: But I guess we could also say "decorated middlewares".

test/middleware-config/pages/middlewareVike.ts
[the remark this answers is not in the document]
magne4000: :thinking: that could work.

Currently I'm relying on servers themselves to handle the routing, and fallback to using `rou3` for the `UniversalRouter`. But this is the cause of what I mentioned above.

If I use `rou3` for all servers in catch-all, I think that would also simplify some other future ideas I had for handling deployments. 
For now I chose `rou3` because it's tree-shakable and minimal: asdasd (it would be directly bundled into `universal-middleware`)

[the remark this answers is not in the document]
brillout: :+1: Sounds good.

Another idea: we only support a trivial path syntax and we tell users to define a catch-all with a `if (!url...) return` condition if they want to define advanced routing logic.

Kinda like Vike: Route Strings are very basic while Route Functions enable programmatically defined routes. This strategy has been working well: no one's complaining about the limitations of Route Strings.

But using `rou3` sounds good to me as well.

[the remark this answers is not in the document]
magne4000: `rou3` also have the benefit of prioritizing nested routes.

For instance, with the following patterns `/*` and `/a/**`, the path `/a/b/c` will always match the second pattern.
- This is usually what the user will want
- This does not require the user to register route in a specific order
- This does not require a lib to specify an `order` for middlewares with a `path`

test/middleware-config/pages/middlewareTelefunc.ts
magne4000: In the end I want for `enhance` :)

magne4000:
Some documentation around the `enhance` and `apply` helpers: https://d80b2041.universal-middleware.pages.dev/helpers/enhance (PR: https://github.com/magne4000/universal-middleware/pull/97)

brillout:
Love it

PR #2070, opened 2025-01-10
fix: make vite peer dependency optional

brillout:


brillout:
CC @magne4000 This is quite a neat technique; also relevant for the Stem vision. It seems to be widely supported, see [package manager discussions](https://stackoverflow.com/questions/72750797/should-a-peer-dependency-be-added-in-both-peerdependencies-and-dependencies-in-p/79347029#79347029).

magne4000:
I did not know that, that's neat!

PR #2082, opened 2025-01-16
Added new fastify+vike example

whyte50:


brillout:
Nice, thank you :green_heart: How about adding some description to the repository's readme?

PR #2096, opened 2025-01-20
refactor: improve resolve entry routing

rtritto:


brillout:
Let's focus on real things, instead of superficial changes.

rtritto:
But this is an easy change that has 0 impacts.
Code quality and performance are important.
If our eye falls for corrections, doing quick PR, it doesn't waste our time.

brillout:
We can agree to disagree. As always: contributing only works if you trust the maintainers that they know what they're doing.

Issue #2093, opened 2025-01-20
[Build] Prevent /server/assets creation in Vike V1


rtritto:
### Description

With Vike design V0, `/dist/server/assets` files aren't bundled, while with V1 one they are bundled.

The bundle logic for Vike V1 was fixed in https://github.com/vikejs/vike/issues/2034 where `/dist/server/assets/*` files are created during the build and then removed.

The `/dist/server/assets` should not be bundled with Vike design V1 as V0 one.

### Informations
- vike v0.4.218
- [Vike design V1 migration guide](https://vike.dev/migration/v1-design)

### Vike design V0
Run `yarn build` on [vike-with-vercel#vike-design-v0](https://github.com/rtritto/vike-with-vercel/tree/vike-design-v0)

Build logs:
```console
Running "yarn run build"
vite v6.0.8 building for production...
transforming...
9:58:50 AM [vike][Warning] Update to the new V1 design to get rid of Vite's warning: The glob option "as" has been deprecated in favour of "query". See https://vike.dev/migration/v1-design for how to migrate.
The glob option "as" has been deprecated in favour of "query". Please update `as: 'extractAssets'` to `query: '?extractAssets'`.
The glob option "as" has been deprecated in favour of "query". Please update `as: 'extractExportNames'` to `query: '?extractExportNames'`.
The glob option "as" has been deprecated in favour of "query". Please update `as: 'extractExportNames'` to `query: '?extractExportNames'`.
The glob option "as" has been deprecated in favour of "query". Please update `as: 'extractExportNames'` to `query: '?extractExportNames'`.
The glob option "as" has been deprecated in favour of "query". Please update `as: 'extractAssets'` to `query: '?extractAssets'`.
✓ 165 modules transformed.
rendering chunks...
computing gzip size...
dist/client/_temp_manifest.json                                                     3.39 kB │ gzip:  0.63 kB
dist/client/assets/static/logo.2_7Lo9tV.svg                                         5.79 kB │ gzip:  1.97 kB
dist/client/assets/static/renderer_PageLayout-5d219a08.Cjl9f7UM.css                 0.09 kB │ gzip:  0.10 kB
dist/client/assets/static/renderer_PageLayout-83e8fff7.B8X9vkNx.css                 0.38 kB │ gzip:  0.23 kB
dist/client/assets/entries/renderer_default.page.server.extractAssets.C6A_S5ZQ.js   0.07 kB │ gzip:  0.04 kB
[...]

PR #2115, opened 2025-01-27
fix: improve the server assets logic

rtritto:
Implementation of https://github.com/vikejs/vike/issues/2093#issuecomment-2615231923

Fix #2093

### TODO
- [ ] remove the logic of [commit1](https://github.com/vikejs/vike/commit/c1cd2d978b9b811c159551927d0f843d7709066e) and [commit2](https://github.com/cr7pt0gr4ph7/vike/commit/87669c5c05584e77dc783b2aa9104b3e0f47e5bf) (introduced to fix #2034) because, with `ssrEmitAssets=true`, the `dist/server/assets` files went deleted

PR #2126, opened 2025-01-29
fix: dupe server assets logic

rtritto:
Implementation of https://github.com/vikejs/vike/issues/2093#issuecomment-2615231923

Follow up of #2115

Fix https://github.com/vikejs/vike/issues/2093

### Changes
- disable `ssrEmitAssets` by default as Vite behavior
- add feature `ssrEmitAssets=true` (the user can set `ssrEmitAssets=true`) to bundle the server assets
  - in the previous logic, `ssrEmitAssets` is always enabled and the bundled server assets were removed at the end of build
- improve the _create-and-delete_ server assets generated by `ssrEmitAssets=true`

### DIFF
--- a/vike/node/plugin/plugins/buildConfig.ts
+++ b/vike/node/plugin/plugins/buildConfig.ts
@@ -68,8 +68,10 @@ function buildConfig(): Plugin[] {
           {
             isServerAssetsFixEnabled = fixServerAssets_isEnabled() && (await isV1Design(config))
             if (isServerAssetsFixEnabled) {
+              // https://github.com/vikejs/vike/issues/2093
+              // `ssrEmitAssets: true` not needed because it dupe the server assets
+              // if needed, the user can set `ssrEmitAssets: true` in vite.config.js
               // https://github.com/vikejs/vike/issues/1339
-              config.build.ssrEmitAssets = true
               // Required if `ssrEmitAssets: true`, see https://github.com/vitejs/vite/pull/11430#issuecomment-1454800934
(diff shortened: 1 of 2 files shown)


brillout:
I don't think your changes make any sense? Have you read the links I've shared with you at https://github.com/vikejs/vike/issues/2093#issuecomment-2613930880?

rtritto:
> I don't think your changes make any sense?

Why?

> Have you read the links I've shared with you at [#2093 (comment)](https://github.com/vikejs/vike/issues/2093#issuecomment-2613930880)?

Yes, I answered here https://github.com/vikejs/vike/issues/2093#issuecomment-2615231923

brillout:
You're removing `config.build.ssrEmitAssets = true` which disables the whole workaround. Once you disabled it, of course, you can start removing "useless code".

I think the crux of the problem here is that you assume certain things that are false. You then proceed with making changes based on your false assumptions. So I'd kindly ask you to *double*, *triple*, *quadruple* check your assumptions before starting making changes. From my perspective, it's a bit like if someone would do random changes and then ask me "why do you think my changes are breaking things?". As you can imagine it's time consuming for me (while time is a very limited asset as an open source developer). Thank you.

brillout:
Closing. From now let's do it like this: let's be crystal clear about the problem before opening any PR.

If you can publish a *clear* reproduction with *clear* steps with how to reproduce a *concrete* problem then, and only then, will I consider reviewing a PR.

Thank you for understanding.

Issue #624, opened 2023-02-03
Don't early hint fallback assets


brillout:
### Description

See https://github.com/brillout/vite-plugin-ssr/issues/618.

brillout:
The implementation (https://github.com/brillout/vite-plugin-ssr/commit/693ff006e5971ed310a3916304157586e255c896) is brittle as it simply picks the first one. The user has to define the fallback assets after the primary one, in the hope that Vite doesn't shuffle the order in its internal asset graph representation.

brillout:
Fix released in `0.4.75`. I didn't test it but it should work. Let me know if it doesn't.

brillout:
Indeed, let me have a closer look at it.

brillout:
Alright, new fix released at `0.4.76` which I tried and works with your repro. Sorry for the back-and-forth.

FlorianCassayre:
I think this issue is back. In my case I can see both `.woff2` and `.woff` fonts registered for preloading (while it would make sense to only preload `.woff2`). I'm pretty much running the latest version of Vike.

Page: https://florian.cassayre.me/ ([source code](https://github.com/FlorianCassayre/florian.cassayre.me))

Here is another website from the community, written with Vike suffering from the same problem: https://ark-unity.com/ (this time with `.woff` and `.ttf`)

brillout:
@FlorianCassayre This was [only applied to early hints](https://github.com/vikejs/vike/blob/385293c9822f2688fa01667ad2d73b966937a422/vike/node/runtime/renderPage/getEarlyHints.ts#L12-L17). But, indeed, we [should also apply it to preload tags](https://github.com/vikejs/vike/blob/385293c9822f2688fa01667ad2d73b966937a422/vike/node/runtime/html/injectAssets/getHtmlTags.ts#L76). I'll implement it this week (or PR welcome in case that's something you'd be up to).

PR #2144, opened 2025-02-05
fix: fallback fonts should not be preloaded

FlorianCassayre:
As discussed on https://github.com/vikejs/vike/issues/624#issuecomment-2636175999.


FlorianCassayre:
Tested locally, appears to work as desired. I see `test/preload`, I'll try to add a testcase there.

brillout:
:100: Neat.

I think it's fine; no need to write a test for it.

Let me know once this is ready for review.

test/preload/__snapshots__/prod.spec.ts.snap
```diff
@@ -52,7 +52,7 @@ exports[`preload tags > Preload Default 2`] = `
     <html>
       <head>
         <link rel="stylesheet" type="text/css" href="/assets/static/renderer_Layout-031b266d.$HASH.css">
-        <link rel="preload" href="/assets/static/Sono-Light.$HASH.ttf" as="font" type="font/ttf" crossorigin>
+        <link rel="preload" href="/assets/static/Sono-Light.$HASH.woff2" as="font" type="font/woff2" crossorigin>
```
FlorianCassayre, line 55: Production-like preloading, fallback font is omitted.

test/preload/__snapshots__/prod.spec.ts.snap
```diff
@@ -177,6 +177,7 @@ exports[`preload tags > Preload Eager 2`] = `
     <html>
       <head>
         <link rel="stylesheet" type="text/css" href="/assets/static/renderer_Layout-031b266d.$HASH.css">
+        <link rel="preload" href="/assets/static/Sono-Light.$HASH.woff2" as="font" type="font/woff2" crossorigin>
         <link rel="preload" href="/assets/static/Sono-Light.$HASH.ttf" as="font" type="font/ttf" crossorigin>
```
FlorianCassayre, line 181: Development-like preloading, both fonts are included.
Debatable if we should keep it like this, but I guess it doesn't hurt since it was already like that.

FlorianCassayre (commented):
Alright, I'm pretty sure it's correct. Thanks for the pointers by the way, that made it really easy to make the changes.

brillout:
:rocket: Thank you Florian.

brillout:
Pre-released as `0.4.220-commit-af5c91f`.

PR #2146, opened 2025-02-06
feat: add compatibility with Vite 6 Environment API

magne4000:
This is a WIP to add Environment API compatibility, by calling `buildApp` instead of `vite.build` (keeping current behaviour as default).

- new ~~`useEnvironmentAPI`~~ `viteEnvironmentAPI` config. Naming, location and whether it should be on the Vike CLI only -> to be discussed
- `autoFullBuild` is disabled in this mode
- New `vike:buildApp` plugins to mimic `autoFullBuild`


brillout:
Btw. I should be done with my refactoring soon. The bulk of it is done, it's now "only" about small things.

brillout:
> * Naming

I'd suggest `viteEnvironmentAPI`.

> `vike.dev/TODO`

I guess we don't need this for now. It's mostly/only going to be used by our own extensions anyways.

brillout:
Your IDE seems to re-order imports. Can we revert these changes? We can eventually agree on some new linting rules after I merged all pending PRs & branches.

magne4000:
> Your IDE seems to re-order imports. Can we revert these changes? We can eventually agree on some new linting rules after I merged all pending PRs & branches.

Indeed, and I still don't know why, as the option is disabled :/

magne4000:
> > Your IDE seems to re-order imports. Can we revert these changes? We can eventually agree on some new linting rules after I merged all pending PRs & branches.
> 
> Indeed, and I still don't know why, as the option is disabled :/

It seems like the IntelliJ Biome plugin does `biome check --apply` even if `Run Biome check with --apply` is unchecked... I'm disabling it and reversing the changes

vike/node/plugin/plugins/buildConfig.ts
```diff
@@ -109,7 +109,8 @@ function buildConfig(): Plugin[] {
         order: 'pre',
         sequential: true,
         async handler(options, bundle) {
-          if (isSsrBuild) {
+          if (isSsrBuild || this.environment.name === 'ssr') {
```
magne4000, line 112: I need to check if `this.environment` exists in Vite 5.

magne4000: It does not

magne4000:
You can do your refactoring @brillout. I do not know how I can manage to create a unit test with vite@5 in the same monorepo. Calling `vike build` always use the Vite version in the linked `vike` package. We could call `vite build` instead, but that doesn't seem really wise.

brillout:
I see. Indeed I think we are currently not testing Vite 5 at all.

brillout:
https://github.com/vikejs/vike/pull/2146/commits/d296d913d3721c601e9366e03a591460b11877d5; let me know if that's an issue.

brillout:
I made it that some `test-deprecated-design/` are testing Vite 5, e.g.:
https://github.com/vikejs/vike/blob/6ff9bd75b3077f98a95ac80a6b56f57ece0f52fb/test-deprecated-design/vue-full/package.json#L18

Indeed, once Vike removes support for Vite's CLI, this won't be possible anymore. My idea is that some `vike-{react,vue}` examples can be testing Vite 5, that should be good enough for now (at least between Vite 6 and Vite 5).

PR #2147, opened 2025-02-06
chore: bump pnpm

magne4000:
Most dependencies have been updated as a side effect, and `e2e` prod tests all fail unexpectedly


package.json
```diff
@@ -43,8 +43,9 @@
   },
   "pnpm": {
     "overrides": {
-      "vike": "link:./vike/"
+      "vike": "link:./vike/",
+      "vike>vite": "^6.0.5"
```
magne4000, line 47: Without it, `vike` package reuses the `vite` version used by `vitest` declared in the workspace root.
It just ignores the `"devDependencies": { "vite": "^6.0.5" }` declared in `vike`.

brillout:
The Vitest failed tests can be updated, similarly to what you did in your last commit. As for the TS failure of test-deprecated-design/vue-full/ we can pin some older version if it helps.

package.json
brillout: Even if setting `"vite": "6.0.5"` (wtihout `^`)?

magne4000:
> Most dependencies have been updated as a side effect, and `e2e` prod tests all fail unexpectedly

Hum, they only fail on my machine it seems, `playwright` really doesn't like being updated...

package.json
magne4000: > Even if setting `"vite": "6.0.5"` (wtihout `^`)?

Yes, even like that

magne4000: > If possible, it'd be nice to avoid overriding the vite version globally.

It declares it globally, but overrides it only for `vike`. And I haven't found another workaround

brillout:
LGTM once the failed Vitest test is updated.

magne4000:
I reverted my "fix" on the test file. I actually have no clue of what's happening.
In my terminal the test passes only with the "fix", but in my IDE and on the CI, it passes only without it.

package.json
brillout: Actually, the latest Vitest works with Vite 6: https://github.com/vitest-dev/vitest/blob/12eaf3ea4369c46ab576e371e729a7cad9dfb647/packages/vitest/package.json#L174.

So maybe a better workaround could be `vitest>vite` instead of `vike>vite`.

brillout:
It's because of https://github.com/brillout/picocolors/blob/b659f67bbc1f9a0f34a740077953fe83a930c822/picocolors.js#L3.

brillout:
And this: https://github.com/brillout/picocolors/blob/b659f67bbc1f9a0f34a740077953fe83a930c822/picocolors.browser.mjs#L6. I'm hesitating whether this was a good idea :thinking: 

Can we make sure to align the value `isColorSupported()` between local and CI? Maybe we can add some env var to Vitest tests?

magne4000:
> And this: https://github.com/brillout/picocolors/blob/b659f67bbc1f9a0f34a740077953fe83a930c822/picocolors.browser.mjs#L6. I'm hesitating whether this was a good idea 🤔
> 
> Can we make sure to align the value `isColorSupported()` between local and CI? Maybe we can add some env var to Vitest tests?

I was thinking the same thing. I'll enforce `NO_COLOR`

package.json
brillout: If everyone is using the same Vite version, then there shouldn't be a problem in the first place, I guess.

brillout:
> FORCE_COLOR

:+1: Even better. LGTM, although maybe it could be worth it to have another look at https://github.com/vikejs/vike/pull/2147#discussion_r1945006593.

package.json
magne4000: We could bump `vitest`, let me check if it breaks anything.

brillout: Currently Vike is using `    "vitest": "^2.1.3"` so maybe bumping Vitest does the trick.

brillout: > We could bump `vitest`, let me check if it breaks anything.

It usually doesn't, but let me know if it does — I can give a hand.

magne4000: Seems fine

magne4000:
> [remove superfluous vite dep](https://github.com/vikejs/vike/pull/2147/commits/1f9c5e3b1d7c620fd77e4358cd63b2e900c48ed8)

It's risky. I had to add it because otherwise it kept using `vite@5` (after a fresh install). So it's possible that any dependency update could rollback to using `vite@5` without it.

magne4000:
> > [remove superfluous vite dep](https://github.com/vikejs/vike/pull/2147/commits/1f9c5e3b1d7c620fd77e4358cd63b2e900c48ed8)
> 
> It's risky. I had to add it because otherwise it kept using `vite@5` (after a fresh install). So it's possible that any dependency update could rollback to using `vite@5` without it.

For instance, without it, `rm -fr node_modules/ pnpm-lock.yaml && pnpm i` installs `vite@5`

brillout:
Hm that's unexpected (I guess a pnpm bug? Why would pnpm pick Vite 5 - a lower version - if everyone can use Vite 6?). I went ahead and merged; I'll revive the explicit vite dep when it becomes necessary.

magne4000:
> Hm that's unexpected (I guess a pnpm bug? Why would pnpm pick Vite 5 - a lower version - if everyone can use Vite 6?). I went ahead and merged; I'll revive the explicit vite dep when it becomes necessary.

This whole vite version thing is weird, `pnpm` doesn't act as I would expect, I agree

magne4000:
> > Hm that's unexpected (I guess a pnpm bug? Why would pnpm pick Vite 5 - a lower version - if everyone can use Vite 6?). I went ahead and merged; I'll revive the explicit vite dep when it becomes necessary.
> 
> This whole vite version thing is weird, `pnpm` doesn't act as I would expect, I agree

I created [an issue](https://github.com/pnpm/pnpm/issues/9058) for it.

PR #2183, opened 2025-02-14
feat: universal-middleware

magne4000:


brillout:
LGTM.

Most notable change: https://github.com/vikejs/vike/pull/2183/commits/02a60c4047f82c63bbbff0536c69c4c881ed95cf

Feel free to squash & merge whenever you want.

When squashing:

image

I.e. remove all commit text except of the last line. (That's also what the Vite team does.)

magne4000:
> Most notable change: [02a60c4](https://github.com/vikejs/vike/commit/02a60c4047f82c63bbbff0536c69c4c881ed95cf)

IIRC, at some point, this was causing crashes with react when `stream: true`, because internally Vike was instanciating some Node Stream, and using `getReadableWebStream()` in this case was causing errors. Using `response.pipe` however did not.
I do not remember the complete test case though.

brillout:
Actually, thinking of it, I like the transform: it allows Vike land to use any stream type (Web Stream or Node.js Stream) while converting it to a Web Stream for universal-middleware. I went ahead and reverted.

magne4000:
> Ah, no, wait. It actually doesn't transform the stream types. Wishful thinking 😅. How about we remove `new TransformStream()` and, if it causes issues, I'll dig into it? I went ahead and re-applied my change.

LGTM

PR #2193, opened 2025-02-17
perf: switch from `fast-glob` to `tinyglobby`

benmccann:
The docs say:

>## Minimal and self-contained (no deps)
>
>[All dependencies](https://github.com/vikejs/vike/blob/3d042da/vike/package.json#L15-L26) are either shared with Vite (e.g. [`fast-glob`](docs/pages/nextjs/+Page.mdx:))

But actually Vite switched to tinyglobby. There's also a PR out to switch vitest to tinyglobby. I thought it'd be good to stay aligned here

`fast-glob` has 17 dependencies vs just 2 for `tinyglobby`


brillout:
Thank you Ben!

PR #2208, opened 2025-02-21
fix: compatibility with bun runtime

Octonath:
Fix compatibility with the bun runtime by not removing the extension from the import when resolving it

### DIFF
--- a/vike/utils/requireResolve.ts
+++ b/vike/utils/requireResolve.ts
@@ -18,7 +18,9 @@ assertIsNotProductionRuntime()
 function requireResolve(importPath: string, cwd: string): string | null {
   assertPosixPath(cwd)
   const clean = addFileExtensionsToRequireResolve()
-  importPath = removeFileExtention(importPath)
+  if (!globalThis.navigator) {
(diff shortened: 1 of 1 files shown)


brillout:
Isn’t `globalThis.navigator` a brittle way to check whether the runtime is Bun?

brillout:
Any reason to not use [this](https://bun.sh/guides/util/detect-bun)?

Octonath:
Not really I was using the [Hono bun adaptater](https://github.com/honojs/vite-plugins/blob/main/packages/dev-server/src/adapter/bun.ts) from the vite dev server as a reference but i'll try that

Octonath:
The only thing is we would need to add @types/bun for typescript

brillout:
Or let's `@ts-ignore` it.

Octonath:
Had my time wrong on my computer but done

brillout:
Thank you!

brillout:
Pre-released as `0.4.223-commit-f1089b4`.

Octonath:
Retested with the pre-release and works, thank you too for the quick comments!

PR #2263, opened 2025-03-11
fix: pipeable web stream

magne4000:
Fixes errors like this one:
```
Error: [vike][Wrong Usage] pageContext.httpResponse.getReadableWebStream() can't be used because the onRenderHtml() hook defined by vike-solid/__internal/integration/onRenderHtml provides a Web Stream Pipe. Make sure the onRenderHtml() defined by vike-solid/__internal/integration/onRenderHtml hook provides a Readable Web Stream instead. See https://vike.dev/streaming for more information.
```


brillout:
Makes sense :ok_hand: Let me know if a (pre-)release would be convenient for you.

magne4000:
> Let me know if a (pre-)release would be convenient for you.

I'll `release:commit` myself, and I'll ping you when we'll need it on stable version.

Issue #1270, opened 2023-11-20
throw redirect() falsely detects infinite redirect loop


nitedani:
### Description

Reproduction: https://stackblitz.com/github/nitedani/vike-redirect-error?file=pages%2Findex%2Findex.page.jsx
Steps:
1. open in new tab
2. go to /?redirect=true -  you will be redirected to the /about page
3. go to /about?redirect=true

If I comment out `validate(graph);` in `assertNoInfiniteHttpRedirect`, it works as expected.

It would be nice If I could use the following for navigation/redirect:
```ts
import { redirect as navigateServer } from "vike/abort";
import { navigate as navigateClient } from "vike/client/router";

export const navigate = (
  url: `/${string}` | `https://${string}` | `http://${string}`
) => {
  if (import.meta.env.SSR) {
    throw navigateServer(url);
  } else {
    navigateClient(url);
    return null;
  }
};
```

brillout:
👍 I could reproduce and it's indeed a bug.

brillout:
Fixed, thanks for opening the bug report 👌

nitedani:
Does this work if the redirect condition is not based on the url? For example, I would like to use my custom useProfile hook, that calls the backend(sending the session cookie), and either returns the profile or undefined.

brillout:
Exactly, that isn't possible. That's why redirection loops are forbidden (regardless of whether they are contidional or not since Vike cannot know that).

> If no profile on dashboard, redirect to login. If profile on login, redirect to dashboard

I'm realizing that's quite a common use case, and it's quite a bummer that Vike is limiting in that regard.

I guess it's either we remove the infinite loop detection, or we limit the user 🤔

nitedani:
I think the error should be a warning (possible redirect loop detected), shown only once.

brillout:
An infinite loop of HTTP redirects is quite a bad outcome (overloads the server), that's why I'm inclined to make it an error in order to break the infinite loop.

brillout:
So far no one has complained about the infinite redirect loop catcher being too strict, so I guess the solution would be to have a new setting `+doNotCatchInfiniteRedirect` if someone complains.

Issue #1103, opened 2023-08-31
redirect to absolute URL not working


bagr001:
### Description

Hi, I found a bug while migrating to new `redirect()` from `'vite-plugin-ssr/abort'`

The signature promises support of relative and absolute urls, but redirection to absolute url crashes the app.

```
declare function redirect(url: `/${string}` | `https://${string}` | `http://${string}`): AbortRedirect;
```

Here is the error stack:
```
23:03:41 [vite-plugin-ssr][request(1)] throw redirect(302) intercepted while rendering /redirect
Error: [vite-plugin-ssr@0.4.140][Bug] You stumbled upon a bug in vite-plugin-ssr's source code. Go to https://github.com/brillout/vite-plugin-ssr/issues/new and copy-paste this error; a maintainer will fix the bug (usually under 24 hours).
    at assertNoInfiniteHttpRedirect (file:///~app/node_modules/vite-plugin-ssr/dist/esm/node/runtime/renderPage/createHttpResponseObject/assertNoInfiniteHttpRedirect.js:8:5)
    at createHttpResponseObjectRedirect (file:///~app/node_modules/vite-plugin-ssr/dist/esm/node/runtime/renderPage/createHttpResponseObject.js:38:5)
    at handleAbortError (file:///~app/node_modules/vite-plugin-ssr/dist/esm/node/runtime/renderPage.js:385:30)
    at renderPageAlreadyPrepared (file:///~app/node_modules/vite-plugin-ssr/dist/esm/node/runtime/renderPage.js:141:35)
    at renderPageAndPrepare (file:///~app/node_modules/vite-plugin-ssr/dist/esm/node/runtime/renderPage.js:93:12)
    at file:///~app/node_modules/vite-plugin-ssr/dist/esm/node/plugin/shared/getHttpRequestAsyncStore.js:68:35
    at renderPage (file:///~app/node_modules/vite-plugin-ssr/dist/esm/node/runtime/renderPage.js:46:50)
    at file:///~app/server/index.ts:48:29
```

I see the problem in `assertNoInfiniteHttpRedirect` where the check `assert(urlSource.startsWith('/'))` fails.

Thanks :-)

brillout:
I see, I'll fix it tomorrow.

bagr001:
Thanks! 

I just have a question, why is the public API of `redirect()` limited to 302. What is the benefot of disallowing 301? Can you pls explain?

brillout:
[...]

PR #2264, opened 2025-03-11
fix: fix catch of infinite redirections (fix #2262)

### LINKED ISSUE #2262: Infinite Redirect Loop When Redirecting to Default Language Without a Language Slug
## Description

Redirecting from a language-specific URL (e.g., /fr/) to the same page in the default language (e.g., /) incorrectly triggers an infinite redirect loop in Vike. This issue occurs across all pages.

## Reproduction Steps

1. Create a Vike project with language routing
2. Implement a redirect mechanism that removes the language slug when the user's language (from cookies) matches the default language.
3. Visit a non-default language URL (e.g., `/fr/`) with a cookie set to the default language
4. Observe the error: `Error: [vike][Wrong Usage] Infinite loop of HTTP URL redirects: / -> /`

## Minimal Reproducible Example

```typescript
// renderer/+onBeforeRoute.ts
import { OnBeforeRouteSync, PageContext } from 'vike/types';

const onBeforeRoute = (pageContext: PageContext): ReturnType<OnBeforeRouteSync> | undefined => {
  if (typeof window === 'undefined') {
    const { urlOriginal, cookies } = pageContext;
    const lang = getLanguageFromUrl(urlOriginal);
    const urlWithoutLocale = urlOriginal.replace(`/${langParam}/`, '/');
    
    return {
      pageContext: {
        lang: lang || 'en',
        langCookie: cookies?.language,
        urlLogical: urlWithoutLocale,
      },
    };
  }
};

const getLanguageFromUrl = (url: string) => {
  const match = url.match(/^\/([a-z]{2})\//);
  return match ? match[1] : '';
};

export { onBeforeRoute };

// renderer/+onBeforeRender.tsx
const onBeforeRender = async (pageContext: PageContext): ReturnType<OnBeforeRenderAsync> => {
  if (typeof window === 'undefined') {
    const { langCookie, lang, urlLogical } = pageContext;
  
    // If user's cookie language is different from URL language
    if (langCookie && langCookie !== lang) {
        const redirectLang = langCookie === 'en' ? '' : `/${langCookie}`;
        throw redirect(`${redirectLang}${urlLogical}`, 302);
    }
  }
};

export { onBeforeRender };
```

## Expected Behavior

Users should be redirected from `/fr/` to `/` without errors.

## Actual Behavior

Vike throws an error: `Error: [vike][Wrong Usage] Infinite loop of HTTP URL redirects: / -> /`
I think problem is that Vike is only comparing the logical URL (`/`) and not considering that the original URL was different (`/es/`).

Commenting out `validate(graph)` in `assertNoInfiniteHttpRedirect.js` resolves the issue. Adding a query param (/?redirected=true) also works, but it isn’t a viable solution due to SEO reasons.

I’m in the process of migrating my project to Vike, but this issue is a major blocker. A fix or workaround would be greatly appreciated. Maybe I'm cooking it wrong and there is another way to do it but in the end I need this exact redirect logic in my app.

brillout:
Fixes:
- #2262

History of `assertNoInfiniteHttpRedirect()`:
 - https://github.com/vikejs/vike/issues/1270
 - https://github.com/vikejs/vike/issues/1103
 - https://github.com/vikejs/vike/commit/522bd0fa3190e31aa4f59e781f7661d1a9569b12

I wonder why I used `urlLogical` instead of `urlOriginal`.

>   // The URL we assume the redirect to be logically based on
>   urlLogical: string

Isn't it overthinking it? Can't we just simply observe the URL redirection chain for loops, regardless of the semantics behind the URL redirections? This PR makes `assertNoInfiniteHttpRedirect()` use `urlOriginal` instead of `urlLogical`, let's see if a user complains.

### DIFF
--- a/vike/node/runtime/renderPage.ts
+++ b/vike/node/runtime/renderPage.ts
@@ -24,7 +24,6 @@ import {
   prependBase,
   removeUrlOrigin,
   setUrlOrigin,
-  createUrlFromComponents,
   isUri,
   type UrlPublic,
   getUrlPretty
@@ -529,7 +528,7 @@ async function normalizeUrl(
     httpRequestId,
     'info'
   )
-  const httpResponse = createHttpResponseRedirect({ url: urlNormalized, statusCode: 301 }, pageContextInit.urlOriginal)
+  const httpResponse = createHttpResponseRedirect({ url: urlNormalized, statusCode: 301 }, pageContextInit)
   const pageContextHttpResponse = createPageContext(pageContextInit)
   objectAssign(pageContextHttpResponse, { httpResponse })
   return pageContextHttpResponse
@@ -576,7 +575,7 @@ async function getPermanentRedirect(
     httpRequestId,
     'info'
   )
-  const httpResponse = createHttpResponseRedirect({ url: urlTarget, statusCode: 301 }, urlWithoutBase)
+  const httpResponse = createHttpResponseRedirect({ url: urlTarget, statusCode: 301 }, pageContextInit)
   const pageContextHttpResponse = createPageContext(pageContextInit)
   objectAssign(pageContextHttpResponse, { httpResponse })
   return pageContextHttpResponse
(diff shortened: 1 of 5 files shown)


brillout:
> I wonder why I used `urlLogical` instead of `urlOriginal`.
> 
> > // The URL we assume the redirect to be logically based on
> > urlLogical: string
> 
> Isn't it overthinking it? Can't we just simply observe the URL redirection chain for loops, regardless of the semantics behind the URL redirections? This PR makes `assertNoInfiniteHttpRedirect()` use `urlOriginal` instead of `urlLogical`, let's see if a user complains.

An infinite semantic loop translates into an infinite `urlOriginal` loop, unless the semantic loop produces an infinite amount of `urlOriginal` but I can't think of a setup where that would happen.

Going for `urlOriginal` instead of `urlLogical` seems to be the way to go.

PR #2284, opened 2025-03-19
feat: [+mode](https://vike.dev/mode)

brillout:


brillout:
Pre-released as 0.4.225-commit-b8fc36e

PR #2289, opened 2025-03-20
doc: vike-server

magne4000:


docs/pages/server/+Page.mdx
```diff
@@ -0,0 +1,71 @@
+import { Link } from '@brillout/docpress'
+
+`vike-server` comes with the following benefits:
+- Out-of-the-box support for popular servers (Express, Hono, Hattip, Elysia, etc.)
+- Out-of-the-box support for popular serverless/edge environment (Cloudflare, Vercel, etc.)
+- Run server-side code with HMR support in development
```
brillout, line 6:
```suggestion
- HMR
```

magne4000:
Should we move most of the documentation of `vike-server` README directly in vike.dev?

docs/pages/server/+Page.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+> Coming soon.
+
+## Manual integration
+
+If you want more control over your server integration, see <Link href="/server-integration" />
+
+## See also
+
+- <Link href="/renderPage" />
```
brillout, line 69:
```suggestion
- <Link href="/server-integration" />
```

docs/pages/server/+Page.mdx
```diff
@@ -0,0 +1,71 @@
...
+// package.json
+
+"scripts": {
+  "dev": "vike dev",
+  "build": "vike build",
++ "prod": "NODE_ENV=production node dist/server/index.js"
+}
+```
+
+> Check <a href="https://github.com/vikejs/vike-server">vike-server</a> repository for more information.
```
brillout, line 53: How about we move the entire docs in here? The issue with `README.md` is that would lead to a lot of code duplication if we want a section per server `## Express.js`, `## Hono`, ...

brillout:
> Should we move most of the documentation of `vike-server` README directly in vike.dev?

I think so yes, see rationale in my sibling comment.

magne4000:
@brillout `vike-server`'s README have been moved to this PR.

The wall of servers snippets is quite huge, I think it would look better behind tabs, like the one I used in https://universal-middleware.dev/recipes/runtime#available-properties. Do we already have a Component to do that?

Also, the next steps that first need to be implemented in `vike-server`:
- Support for _less-magical_ entries: one entry for node, one for cloudflare, etc. This will not require the use `serve` (`apply` will still be required, but could also be replaced by `app.use(vike)` if deemed necessary) 
- Support for `vike-cloudflare` is on its way
- Support for `vike-vercel` still need some work

brillout:
> The wall of servers snippets is quite huge, I think it would look better behind tabs, like the one I used in https://universal-middleware.dev/recipes/runtime#available-properties.

Agreed.

> Do we already have a Component to do that?

No.

> Also, the next steps that first need to be implemented in `vike-server`:

Sounds good :eyes:

magne4000:
@brillout I invite you to take a look at the Tabs usage. It uses [react-tabs](https://github.com/reactjs/react-tabs), with a custom CSS.

image

I think that once we're satisfied with how it looks, it could be moved to `@brillout/docpress`.

brillout:
I like the aesthetics :+1: Do you want me to have a closer look at something else beyond the aesthetics?

magne4000:
Not really, we can always go back to update the doc when needed :)

PR #2303, opened 2025-03-25
chore: remove vite as dependency

rtritto:
`vite` is a peer and dev dependency and can be removed from dependencies.


brillout:
https://github.com/vikejs/vike/pull/2070#issuecomment-2584209718

PR #2305, opened 2025-03-26
docs: document hono serve passed down options (vikejs/vike-server#88)

brillout:
@magne4000 Is that accurate? It's only for Hono right?


brillout:
I went ahead and merged, but let me know if it isn't accurate.

magne4000:
It's for all servers now

PR #2310, opened 2025-03-27
docs: add example of real application using Vike + SST and Hono

sbaidon:
## Overview

Added an example of a real app using Vike with SSR Streaming + SST + Hono

You can access the app at https://rentvsbuy.io/

I am happy to also create a dedicated starter template repository, would be good for someone to give my repo a quick look to make sure I am not doing anything too egregious


brillout:
:+1: Yea I think a minimal example would also be great.

PR #2313, opened 2025-03-27
chore: migrate to express 5 and use devDependencies

vanvuongngo:
- to increase the security
- reduce development dependencies in production


brillout:
Also, the CI is red, let's make it green.

brillout:
Another thing: `vike` isn't a dev dependency, it's a prod dependency for SSR apps.

brillout:
Seeing that:
image

This should be reverted.

brillout:
Let's close this PR in favor of separate PRs for each change:
 - One minimal PR for Express.js
 - Another one for devDependencies
 - ... more welcome

vanvuongngo:
> > app.get(['/', '/about', '/star-wars']
> 
> I don't think it's worth it: I understand it reduces the penetration surface, but only minimally so. The DX degradation, on the other hand, is significant.
> 
> Let's revert that and then merge.
> 
> Thank you for the PR 💚

Hi @brillout,

Thank you for your recview. A change is required here, as I pointed out in my discussion https://github.com/vikejs/vike/discussions/2301.
Express 5 no longer allows wildcards as a path like `app.get(‘*’)`, which reduces the DX and increases security, see Express Breaking Changes: https://expressjs.com/en/guide/migrating-5.html#path-syntax
If you can find a better approach with more DX, feel free to change this.

What is your concern about the `catalog:` workspace protocol? I have ruled this out for the boilerplates, and I don't have the time to maintain the dependency bumping effort with the current approach. The result of the current approach is that there are too many different versions of dependencies for the same package, which is difficult to maintain.

Please don't get me wrong, but the idea of a monorepo (workspace) is that you have all your packages/applications aligned and less to maintain.

brillout:
How about `app.get('/{*pages}'` ?

As for `catalog:` it's unfortunately actually more difficult to maintain for us (for many reasons).

vanvuongngo:
> How about `app.get('/{*pages}'` ?

would only work if you want to rename your routes like `/about-pages` and `/star-wars-pages`

> As for `catalog:` it's unfortunately actually more difficult to maintain for us (for many reasons).
Hm, so how do you maintain it by now? I see examples with react v18 and react v19. Do you have a migration script?

brillout:
It's just naming the wildcare and behaves like `/*`, see https://expressjs.com/en/guide/migrating-5.html#path-syntax.

I'd say let's go with `app.get('/{*vike-catch-all}'` to make it more explicit. Let's also add a comment as it's confusing for those unfamiliar with Express.js 5.

Let's first create a PR that only does the Express.js update, that's the most important for now. Let's then see if/what subsequent PR can make sense.

vanvuongngo:
> It's just naming the wildcare and behaves like `/*`, see https://expressjs.com/en/guide/migrating-5.html#path-syntax.
> 
> I'd say let's go with `app.get('/{*vike-catch-all}'` to make it more explicit. Let's also add a comment as it's confusing for those unfamiliar with Express.js 5.
> 
> Let's first create a PR that only does the Express.js update, that's the most important for now. Let's then see if/what subsequent PR can make sense.

Good catch, so the postfix is only a dummy text. I created a new PR https://github.com/vikejs/vike/pull/2317

PR #2347, opened 2025-04-05
Align right-side numbers in chip diagram

teddybradford:
Oops. I missed one of the styles in https://github.com/vikejs/vike/pull/2341


brillout:
:+1:

PR #2373, opened 2025-04-14
feat: `pageContext.runtime` (vike-server)

nitedani:
https://vike.dev/server#custom-pagecontext

>The runtime object is also available at pageContext.runtime.

### DIFF
--- a/vike/node/runtime/universal-middleware.ts
+++ b/vike/node/runtime/universal-middleware.ts
@@ -5,7 +5,13 @@ export default async function universalVikeHandler(
   context: Record<string, unknown>,
   runtime: Record<string, unknown>
 ) {
-  const pageContextInit = { ...context, ...runtime, runtime, urlOriginal: request.url, headersOriginal: request.headers }
+  const pageContextInit = {
+    ...context,
+    ...runtime,
+    runtime,
+    urlOriginal: request.url,
+    headersOriginal: request.headers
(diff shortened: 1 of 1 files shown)


brillout:
@magne4000 WDYT? LGTM :eyes:

brillout:
Maybe `{ ...runtime, runtime }` but I don't know.

magne4000:
> Maybe `{ ...runtime, runtime }` but I don't know.

We should probably do that instead yes, otherwise it would be a breaking change.

vike/node/runtime/universal-middleware.ts
```diff
@@ -5,7 +5,7 @@ export default async function universalVikeHandler(
   context: Record<string, unknown>,
   runtime: Record<string, unknown>
 ) {
-  const pageContextInit = { ...context, ...runtime, urlOriginal: request.url, headersOriginal: request.headers }
+  const pageContextInit = { ...context, runtime, urlOriginal: request.url, headersOriginal: request.headers }
```
brillout, line 8:
```suggestion
  const pageContextInit = { ...context, ...runtime, runtime, urlOriginal: request.url, headersOriginal: request.headers }
```

brillout:
@nitedani Little reminder: let's always prefix branch names `nitedani/ctxruntime`.

PR #2435, opened 2025-05-03
fix: do not close web stream prematurly

magne4000:
@brillout Vike wasn't using `doNotClose`, which is what caused [this issue](https://github.com/vikejs/vike-server/issues/119).
I tested the fix with Hono, Express and Fastify, they seem to always work as intended now.

Prior to this fix, calls to `injectToStream` would potentially be dropped.

### DIFF
--- a/vike/node/runtime/html/renderHtml.ts
+++ b/vike/node/runtime/html/renderHtml.ts
@@ -134,7 +134,12 @@ async function renderHtmlStream(
       return injectAtStreamAfterFirstChunk()
     }
   }
+  let resume = () => {}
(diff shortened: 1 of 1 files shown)


brillout:
We can definitely merge this (always good to have), but I wonder why and who prematurely closes the stream? Or is it just because `injectToStream()` is called too late after React's stream already ended?  In other words: is this a workaround or the real fix?

brillout:
The answer can be "we don't know" and we merge this while adding a comment.

magne4000:
> because injectToStream() is called too late after React's stream already ended
It's exactly that. `onBeforeEnd` is called before `injectToStream` have some time to be called.

Issue #1268, opened 2023-11-20
Flat `pageContext`, and `pageContext.definedAt`


brillout:
## Flat `pageContext`

Merge everything into `pageContext` while providing a `pageContext.from`:
  - `pageContext.data` as well as `pageContext.from.hooks.onBeforeRender[hookPath].data` (or without `hookPath`?)
  - `pageContex.config` as well as `pageContext.from.config.Page` (or `pageConext.from.config[configPath].Page`)?
  - `pageContext.locale` as well as `pageContext.from.failedRenderAttempts[0].hooks.onBeforeRender.locale` (see https://github.com/vikejs/vike/issues/1112)
  - `pageContext.pageId` as well as `pageContext.from.vike`

Providing the best of both worlds: a "think-free" `pageContext` that just works, while providing precise access over `pageContext.from`.

## `pageContext.definedAt`

```ts
// pageContext.definedAt: Record<string, DefinedAt>

// Corresponds to merging of Vike's internal filePathToShowToUser and fileExportPathToShowToUser
type DefinedAt = string // AFAICT making DefinedAt an object with more detailed information is useless.
                        // (Because AFAICT the only use cases for pageContext.definedAt are debugging
                        // and logging.)
```

Throw error when trying to read `pageContext.definedAt[prop]` for cumulative or computed values.

For cumulatives values or to retrieve overriden values: `pageContext.definedAtAll: Record<string, DefinedAt[]>`. Note that overriden values are usually not loaded so they'll usually be missing.

brillout:
Failed full-fledged implementation:
- https://github.com/vikejs/vike/pull/2458

PR #2458, opened 2025-05-15
[failed-attempt] flat pageContext (#1268)

brillout:
Using a proxy for an immutable `pageContext` data structure turns out to be more complex than I initially thought. Let's try again later, and let's go for something less clean but simpler in the meantime.


brillout:
[Reverted](https://github.com/vikejs/vike/commit/8fc54d6df29be3f20aedb5b8a9649bc0b2fbc0d7).

PR #2534, opened 2025-07-01
feat: allow getGlobalContextSync() to be used in production

brillout:


brillout:
@magne4000 Pre-released as `0.4.235-commit-16164c3` — let me know if you need a stable release.

Issue #2551, opened 2025-07-11
Session-scoped context for client-side persistence without pageContext.json requests


nitedani:
## Problem

In Vike applications, there's currently no built-in way to efficiently handle values that are:

1. Request-specific on the server (different for each user)
2. Session-persistent on the client (same across all navigations for a user)
3. Don't need to be re-fetched on each navigation

Vike has two primary context mechanisms:
- `globalContext`: Process-level on server, browser session-level on client
- `pageContext`: Request-specific on server, navigation-specific on client

When using `passToClient` to transfer server values to the client, Vike makes `pageContext.json` requests on every client-side navigation. This is unnecessary for values that are set once on the server and don't change between navigations (like tenant information, user preferences, etc.).

## Current Workaround

Using module-level variables in client-side `onCreatePageContext` hooks to capture the value during hydration and reuse it during navigation:

```typescript
// pages/+onCreatePageContext.client.ts
let tenant: string;

export async function onCreatePageContext(pageContext) {
  if (pageContext.isHydration) {
    tenant = pageContext.tenant;
  } else {
    pageContext.tenant = tenant;
  }
}

// pages/+onCreatePageContext.server.ts
export async function onCreatePageContext(pageContext) {
  pageContext.tenant = ((pageContext.runtime as any).req as Request).tenant;
}

declare global {
  namespace Vike {
    interface PageContext {
      tenant: string;
    }
  }
}
```

While this works, it:
- Requires manual state management
- Doesn't clearly communicate intent
- Isn't documented as an official pattern

## Feature Request

Add a mechanism for session-scoped context that:
1. Is request-specific on the server (like pageContext)
2. Persists across navigations on the client (like globalContext)
3. Is transferred once during initial page load
4. Doesn't trigger pageContext.json requests during navigation

This would provide a clean solution for values like tenant information, user preferences, and other session-level data that are determined server-side but should persist client-side without re-fetching.

nitedani:
It's not a big issue, it can be solved using a store or such on the client. But maybe Vike can have something documented about this pattern.

brillout:
I'd say that using `+onCreatePageContext` is a natural way to do it? FYI you can use the client-side `globalContext` instead of `globalThis`.

I agree it isn't entirely trivial, but I can't think of a simpler solution? (Also, long term the plan is that eventually most users will use Vike extenions that will do this on their behalf.)

[...]

PR #2557, opened 2025-07-15
document & test passToClient once trick (#2551)

brillout:


brillout:
@nitedani Pre-released in `0.4.235-commit-d2ac38a`. Docs: http://vike.dev/passToClient#once.

PR #2560, opened 2025-07-17
fix: eagerly set globalContext [enabling early `getGlobalContextSync()` calls]

brillout:


brillout:
@magne4000 Can you confirm this works? Pre-released as `0.4.235-commit-9b9289b`.

magne4000:
> @magne4000 Can you confirm this works? Pre-released as `0.4.235-commit-9b9289b`.

This indeed works :rocket:

PR #2584, opened 2025-07-29
fix: preserve server-computed urlParsed when urlLogical is used

### LINKED ISSUE #2580: `urlParsed` on client ignores `urlLogical` set on server
Hi!

According to the [docs](https://vike.dev/i18n), when using `urlLogical`, some of the url's sections can be removed from `pageContext.urlParsed` for routing (locale in this example):

> Vike's router will use `pageContext.urlLogical` instead of `pageContext.urlOriginal` and the locale is removed from `pageContext.urlParsed`.

The `passToClient` [documentation](https://vike.dev/passToClient) states that `pageContext.urlParsed` is available on the client-side by default when using Client Routing

**Issue:**

However, when using `urlLogical`, `pageContext.urlParsed` is **not** passed from server to client as expected. 

In my use case, I need to access the locale-agnostic URL on the client for navigation purposes. Since `urlParsed` is not available, the only workaround is to create a custom property (`truePath` in my repro) and manually pass the calculated route.

**Reproduction:**

https://stackblitz.com/edit/mqwex1cv-sbkivesx?file=pages%2F%2BonBeforeRoute.ts

Upon first load all values are consistent:
```
urlParsed: /
urlPathname: /
urlOriginal: /
truePath: /
```
But when a locale is added to the path (e.g. `/it`), only custom property shows locale-agnostic url:
```
urlParsed: /it
urlPathname: /it
urlOriginal: /it
truePath: /
```

naaa760:
### Problem
When using `urlLogical` in Vike, `pageContext.urlParsed` was not being properly passed from server to client. The client-side routing was recomputing `urlParsed` using `urlOriginal` (with locale) instead of preserving the server-computed logical URL (locale-agnostic).

### Solution
- Added `urlParsed` to built-in `passToClient` list for automatic client-side availability
- Modified `removeBuiltInOverrides` to preserve server-provided `urlParsed`
- Enhanced `getPageContextUrlComputed` to check for existing `urlParsed` before recomputing

### Changes
- `packages/vike/node/runtime/html/serializeContext.ts`: Add `urlParsed` to `passToClientBuiltInPageContext`
- `packages/vike/client/runtime-client-routing/getPageContext/removeBuiltInOverrides.ts`: Preserve `urlParsed` from server
- `packages/vike/shared/getPageContextUrlComputed.ts`: Check for existing `urlParsed` before recomputing

Fixes #2580

### DIFF
--- a/packages/vike/client/runtime-client-routing/getPageContext/removeBuiltInOverrides.ts
+++ b/packages/vike/client/runtime-client-routing/getPageContext/removeBuiltInOverrides.ts
@@ -2,7 +2,7 @@ export { removeBuiltInOverrides }
 
 import { assert, assertWarning } from '../utils.js'
 
-const BUILT_IN_CLIENT_ROUTER = ['urlPathname', 'urlParsed'] as const
+const BUILT_IN_CLIENT_ROUTER = ['urlPathname'] as const
 const BUILT_IN_CLIENT = ['Page', 'pageExports', 'exports'] as const
 type DeletedKeys = (typeof BUILT_IN_CLIENT)[number] | (typeof BUILT_IN_CLIENT_ROUTER)[number]
 function removeBuiltInOverrides(pageContext: Record<string, unknown> & { [key in DeletedKeys]?: never }) {
@@ -29,4 +29,8 @@ function removeBuiltInOverrides(pageContext: Record<string, unknown> & { [key in
       delete pageContext[prop]
     }
   })
+
+  // should be preserved instead of being recomputed on the client
+  if ('urlParsed' in pageContext) {
+  }
 }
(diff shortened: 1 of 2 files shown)


brillout:
I think a cleaner solution is to run `onBeforeRoute()` upon hydration. We then don't need to pass `pageContext.urlParsed` to the client (`urlParsed` is and should only be a computed property, so let's recompute it on the client-side instead of sending it).

The change isn't completely trivial so it's probably best I'll tackle this.

Contributing on other issues — see https://github.com/vikejs/vike/labels/contribution-welcome%20%3Atwo_hearts%3A — is very much welcome :green_heart:

nehaaprasad:
@brillout 
Okay!

PR #2592, opened 2025-07-30
fix: fix globalContext HMR caching behavior

brillout:


brillout:
Fix pre-released as `0.4.236-commit-79f1ef0`.

PR #2622, opened 2025-08-12
fix: clear `once` cache upon calling `reload()`

brillout:


brillout:
Follow up: https://github.com/vikejs/vike/pull/2632.

PR #2641, opened 2025-08-19
feat(core): default/clear

nitedani:


packages/vike/node/vite/shared/resolveVikeConfigInternal.ts
```diff
@@ -885,6 +885,7 @@ function getConfigValueSource(
       valueIsLoadedWithImport: false,
       valueIsDefinedByPlusValueFile: false,
       definedAt: definedAtFilePath,
+      cumulativeModifiers: computeCumulativeModifiersForSource(plusFile),
```
brillout: AFAICT we don't need `cumulativeModifiers`. How about we just call `applyCumulativeModifiers()` inside `getConfigValueSourcesRelevant()`?

nitedani: Okay, I'll try 😁

packages/vike/shared/page-configs/serialize/serializeConfigValues.ts
```diff
@@ -323,8 +323,13 @@ function getConfigValuesBase(
       } as const
       return { configValueBase, sourceRelevant: source, configName }
     } else {
-      const sourcesRelevant = getConfigValueSourcesRelevant(configName, runtimeEnv, pageConfig)
+      // Cumulative: gather all relevant sources then apply clear/default semantics
+      let sourcesRelevant = getConfigValueSourcesRelevant(configName, runtimeEnv, pageConfig)
+      if (sourcesRelevant.length === 0) return 'SKIP'
+
+      sourcesRelevant = applyCumulativeModifiers(sourcesRelevant, pageConfig, configName)
```
brillout: If we call applyCumulativeModifiers() inside getConfigValueSourcesRelevant() then it will also apply to other places where it needs to apply. I didn't check but IIRC simply moving applyCumulativeModifiers() inside getConfigValueSourcesRelevant() does the trick.

packages/vike/node/vite/shared/resolveVikeConfigInternal.ts
brillout: You mean augmented code will try :grin:

brillout:
Let's only support the feature over the file extensions for now (supporting the feature over `meta` isn't trivial and AFAICT we don't need it).

Since the test is green, this should almost be ready to be merged... I'll double check everything about this PR later today.

PR #2673, opened 2025-08-31
docs: JavaScript toggle

brillout:
See https://github.com/brillout/docpress/pull/47.


brillout:
@phonzammi I went ahead and used your branch to open a PR — let me know if you have a different plan.

phonzammi:
No problem, let me revert a few changes that aren't needed anymore.

phonzammi:
I've pushed some changes. Let me know if I missed anything.

brillout:
:eyes:  Let me check

brillout:
I think wrapping whole sections with TypeScriptOnly is a bit too disruptive (small parts are okay). Intervention++ :wink:

Some first lines need to have a line earlier:
image

docs/pages/auth/+Page.mdx
```diff
@@ -23,12 +23,13 @@ import type { PageContextServer } from 'vike/types'
 export async function onCreatePageContext(pageContext: PageContextServer) {
   // vike-server exposes the request object at pageContext.runtime.req
   const { req } = pageContext.runtime
-
+  let user: User
+  
   // Authentication middlewares (e.g. Passport.js and Grant) usually set req.user
   // to expose information about the logged-in user.
-  const { user } = req
+  user = req
```
brillout:
```suggestion
  user = req.user
```

docs/pages/broken-npm-package/+Page.mdx
```diff
@@ -166,7 +166,7 @@ Node.js v18.0.0
 
 For default exports, the workaround proposed by Node.js may not work and you may need to do this instead:
 
-```ts
+```ts ts-only
```
brillout:
```suggestion
```js
```

Intervention++

docs/pages/broken-npm-package/+Page.mdx
```diff
@@ -311,7 +311,7 @@ Use the component trace to find out which component is `undefined` / an `object`
 
 A local workaround is usually this:
 
-```ts
+```ts ts-only
```
brillout:
```suggestion
```js
```

docs/pages/config/+Page.mdx
```diff
@@ -295,7 +295,7 @@ Into:
 // Environment: config
 
 import type { Config } from 'vike/types'
-import Layout from '../layouts/LayoutDefault.jsx'// [!code --]
+//import Layout from '../layouts/LayoutDefault.jsx'// [!code --]
```
brillout: Hm, that's quite unfortunate. It's really misleading for the user. Ideas for solutions?

docs/pages/error-page/+Page.mdx
```diff
@@ -40,8 +40,15 @@ function Page() {
 
   return <p>{msg}</p>
 }
+```
+
+<TypescriptOnly>
```
brillout, line 51: I think it's safe to use it here, because the JavaScript toggle is shown in the code snippet right above.

docs/pages/getPageContext/+Page.mdx
```diff
@@ -8,7 +8,7 @@ The `getPageContext()` function allows you to access the <Link href="/pageContex
 For example:
 
 ```ts
- // /pages/some-page/+data.ts
+// /pages/some-page/+data.ts
```
brillout, line 11: Good catch :sparkles:

docs/pages/getPageContext/+Page.mdx
```diff
@@ -25,15 +25,15 @@ function data() {
...
 
 export { data }
 export type Data = Awaited<ReturnType<typeof data>>
 
 import { getPageContext } from 'vike/getPageContext'// [!code --]
 
-function data() {// [!code --]
-  const pageContext = getPageContext()// [!code --]
+// function data() {// [!code --]
```
brillout: Also very unfortunate here.

docs/pages/i18n/+Page.mdx
```diff
@@ -36,12 +36,12 @@ function extractLocale(url: Url) {
   // Determine the locale, for example:
   //  /en-US/film/42 => en-US
   //  /de-DE/film/42 => de-DE
-  const locale = /* ... */
+  const locale = `/* ... */`
```
brillout: I think we can find a solution for this as well?

docs/pages/migration/vike-cloudflare/+Page.mdx
```diff
@@ -51,7 +51,7 @@ If your are using <Link href="/hono" /> or <Link href="/hattip" />, some additio
 
     export default defineConfig({
       plugins: [
-        devServer(...),// [!code --]
+        devServer(/**...*/),// [!code --]
```
brillout:
```suggestion
        devServer(/*...*/),// [!code --]
```

docs/pages/error-page/+Page.mdx
phonzammi: The comment below doesn’t get removed when switching to JavaScript. I tested it again — comments only get removed when placed above a type or interface definition. 

```
// When using TypeScript you can define the type of `abortReason`
```

That’s why I split the code block.

brillout: Makes sense :+1:

docs/pages/error-page/+Page.mdx
```diff
@@ -40,8 +40,15 @@ function Page() {
...
+
+When using TypeScript you can define the type of `abortReason`
```
brillout, line 47:
```suggestion
When using TypeScript you can define the type of `abortReason`.
```

Intervention++

brillout:
```suggestion
You can define the type of `pageContext.abortReason`:
```

Intervention++ (the missing punctuation).

docs/pages/config/+Page.mdx
phonzammi: I couldn’t find any solution other than manually replacing all those code blocks with these:

~~~mdx
<CodeSnippets>
<CodeSnippet codeLang="ts">
```ts ts-only
// /pages/+config.ts
// Environment: config

import Layout from '../layouts/LayoutDefault.jsx'// [!code --]
const Layout = 'import:../layouts/LayoutDefault.jsx:default'// [!code ++]

export default {
  Layout
} satisfies Config
```
</CodeSnippet>
<CodeSnippet codeLang="js">
```js
// /pages/+config.js
// Environment: config

import Layout from '../layouts/LayoutDefault.jsx'// [!code --]
const Layout = 'import:../layouts/LayoutDefault.jsx:default'// [!code ++]

export default {
  Layout
}
```
</CodeSnippet>
</CodeSnippets>
~~~

brillout: Or we only show JS.

Or we implement some kind of magic comment.

brillout: For example:

```js
// @detype-temp-replace Layout Layout2
import Layout from '../layouts/LayoutDefault.jsx'// [!code --]
```

phonzammi: Alternatively, we can add a temporary (unique) comment:
```ts
import Layout from '../layouts/LayoutDefault.jsx'// [!code --]
//~const Layout = 'import:../layouts/LayoutDefault.jsx:default'// [!code ++]
```
Then, after running detype.transform() , we remove that comment.

brillout: Good idea, I like it

brillout: Although the advantage of `@detype-temp-replace` is that detype will still apply.

But `//~` is maybe easier to implement.

As you wish.

brillout:
```diff
- @detype-temp-replace
+ @detype-temp-rename
```

phonzammi: Yeah, if that line contains TypeScript code, then a temporary comment isn’t enough. Let's see.

brillout:
I resolved the merge conflicts, let me know if I made any mistake.

docs/pages/vike-server/+Page.mdx
```diff
@@ -198,7 +198,7 @@ In production run `$ NODE_ENV=production node dist/server/index.js` (or Bun/Deno
 
 ## `serve()` options
 
-```ts
+```ts ts-only
```
brillout:
```suggestion
```js
```
Shouldn't it be this?

phonzammi:
> I resolved the merge conflicts, let me know if I made any mistake.

Let me have a look

phonzammi:
I think for `<ConfigSpec>`'s children we can use `js` or `ts ts-only`. Wdyt ?

brillout:
Yes `ts ts-only`

brillout:
FYI it's also useful for JS users.

phonzammi:
Using `ts ts-only` won't hide the ts code block when users switch to js.
> `ts-only` sets an empty `style = {}`, making the code block always visible regardless of the selected language. This is useful when the code doesn’t require a select dropdown, but still want to wrap it with `<CodeSnippet>` component to keep the copy button functionality

brillout:
Yea I know and that's a good thing: the TypeScript type of a config is also helpful for JavaScript users. They are easy to understand also for non-TypeScript users.

phonzammi:
I've updated the `<ConfigSpec>` children code blocks: 
- added `ts-only` for multi-line blocks, and 
- changed `ts` to `js` for the `isTypeOneLiner` ones.

The reason for the second change is that the copy-to-clipboard button overlaps too much on single-line code blocks.
If we reduce the button's opacity and only show it on hover (within the code blocks), we can switch the `isTypeOneLiner` blocks back to ```ts ts-only.

brillout:
Makes sense!

docs/pages/getPageContext/+Page.mdx
phonzammi: There is another issue here.
Prettier and Babel (from detype) add a new line for a comment right after `{` when formatting multiline expressions
```diff
-- function data() {// [!code --]
++ function data() {
++    // [!code --]
```
For Prettier, we can add a `// prettier-ignore` comment above the line — though we'll need to remove it afterward.

For Babel, we have to set the `customizeBabelConfig(config)` detype prop like this:
```ts
customizeBabelConfig(config) {
   config.retainLines = false
},
```
But that ends up messing with the formatting in other places.

My solution is append this at the end of `processMagicComments()`: 
```ts
.replaceAll(/\n\s*\/\/\s\[!/g, '// [!')
```

Please, let me know what do you think.

phonzammi: Never mind, I'm pretty sure we can use `/* [!code --] */` and `/* [!code ++] */`, right ?

phonzammi: Wdyt : https://github.com/vikejs/vike/pull/2673/commits/a8957a2ff08f0e786281019fbe1b169742bbb5d8 ?

brillout: > Never mind, I'm pretty sure we can use `/* [!code --] */` and `/* [!code ++] */`, right ?

Nice, I didn't know we can do that :+1:

I also like the solutions you mentioned at https://github.com/vikejs/vike/pull/2673#discussion_r2328330300.

One thing I'm worried about the whole JS toggle is that it may complexify writing documentation. FYI Joël is starting to use DocPress for https://photonjs.dev. So it's somewhat of a priority to make using DocPress simple to use (also for contributors).

If there is a solution on DocPress's side, I'd be inclined to go for that instead of having to use workarounds while writing documentation.

Ideally, the JS toggle should be "free" (no DX degradation) from the perspective of the DocPress user.

If we can't quickly find a solution on DocPress's side then let's use user-land workaround for now and let's add it as a `TODO` in https://github.com/brillout/docpress/issues/48. Let's also add a `// TO-DO/javascript-toggle` comment each time we use the workaround (unless it's easy to find each place we use the workaround with a simple `grep` command).

FYI SvelteKit's docs also has a JavaScript toggle. Could be worth to see how they implemented it :eyes:

phonzammi: Those are totally true. Thanks for the reminder.

> FYI Joël is starting to use DocPress for https://photonjs.dev/. So it's somewhat of a priority to make using DocPress simple to use (also for contributors).

Yeah, I’ve checked out those docs. Let’s not complicate things for him or anyone else using DocPress just because of a new feature.

phonzammi: > FYI SvelteKit's docs also has a JavaScript toggle. Could be worth to see how they implemented it 👀

Thanks for the info. I went through parts of their [convert_to_ts](https://github.com/sveltejs/svelte.dev/blob/main/packages/site-kit/src/lib/markdown/renderer.ts#L403) code 🤯, they built it all from scratch, and it’s super impressive. Definitely a few ideas we can borrow for ours.

phonzammi: I'm sure this issue is resolved now.

brillout:
Is this ready for review as well?

phonzammi:
There are still '```'ts code blocks that don't contain TypeScript-specific code (e.g. [pages/api/+Page.mdx](https://github.com/vikejs/vike/blob/main/docs/pages/api/%2BPage.mdx). Do we want to add `ts-only` to all of them?

brillout:
How about we treat them as `ts-only` if detype's result doesn't make any difference?

phonzammi:
> How about we treat them as `ts-only` if detype's result doesn't make any difference?

Got it 👍

brillout:
(Small note: if the code highlighting is notably different for the same code but with TS highlting versus JS highlighting, maybe it could be worth it to still show the JS toggle. We can add it to the list of optimizations we want to do later, or do it now if it's the case and easy to implement.)

phonzammi:
As far as I can tell, if I’m understanding you right, the highlighting looks the same for both.

phonzammi:
I've added `ts-only` to all code blocks where detype's output doesn't change anything or when the code is TypeScript-only. 

I'm leaving out `ts-only` for blocks that include file paths like // /path/to/file.ts, even if detype's output doesn't change anything.

I haven't updated the `yaml` code block yet.

brillout:
I was thinking more of automatically treating any ts block as `ts-only` (without having to explicitly mark them as such) if detype (and the `.js` => `.ts` transformation) doesn't produce any difference. So we still run detype, and then check whether the result is equal to the original. WDYT?

phonzammi:
Yeah, we can do that — just need to ensure the formatting is consistent.

phonzammi:
I'm happy with this PR now—let me know if I missed anything.

brillout:
The layout is a bit off:

image

Compare this to:

image

brillout:
> The layout is a bit off:

Let's fix this and then we can merge this! Exciting :sparkles:

phonzammi:
Let me have a look, I think we need to modify the styling a bit

phonzammi:
> Let me have a look, I think we need to modify the styling a bit

Done : https://github.com/vikejs/vike/pull/2673/commits/7208806fe8945e55abc3387fc7708ae94d7bb7a2

What do you think, LGTM ?

brillout:
Perfect :sparkles: 

Merged :heart_eyes:

PR #2682, opened 2025-09-03
fix: ensure proper replacement of __VIKE__IS_NON_RUNNABLE_DEV

magne4000:
While working on Photon + Cloudflare + Vike, I noticed that Cloudflare can append hash IDs to certain files, which sometimes prevents them from matching in `vike:pluginNonRunnableDev` hook.

```js
{
  id: '/home/magne/workspace/vike-server/node_modules/.pnpm/vike@0.4.238_react-streaming@0.4.3_react-dom@19.1.1_react@19.1.1__react@19.1.1__vite@7._26f16f1ae2bc3a83a214ce6877b50b7a/node_modules/vike/dist/esm/utils/isNonRunnableDev.js?v=e9453e59',
  distFileIsNonRunnableDev: '/home/magne/workspace/vike-server/node_modules/.pnpm/vike@0.4.238_react-streaming@0.4.3_react-dom@19.1.1_react@19.1.1__react@19.1.1__vite@7._26f16f1ae2bc3a83a214ce6877b50b7a/node_modules/vike/dist/esm/utils/isNonRunnableDev.js'
}
```

### DIFF
--- a/packages/vike/node/vite/plugins/pluginNonRunnableDev.ts
+++ b/packages/vike/node/vite/plugins/pluginNonRunnableDev.ts
@@ -46,13 +46,14 @@ function pluginNonRunnableDev(): Plugin {
     },
     transform(code, id) {
       if (!config._isDev) return
-      if (id !== distFileIsNonRunnableDev && id !== distFileGlobalContext) return
+      const idWithoutHash = id.split('?')[0]!;
+      if (idWithoutHash !== distFileIsNonRunnableDev && idWithoutHash !== distFileGlobalContext) return
       if (isRunnableDevEnvironment(this.environment)) return
       const { magicString, getMagicStringResult } = getMagicString(code, id)
-      if (id === distFileIsNonRunnableDev) {
+      if (idWithoutHash === distFileIsNonRunnableDev) {
         magicString.replaceAll('__VIKE__IS_NON_RUNNABLE_DEV', JSON.stringify(true))
       }
-      if (id === distFileGlobalContext) {
+      if (idWithoutHash === distFileGlobalContext) {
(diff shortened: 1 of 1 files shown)


magne4000:
I'm wondering, couldn't we just do something like that?
```js
if (code.includes('__VIKE__IS_NON_RUNNABLE_DEV')) {
 ...
}
```

brillout:
It would work yes, but I ain't sure about the performance implications. I don't know how fast JS is with testing big strings, but I'm inclined to think it isn't negligible given that it will do this for the entire code base (including all node_modules imports with full-bundle Rolldown mode).

brillout:
Pre-released as `vike@0.4.238-commit-d48a597`

magne4000:
There is a new [filter](https://vite.dev/guide/rolldown.html#hook-filter-feature) option on some hooks that would probably nullify the performance impact on latest Vite versions:
```js
const plugin = {
  name: 'plugin',
  transform(code, id): {
    filter: {
      // First pass is fast, few IDs should match this anyway
      id: /isNonRunnableDev/,
      // Second pass matches the code, to filter only the right ones
      code: /__VIKE__IS_NON_RUNNABLE_DEV/gm,
    },
    handler() { ... }
  },
}
```

brillout:
Neat, good to know. I'll eventually start using it.

Although, I'm still not sure how fast it is to apply a regex in Rust to the entire code base.

brillout:
I'm guessing that hook filtering will be important for Rolldown('s full-bundle mode). So I went ahead and added "static filtering" (mostly hook filters but also `applyToEnvironment` filtering) to all Vike's hooks:
- https://github.com/vikejs/vike/pull/2688

PR #2692, opened 2025-09-04
feat: `globalContext.prerenderContext`

brillout:


brillout:
Pre-released as `vike@0.4.239-commit-33e55d4`.

PR #2737, opened 2025-09-18
Magne4000/dev

magne4000:
### TODO
- [ ] migration doc from `vike-server` to `vike-photon`
- [ ] update vercel doc
- [ ] update cloudflare doc

### DIFF
--- a/docs/pages/cloudflare/+Page.mdx
+++ b/docs/pages/cloudflare/+Page.mdx
@@ -12,13 +12,14 @@ We recommend using `vike-cloudflare` to deploy your app to [Cloudflare](https://
 
 > <UseScaffolder>`vike-cloudflare`</UseScaffolder>
 
+TODO migrate this doc to photon
 
 ## `vike-cloudflare`
 
-Version history: [`CHANGELOG.md`](https://github.com/vikejs/vike-cloudflare/blob/main/CHANGELOG.md)  
-Examples: <Link href="/new">vike.dev/new</Link> and [`examples/`](https://github.com/vikejs/vike-cloudflare/tree/main/examples)  
-Source code: [GitHub > `vikejs/vike-cloudflare`](https://github.com/vikejs/vike-cloudflare)  
-npm package: [`vike-cloudflare`](https://www.npmjs.com/package/vike-cloudflare)  
+Version history: [`CHANGELOG.md`](https://github.com/vikejs/vike-cloudflare/blob/main/CHANGELOG.md)
+Examples: <Link href="/new">vike.dev/new</Link> and [`examples/`](https://github.com/vikejs/vike-cloudflare/tree/main/examples)
+Source code: [GitHub > `vikejs/vike-cloudflare`](https://github.com/vikejs/vike-cloudflare)
+npm package: [`vike-cloudflare`](https://www.npmjs.com/package/vike-cloudflare)
 
 You can use Vike with the `vike-cloudflare` extension for a zero-config Cloudflare deployment of your SSR or pre-rendered (aka SSG) app.
 
--- a/docs/headings.ts
+++ b/docs/headings.ts
@@ -643,8 +643,8 @@ const headings = [
   },
   {
     level: 2,
-    title: '`+server`',
-    url: '/server',
+    title: '`+photon`',
+    url: '/photon',
   },
   {
     level: 2,
@@ -997,12 +997,6 @@ function extensions() {
       {
         title: '`vike-solid`',
         url: '/vike-solid',
-      },
-      {
-        title: '`vike-server`',
-        url: '/vike-server',
-        sectionTitles: ['HTTPS'],
-        category: 'Overview',
       },
       {
         title: '`vike-photon`',
@@ -1496,6 +1490,16 @@ function deprecated() {
       {
         title: '`_default.page.route.js`',
         url: '/_default.page.route.js',
+      },
+      {
+        title: '`+server.js`',
+        url: '/server',
+      },
+      {
+        title: '`vike-server`',
+        url: '/vike-server',
+        sectionTitles: ['HTTPS'],
+        category: 'Overview',
       },
     ] as const
   ).map((h) => ({ ...h, category: 'Deprecated' as const })) satisfies HeadingDetachedDefinition[]
--- a/docs/headings.ts
+++ b/docs/headings.ts
@@ -1004,6 +1004,12 @@ function extensions() {
         sectionTitles: ['HTTPS'],
         category: 'Overview',
       },
+      {
+        title: '`vike-photon`',
+        url: '/vike-photon',
+        sectionTitles: ['HTTPS'],
+        category: 'Overview',
+      },
     ] as const
   ).map((h) => ({ ...h, category: 'Extensions' as const })) satisfies HeadingDetachedDefinition[]
 }
--- a/docs/pages/add/+Page.mdx
+++ b/docs/pages/add/+Page.mdx
@@ -39,7 +39,7 @@ On a high-level, this is how you add Vike to your existing Vite app:
 1. Either:
    - Enable pre-rendering, or
      > See <Link href="/pre-rendering" />
-   - add a <Link href="/express">Express.js</Link>/<Link href="/hono">Hono</Link>/<Link href="/fastify">Fastify</Link>/<Link href="/vike-server">...</Link> server (or add <Link href="/renderPage">Vike's server middleware</Link> if you already have one).
+   - add a <Link href="/express">Express.js</Link>/<Link href="/hono">Hono</Link>/<Link href="/fastify">Fastify</Link>/<Link href="/vike-photon">...</Link> server (or add <Link href="/renderPage">Vike's server middleware</Link> if you already have one).
      > Example: <RepoLink path="/packages/create-vike-core/boilerplate-react/server/index.js" />
 
 1. Either:
--- a/docs/pages/api-routes/+Page.mdx
+++ b/docs/pages/api-routes/+Page.mdx
@@ -16,7 +16,7 @@ To achieve a similar DX as API routes, you can use a server like [Express.js](ht
 
 >  Make sure <Link href="/renderPage">Vike's middleware</Link> is your last middleware. (It's a catch-all middleware so it would override your API routes.)
 
->  If you don't use <Link href="/vike-server">`vike-server`</Link>, then you may need to manually restart your server for changes to take effect.
+>  If you don't use <Link href="/vike-photon">`vike-photon`</Link>, then you may need to manually restart your server for changes to take effect.
 
 
 ## See also
--- a/docs/pages/auth/+Page.mdx
+++ b/docs/pages/auth/+Page.mdx
@@ -21,7 +21,7 @@ import type { PageContextServer } from 'vike/types'
 
 // This hook is called upon new incoming HTTP requests
 export async function onCreatePageContext(pageContext: PageContextServer) {
-  // vike-server exposes the request object at pageContext.runtime.req
+  // vike-photon exposes the request object at pageContext.runtime.req
   const { req } = pageContext.runtime
 
   // Authentication middlewares (e.g. Passport.js and Grant) usually set req.user
@@ -65,9 +65,9 @@ export default {
 
 > You can access `pageContext` in any UI component and Vike hook — see <Link href="/pageContext" />.
 
-> If you use <Link href="/vike-server">`vike-server`</Link>, you can access the request object at `pageContext.runtime.req.user` (see <Link href="/pageContext#runtime">`pageContext.runtime`</Link>).
+> If you use <Link href="/vike-photon">`vike-photon`</Link>, you can access the request object at `pageContext.runtime.req.user` (see <Link href="/pageContext#runtime">`pageContext.runtime`</Link>).
 >
-> If you don't use `vike-server`, you can define a custom property `pageContext.req` at <Link href="/renderPage">`renderPage()`</Link>:
+> If you don't use `vike-photon`, you can define a custom property `pageContext.req` at <Link href="/renderPage">`renderPage()`</Link>:
 > ```ts
 > // server/index.ts
 >
--- a/docs/pages/createDevMiddleware/+Page.mdx
+++ b/docs/pages/createDevMiddleware/+Page.mdx
@@ -2,7 +2,7 @@ import { Link } from '@brillout/docpress'
 
 Create server middleware for development with <Link href="/glossary#hmr">HMR</Link> and <Link href="/lazy-transpiling">lazy transpiling</Link>.
 
-> You don't need `createDevMiddleware()` if you use <Link href="/vike-server">`vike-server`</Link>. (You also don't need it if you <Link href="/pre-rendering">pre-render</Link> and don't use a server at all.)
+> You don't need `createDevMiddleware()` if you use <Link href="/vike-photon">`vike-photon`</Link>. (You also don't need it if you <Link href="/pre-rendering">pre-render</Link> and don't use a server at all.)
 
 ```ts
 import { createDevMiddleware } from 'vike/server'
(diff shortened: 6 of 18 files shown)


brillout:
FYI +redirects.ts

magne4000:
Once we're OK with the doc and Bati is merged+released, we can then merge this PR (keeping it as Draft PR until Bati is released).

brillout:
Niceeee, I'll review later today

docs/pages/vercel/+Page.mdx
```diff
@@ -1,14 +1,75 @@
...
+```
+
+```json
+// package.json
+
+{
+  "scripts": {
+    "dev": "vike dev",
+    "preview": "vike build && vite preview",// [!code ++]
+    "deploy": "vike build && wrangler deploy"// [!code ++]
```
brillout: Typo, I don't know what the right commend is.

docs/pages/cloudflare/+Page.mdx
```diff
@@ -36,17 +18,29 @@ See:
   </TabList>
   <TabPanel>
     ```shell
-    npm i vike-cloudflare wrangler cross-env
+    npm i vike-photon @photonjs/cloudflare wrangler
+    # (optional) you can also install a server compatible with Cloudflare
```
brillout: Not sure what this comment means

docs/pages/cloudflare/+Page.mdx
```diff
@@ -4,29 +4,11 @@ import { Tab, Tabs, TabList, TabPanel } from 'react-tabs'
...
-Source code: [GitHub > `vikejs/vike-cloudflare`](https://github.com/vikejs/vike-cloudflare)  
-npm package: [`vike-cloudflare`](https://www.npmjs.com/package/vike-cloudflare)  
-
-You can use Vike with the `vike-cloudflare` extension for a zero-config Cloudflare deployment of your SSR or pre-rendered (aka SSG) app.
-
-See:
- - <Link href="#with-server" /> if you use <Link href="/ssr">SSR</Link> (it requires a server).
- - <Link href="#without-server" /> if you <Link href="/pre-rendering">pre-render</Link> all your pages (you don't need a server then).
-
-### With server
```
brillout, line 29: :+1: We can remove the two sections `### With server` and `### Without server` but let's improve the flow. It's a bit messy and incomplete (e.g. the deploy script is missing). It should be clear what should be done, depending on whether the user wants to SSG or SSR.

brillout: Idea: maybe we can move the server related stuff as a new section `SSR`.

docs/pages/cloudflare/+Page.mdx
```diff
@@ -4,29 +4,11 @@ import { Tab, Tabs, TabList, TabPanel } from 'react-tabs'
...
-
-
-## `vike-cloudflare`
-
-Version history: [`CHANGELOG.md`](https://github.com/vikejs/vike-cloudflare/blob/main/CHANGELOG.md)  
-Examples: <Link href="/new">vike.dev/new</Link> and [`examples/`](https://github.com/vikejs/vike-cloudflare/tree/main/examples)  
-Source code: [GitHub > `vikejs/vike-cloudflare`](https://github.com/vikejs/vike-cloudflare)  
-npm package: [`vike-cloudflare`](https://www.npmjs.com/package/vike-cloudflare)  
-
-You can use Vike with the `vike-cloudflare` extension for a zero-config Cloudflare deployment of your SSR or pre-rendered (aka SSG) app.
```
brillout, line 23: I think it makes sense to mention that. Not sure if it's obvious to users that Cloudflare also supports static sites.

brillout:
@phonzammi As expected there are some merge conflicts, can you have a look at  `git fetch && git merge origin/main`?

phonzammi:
Sure, I'll have a look now

phonzammi:
I've resolved the merge conflicts, what's left is fixing the incorrect code block syntax error.

docs/pages/cloudflare/+Page.mdx
magne4000: Comment updated to be more straightforward

docs/pages/cloudflare/+Page.mdx
magne4000: I added a section dedicated to server configuration

brillout:
## Docs

- At http://localhost:3000/migration/vike-photon we should add how to migrate from vite-plugin-vercel and vike-cloudflare.
- Is e54be753e78d8c3e4f974e36e66bb49444864728 (follow up of e54be753e78d8c3e4f974e36e66bb49444864728) accurate?
- Is d863a3661013b362fa73633ba900573a5a5e9f84 correct?
- At http://localhost:3000/cloudflare#photon I don't think `"preview": "vike build && vike preview"` makes sense?
- Is `"compatibility_flags": ["nodejs_compat"]` always required? With Cloudflare Pages it wasn't always required.
- Is a90ed658a049789dafc0cfb710352270103e9a9c correct?
- ae9d284e9fc5e02e1c797777c2312e69b842d288
- Are Vercel Edge Middlewares still a thing? I ain't sure, but I vaguely remember that they deprecated (some of) their edge deployment. If it's still at think then let's add a link to some official Edge Middlewares documentation.
- 8efe3852e132eab0c49f64b6970445f84bdc19b8

## API

- How about we rename `onServer()` to `onCreate()`? It fits better the name of other hook `onReady()`? Now is a good time to do such change I guess?
- Can we have `apply` and `serve` be also exported from a single path `vike-cloudflare/hono`? Even better: I wonder why it needs to be imported from `vike-cloudflare` instead of `vike-photon` and whether this can be avoided.
  - Actually, ideally, from `@photonjs/hono`.
- Maybe we should revert the Bati PR. Dependening on whether we want to introduce the aforementioned breaking changes, and dependening on whether we can quickly introduce those breaking changes.

brillout:
- We should mention it's beta.
- What's missing to move Photon from alpha to beta? Only the docs that are little bit chaotic I'd say?

brillout:
I guess the API part is the most urgent, unless we revert the Bati PR then it's less urgent.

We can work on the Photon docs after merging this PR but before I publish the blog post and before I reach out to deployment providers.

magne4000:
> At http://localhost:3000/migration/vike-photon we should add how to migrate from vite-plugin-vercel and vike-cloudflare.

- [x] :+1: 

> Is [e54be75](https://github.com/vikejs/vike/commit/e54be753e78d8c3e4f974e36e66bb49444864728) (follow up of [e54be75](https://github.com/vikejs/vike/commit/e54be753e78d8c3e4f974e36e66bb49444864728)) accurate?

Yes

> Is [d863a36](https://github.com/vikejs/vike/commit/d863a3661013b362fa73633ba900573a5a5e9f84) correct?

Yes

> At http://localhost:3000/cloudflare#photon I don't think `"preview": "vike build && vike preview"` makes sense?

- [x] Indeed, but `vite preview` is supported by `@cloudflare/vite-plugin`

> Is `"compatibility_flags": ["nodejs_compat"]` always required? With Cloudflare Pages it wasn't always required.

~~Yes it is, at least for now (only during dev and build time).~~
- [x] Will be fixed in latest Photon release

> Is [a90ed65](https://github.com/vikejs/vike/commit/a90ed658a049789dafc0cfb710352270103e9a9c) correct?

Yes :+1:

> Are Vercel Edge Middlewares still a thing? I ain't sure, but I vaguely remember that they deprecated (some of) their edge deployment. If it's still at think then let's add a link to some official Edge Middlewares documentation.

~~They [still exist](https://vercel.com/docs/routing-middleware/api), but I rarely see them advertised.~~
- [x] https://github.com/vikejs/vike/pull/2737#issuecomment-3414577608

> [8efe385](https://github.com/vikejs/vike/commit/8efe3852e132eab0c49f64b6970445f84bdc19b8)

- [x] :+1:

magne4000:
> How about we rename `onServer()` to `onCreate()`? It fits better the name of other hook `onReady()`? Now is a good time to do such change I guess?

- [x] It has already been renamed `onCreate`, I just missed it while updating the documentation :sweat_smile: 

> Can we have `apply` and `serve` be also exported from a single path `vike-cloudflare/hono`? Even better: I wonder why it needs to be imported from `vike-cloudflare` instead of `vike-photon` and whether this can be avoided.
>> Actually, ideally, from `@photonjs/hono`

- [x] Same thing, `vike-cloudflare` isn't relevant anymore, and it should indeed come from `@photonjs/hono` now.

brillout:
> Yes it is, at least for now (only during dev and build time).

I'm a bit surprised it's now required? Is that related to Cloudflare Workers?

magne4000:
> We should mention it's beta

- [x] :+1:

> What's missing to move Photon from alpha to beta? Only the docs that are little bit chaotic I'd say?

Yes, a better documentation, and I would perhaps wait to see where we're going with it also.

brillout:
> I just missed it while updating the documentation 😅

Let's double and triple check next time, there were a *lot* of typos and minor mistakes (most of them I guessed, but still very much time consuming for the reviewer overall).

brillout:
> I would perhaps wait to see where we're going with it also.

I'd argue that, it itself, Photon is beta. Whether/how we can merge things with Nitro is a "separate" topic and, more importantly, this isn't going to happen anytime soon.

magne4000:
> > Yes it is, at least for now (only during dev and build time).
> 
> I'm a bit surprised it's now required? Is that related to Cloudflare Workers?

I was about to comment that `wrangler` does some weird static analysis during build and sees side-effects that are not runtime related, but this seems to be fixed now.
So let me try a few things again, and perhaps I can remove it.

brillout:
> So let me try a few things again, and perhaps I can remove it.

Neat, users will appreciate the lightweight impression it gives :sparkles:

brillout:
How about this?

```diff
- photon:cloudflare:photon:server-entry
+ virtual:photon:cloudflare:server-entry
```

If ain't mistaken (correct me if I'm wrong) virtual modules should start with `virtual:` (I don't mind if some internals are dirty, but we should at least present something clean).

magne4000:
> How about this?
> 
> ```diff
> - photon:cloudflare:photon:server-entry
> + virtual:photon:cloudflare:server-entry
> ```
> 
> If ain't mistaken (correct me if I'm wrong) virtual modules should start with `virtual:` (I don't mind if some internals are dirty, but we should at least present something clean).

The only actual convention is for private virtual module that should start with `\0virtual`.
Photon module are not private (they need to be read and manipulated by multiple plugins), and when they appear in stack traces, having `virtual:photon` just bloats the output IMO, so that's why I went for `photon:` instead. So I still prefer using just `photon:`.

magne4000:
> I'd argue that, it itself, Photon is beta

I mean yes, we can say that it entered the beta this week

brillout:
> The only actual convention is for private virtual module that should start with `\0virtual`. Photon module are not private (they need to be read and manipulated by multiple plugins), and when they appear in stack traces, having `virtual:photon` just bloats the output IMO, so that's why I went for `photon:` instead. So I still prefer using just `photon:`.

Sounds good :ok_hand: Can we remove the redudant `:photon:` ?

```diff
- photon:cloudflare:photon:server-entry
+ photon:cloudflare:server-entry
```

Looks very clean then.

brillout:
> Virtual modules in Vite (and Rollup) are prefixed with `virtual:` for the user-facing path by convention

Quoting https://vite.dev/guide/api-plugin.html#virtual-modules-convention

The way I interpret this is the opposite: the user-facing modules should be prefixed with `virtual:`. One benefit of showing `virtual:` to users is that it makes it clearer what it is. I'm inclined to prefer clarity over being too succinct in this case.

magne4000:
> > Virtual modules in Vite (and Rollup) are prefixed with `virtual:` for the user-facing path by convention
> 
> Quoting https://vite.dev/guide/api-plugin.html#virtual-modules-convention
> 
> The way I interpret this is the opposite: the user-facing modules should be prefixed with `virtual:`. One benefit of showing `virtual:` to users is that it makes it clearer what it is. I'm inclined to prefer clarity over being too succinct in this case.

:thinking: That's a fair point. I changed back and forth like 3 or 4 times between the 2, so if I go back to `virtual:photon:`, we must be confident with it.



> Sounds good 👌 Can we remove the redudant `:photon:` ?
> 
> ```diff
> - photon:cloudflare:photon:server-entry
> + photon:cloudflare:server-entry
> ```
> 
> Looks very clean then.

Looks cleaner indeed, but it would break the module resolution with the way it currently works., let me explain:
`photon:cloudflare` is resolved by `@photonjs/cloudflare` plugin, and `@photonjs/core` is then able to resolve `photon:server-entry`. I have no "simple" way around it, and it makes thing easy to work with.
So with `virtual:` it would look like: `virtual:photon:cloudflare:virtual:photon:server-entry`.

brillout:
Follow up discussion about virtual ID naming: https://github.com/photon-js/photon/issues/24.

brillout:
> vike build && vite preview

Hm, that's quite dirty. Let me check if I can make it work with Vike's CLI.

magne4000:
> Are Vercel Edge Middlewares still a thing? I ain't sure, but I vaguely remember that they deprecated (some of) their edge deployment. If it's still at think then let's add a link to some official Edge Middlewares documentation.
>> They [still exist](https://vercel.com/docs/routing-middleware/api), but I rarely see them advertised.

[Edge Runtime](https://vercel.com/docs/functions/runtimes/edge) seems to not be recommended anymore though. And Edge Middlewares is far too niche to be referenced as a top feature.
Perhaps we should just mention ISR?

EDIT: https://github.com/vikejs/vike/pull/2737/commits/9f8f79a0e25f037075f0a8d9df7fe6cff499c485

brillout:
> Hm, that's quite dirty. Let me check if I can make it work with Vike's CLI.

How about a new Vike setting `+cli.preview` with `boolean | 'vite'`? We can then also show a assertUsage() error when `false` (for Vercel).

magne4000:
> > Hm, that's quite dirty. Let me check if I can make it work with Vike's CLI.
> 
> How about a new Vike setting `+cli.preview` with `boolean | 'vite'`? We can then also show a assertUsage() error when `false` (for Vercel).

I'm wondering, what's preventing Vike from working with standard `vite preview` (and so `vike preview` becomes a passthrough)? `@cloudflare/vite-plugin` seems to have made it work, perhaps we should focus on that instead?

brillout:
Because, in order to simulate self-hosting, `$ vike preview` runs `node ./dist/server/index.mjs` instead of spinning up Vite's preview server. That's how we made `$ vike preview` work for `vike-server`.

If `+vite.preview === 'vite'` => spin up Vite's preview server (equivalent to `$ vite preview)`.  
If `+vite.preview === true` => `dist/server/`.  
If `+vite.preview === false` => error.  
If `+vite.preview === undefined` => current behavior (tries to guess which one to do).

magne4000:
:thinking: I was thinking that when we are using `vike-photon`, we can always retrieve the underlying `httpServer` running the user code, but that doesn't fix the issue when running the preview without `vike-photon`. So I guess that a config is indeed required.

brillout:
@magne4000 There are two Elysia installation instrucions. Which one is accurate?

```js
  const app = new Elysia()
  apply(app)
  return serve(app)
```

```js
  const app = apply(new Elysia())
  return serve(app)
```

brillout:
@magne4000 861c7fba1188060a1203eceff433f140a15c7ed4 — quickly looking at the links I didn't see the added value

brillout:
@magne4000 It's live, if you can double check everything is accurate that'd be great.

https://vike.dev/vike-photon
https://vike.dev/vercel
https://vike.dev/cloudflare
https://vike.dev/photon

Also: 861c7fba1188060a1203eceff433f140a15c7ed4 — quickly looking at the links I didn't see the added value.

magne4000:
> ```js
>   const app = new Elysia()
>   apply(app)
>   return serve(app)
> ```
> 
> ```js
>   const app = apply(new Elysia())
>   return serve(app)
> ```

Both are correct.

PR #2738, opened 2025-09-21
docs: Document how to add a page after configuring vike-react

louwers:
For me this was not immediately clear, so would be good to have this information on this page. I think it is an essential part of getting started with `vike-react`.


brillout:
Makes sense, thank you!

I've a coupe of ideas for improving https://vike.dev/add — including making it easier to consume. I'll do it once I focus on writing new guides for newcomers.

PR #2749, opened 2025-09-28
add deno examples

richard-unterberg:
just added a contemporary example for the usage with deno, since the others were a bit outdated. I'm also open to change content or location of the demo package. let me know!


richard-unterberg:
... I will provide a full doc page with explanation and code examples.

brillout:
Nice! Feel free to re-open whenever you feel it's ready and I'll merge!

Issue #95, opened 2021-06-13
Client-side Routing without hitting the server


brillout:
Currently, when using Client-Side Routing (see docs [here](https://vite-plugin-ssr.com/SR-vs-CR) and [there](https://vite-plugin-ssr.com/useClientRouter)), the client-side `pageContext` is fetched from the server.

Some people want to be able to navigate between pages without hitting the Node.js SSR server. I'm opening a ticket for dicussing this.

I'm willing to implement this but this will require a couple of days work, so don't expect this too soon.

👍 this ticket if you want this. The more 👍 this ticket collects the more will I increase the importance of this. And better yet, add a comment elaborating why this is crucial for you and I'll then also increase urgency.

**Edit: Most of the time, this is only about (negligible) server cost servings**. For most apps, performance and cost implications are negligible. (Note that the static middleware of your Node.js server always needs to be hit anyways in order to get the latest code of the new page. It's mostly only about reducing Node.js work load.)

brillout:
>  I have no "loading" indicators while it is being loaded

Check out `onTransitionStart` and `onTransitionEnd` decribed at https://github.com/brillout/vite-plugin-ssr#import--useclientrouter--from-vite-plugin-ssrclientrouter.

> I'd prefer this behavior only on the first page load, after page is loaded, it's better to use client-only routing without hitting the server at all. This will allow much smoother UX

Is your GraphQL server physically closer to your user than to your SSR server? 

> and it greatly reduces load of SSR server.

Yes you are right, with the current `vite-plugin-ssr` design, if your database doesn't live on the SSR server then the load for the SSR server is non-negligibly higher.

Dema:
Yeah.. Is it closer, but the issue is that it will "freeze" the application while transitioning pages. I think, in the era of SPA's users are accustomed to instantaneous page transitions. I more like the idea behind Suspense, when I can delay loading state but still show it if it takes too long. I really like how next.js handles routing — after first page load it starts to use pure client-side routing.

It actually is less about load of the SSR server and more about user experience. For example, let's take CRUD, user is on the list page and clicks on one of the  items to drop into details, as the list is already loaded into the graphql client's cache, actual useQuery would not go to grab data from back-end and can just instantaneously show details page, maybe while showing some spinners for parts that are not in the cache and still loading. But user can already interact with the page. This way SSR is perceived slower than the pure client queries. 

I think it would be really great if you can expose full routes in AST form so we can create compilers for each specific routing library.

[...]

PR #2772, opened 2025-10-09
Update docs for change of vike-server/serve() options

kairi003:
Reflects the fix from vike-server [v1.0.10](https://github.com/vikejs/vike-server/compare/v1.0.9...v1.0.10) (fix: rename onServer hook to onCreate #95)


brillout:
Thank you!

PR #2799, opened 2025-10-27
doc: new `photon.static` and `photon.compress` options

magne4000:
See https://github.com/vikejs/vike-photon/pull/14

### DIFF
--- a/docs/pages/vike-photon/+Page.mdx
+++ b/docs/pages/vike-photon/+Page.mdx
@@ -544,7 +544,7 @@ If HMR isn't what you want (for example if you modify the database connection) y
 
 ## Compression
 
-In production, `vike-photon` compresses all Vike responses.
+In production, Photon compresses all Vike responses.
 
 You can disable it, or enable it only for static files:
 
@@ -567,7 +567,9 @@ export default {
 
 ## Static files
 
-In production, `vike-photon` uses [@universal-middleware/sirv](https://universal-middleware.dev/middlewares/sirv) to serve static files.
+In production, Photon uses [`@universal-middleware/sirv`](https://universal-middleware.dev/middlewares/sirv) to serve static files.
+
+> In development, static files are served by [Vite](https://vitejs.dev)'s development server.
 
 You can disable it, or set options:
 


brillout:
Is it a breaking change? Maybe instead a deprecation warning, and a `CHANGELOG.md` entry (which is need anyways)?

magne4000:
It's not a breaking change, only a deprecation with a warning.

brillout:
How about just a deprecation warning with a link to the new docs?

magne4000:
That's what it does (Or I'm not understanding what you mean).

brillout:
My concern is that it isn't sustainable to add a lot of deprecation warnings to the docs — i.e. docs should be as minimal as possible.

magne4000:
https://github.com/vikejs/vike-photon/blob/ebabdeaad06b60f2ac1c8eb799288363cfd37de6/packages/vike-photon/src/middleware/middlewares/compress.ts#L43

magne4000:
Oh you mean we just remove the `<Warning>` from the doc then? If so yes I'm OK with that.

brillout:
> https://github.com/vikejs/vike-photon/blob/ebabdeaad06b60f2ac1c8eb799288363cfd37de6/packages/vike-photon/src/middleware/middlewares/compress.ts#L43

Neat, I guess that's enough? AFAICT it's easy for the user to imagine the diff.

brillout:
> Oh you mean we just remove the `<Warning>` from the doc then? If so yes I'm OK with that.

Yes

magne4000:
`vike-photon@0.1.13` released that includes this change, feel free to merge this PR.

docs/pages/vike-photon/+Page.mdx
```diff
@@ -546,20 +546,45 @@ If HMR isn't what you want (for example if you modify the database connection) y
...
+    compress: false,// [!code ++]
+    // Only static files will be compressed// [!code ++]
+    // @detype-uncomment compress: 'static'// [!code ++]
+  }// [!code ++]
+} satisfies Config
 ```
 
+## Static files
+
+In production, `vike-photon` uses [sirv](https://universal-middleware.dev/middlewares/sirv) to serve static files.
```
brillout:
```suggestion
In production, `vike-photon` uses [@universal-middleware/sirv](https://universal-middleware.dev/middlewares/sirv) to serve static files.
```

PR #2805, opened 2025-11-01
fix: support other SPA frameworks within same domain (fix #2801)

### LINKED ISSUE #2801: Navigating from Another SPA Breaks Back Button
### Description

As part of [bifrost](https://github.com/Alignable/bifrost), we support non-Vike SPA frameworks on the same domain as Vike. The desired behavior is for the browser to do a full refresh when navigating between legacy pages and new Vike pages. 

This worked in prior versions (last tested in 0.4.193) of Vike by creating a dummy catch-all route and setting `clientRouting: false` on it, so that navigation away from Vike pages would trigger a full page refresh.

However, now finally upgrading to 0.4.244 and finding that while navigation works, using the browser back button to navigate from Vike to a non-Vike page fails.

Possibly introduced by #2117.

Attached minimal [reproduction as a test](https://github.com/vikejs/vike/compare/main...dajinchu:vite-plugin-ssr:dc/turbolinks-back-button?expand=1)

dajinchu:
reproduction for #2801

### DIFF
--- a/test/playground/public/e2e-test.ts
+++ b/test/playground/public/e2e-test.ts
@@ -9,6 +9,7 @@ function testOtherFrameworkNavigation() {
 
     let navPromise = await waitForNavigation()
     await page.click('a[href="/star-wars"]')
+    // Page full reload
     await navPromise()
 
     await expectUrl('/star-wars')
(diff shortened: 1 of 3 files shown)


test/playground/public/e2e-test.ts
```diff
@@ -13,6 +13,7 @@ function testOtherFrameworkNavigation() {
 
     await expectUrl('/star-wars')
     expect(await page.textContent('body')).toContain('Star Wars Movies')
+    await sleep(1000)
```
dajinchu: this reveals a critical issue: if vike has not loaded in yet, the browser will simply no-op the back button.

so it really should be up to the "other" SPA not to call pushState when leaving it's realm.

brillout:
Pre-released as `vike@0.4.245-commit-189d2b8`.

PR #2812, opened 2025-11-04
Prevent prefetch when prefetchStaticAssets is set to false

royalswe:
Add a guard to prevent creating event listeners when prefetchStaticAssets is not used


brillout:
One downside of this is that it increases the client-side KBs of bundles.

So there is a trade-off between client-side KBs and attaching useless event listeners. So far I'm inclined to go for reducing client-side KBs.

royalswe:
> One downside of this is that it increases the client-side KBs of bundles.
> 
> So there is a trade-off between client-side KBs and attaching useless event listeners. So far I'm inclined to go for reducing client-side KBs.

Understand. 
I will use window._disableAutomaticLinkInterception = false for now

PR #2820, opened 2025-11-07
Bug: throw redirect in onBeforeRenderClient breaks back button

dajinchu:
I noticed that `throw redirect()` in `onBeforeRenderClient` breaks back button behavior. It works fine in `onBeforeRender` / `data`. I believe this may be because the browser has changed URL already by the time `onBeforeRenderClient` is called.

Opened this PR with demonstrative failing spec.

Is this intentional? Could not find clear guidance in docs on where `throw redirect()` is allowed

### DIFF
--- a/test/playground/pages/redirect/+Page.tsx
+++ b/test/playground/pages/redirect/+Page.tsx
@@ -1,0 +1,3 @@
+export function Page(){
+
+}
--- a/test/playground/pages/redirect/+onBeforeRenderClient.ts
+++ b/test/playground/pages/redirect/+onBeforeRenderClient.ts
@@ -1,0 +1,8 @@
+export { onBeforeRenderClient }
+
+import type { PageContextClient } from 'vike/types'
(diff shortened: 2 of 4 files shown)


brillout:
I don't see a fundamental blocker for supporting it in `+on{Before}RenderClient`, I'll have a look at it.

brillout:
When I hit the back button in the browser, it goes back to `/`. When I execute `navigation.back()` it goes back to `/redirect` instead (which then redirects to `/about`). Can you confirm this behavior?

Not sure why `navigation.back()` results in a default behavior, maybe it's some kind of browser thing. (In principle, it shouldn't make any difference.)

As for your original request, I ain't sure what you want is a sensible default for all users. If the redirect is performed conditionally then not skipping the redirecting page is usually the desired behavior. For example, if a user goes to admin page => redirection to login page => goes back once logged in => user expects admin page.

What's your use case here?

Closing for now, but let's continue the discussion.

PR #2821, opened 2025-11-07
Bug: Effect Cannot Change Both Value and Env

dajinchu:


brillout:
Yea, effect is very much to be considered experimental. Only a handful of users use it (it's mostly used by Vike extensions).

I'll have a quick look at it.

What's your use case?

Btw. I've messaged you on Discord.

dajinchu:
It's actually from a workaround to #2820. 

**TL;DR: I believe closing #2820 would fully resolve my use case and I would not need to change both value and env in a config effect.**

Details about my use case:

1. I have a setting "proxyMode" which changes a ton of settings to enable proxying rendering to another backend.
2. I need to be able to `throw redirect()` before rendering client. `onBeforeRenderClient` does not work for that, so I moved to `onBeforeRender`. Thus, the `proxyMode` config must set the `onBeforeRender` config and ensure that it only runs client-side, as I do not want `index.pageContext.json` requests.
3. `onBeforeRender.client.js` extension does not seem to move the config over to the client.
4. So I tried setting client env https://github.com/Alignable/bifrost/blob/7177cac4b9a42231aeb8bb5c114b5a25caaa27c4/bifrost/renderer/config.ts#L38-L46

None of this worked, so now in userland, outside the extension I have to set the meta: https://github.com/Alignable/bifrost/blob/7177cac4b9a42231aeb8bb5c114b5a25caaa27c4/tests/vite/pages/proxy/wrapped/%2Bconfig.ts

PR #2827, opened 2025-11-09
chore: add `vike-photon` to CI 

brillout:


brillout:
@copilot Why is the CI failing, even though this PR shouldn't be influencing these failing tests (the tests are green before this PR)

PR #2842, opened 2025-11-17
fix: improve optimizeDeps

brillout:


brillout:
Bati CI is green: https://github.com/vikejs/bati/pull/625.

PR #2877, opened 2025-12-02
feat: `getPageContext({ asyncHook: true })`

brillout:


brillout:
@copilot Inspect errors

PR #2905, opened 2025-12-08
expose `startupLog`

brillout:


brillout:
@magne4000 Pre-released as `0.4.247-commit-87e5f2f`:

```ts
import { startupLog } from 'vike/__internal/vite'
import { type ViteDevServer } from 'vite'

let resolveUrls: ResolveUrls
type ResolveUrls = {
  local: string[]
  network: string[]
}

let viteServer: ViteDevServer

startupLog(resolveUrls, viteServer)
```

Or:

```js
viteConfig.server.startupLog(resolveUrls)
```

In production I suggest we let Photon print a simple one-liner — no need for Vike to determine that log.

PR #2942, opened 2025-12-14
fix: avoid breaking ssrFixStacktrace (vitejs/vite#21269)

brillout:


brillout:
@copilot update failing test

PR #2945, opened 2025-12-14
restore dev assert log

brillout:


brillout:
@copilot Update test (see expectLog)

PR #2950, opened 2025-12-15
fix: support list of import strings

brillout:


brillout:
@copilot fix CI

brillout:
@nitedani Pre-released as `0.4.249-commit-775a43e`.

Issue #2977, opened 2025-12-19
Feature Request: Cumulative hook wrapper for instrumentation (Sentry, OpenTelemetry, etc.)


nitedani:
## Problem

When integrating observability tools like Sentry or OpenTelemetry with Vike, we need to wrap hook executions with spans/traces. This is essential for:

- Measuring execution time of `+data`, `+guard`, `+onBeforeRender`, `+onRenderHtml`, etc.
- Providing users insight into which hooks are slow
- Proper distributed tracing

### The Challenge

Sentry (and OpenTelemetry) spans work with **async local storage** and depend on the **call stack**. This means we need to actually **wrap** the hook function execution:

```ts
Sentry.startSpan(
  {
    name: "pages/product/+data",
    op: "vike.hook",
    attributes: { hookName: "data", pageId, urlOriginal },
  },
  async () => {
    // The actual +data hook runs here
    return await dataHook(pageContext)
  },
);
```

We cannot use before/after style hooks because the span context is lost.

## Proposed Solution

Add a **cumulative global hook wrapper** to vike:

```ts
// vike-react-sentry/+config.ts
import * as Sentry from "@sentry/node"

export default {
  onWrapHook: (hookInfo, execHook) => {
    return Sentry.startSpan(
      {
        name: hookInfo.hookFilePath, // e.g. "pages/product/+data"
        op: "vike.hook",
        attributes: {
          "vike.hookName": hookInfo.hookName,
          "vike.pageId": hookInfo.pageId,
        },
      },
      () => execHook()
    )
  },
  meta: {
    onWrapHook: {
      env: { server: true, client: true },
      cumulative: true,  // Multiple wrappers compose
    }
  }
}
```

### Hook Info

The `hookInfo` parameter should include:
- `hookName`: e.g. `"data"`, `"guard"`, `"onRenderHtml"`, `"onBeforeRender"`
- `hookFilePath`: e.g. `"pages/product/+data"`, `"vike-react/__internal/integration/onRenderHtml"`
- `pageContext`: (or relevant subset like `pageId`, `urlOriginal`, `urlPathname`)

### Cumulative Behavior

Multiple wrappers should compose, allowing:
- `vike-react-sentry` to add Sentry spans
- User to add custom logging/timing
- Other extensions to add their instrumentation

```ts
// All these wrappers compose
onWrapHook: [sentryWrapper, otelWrapper, loggingWrapper]
```

### Implementation Location

[...]

PR #2978, opened 2025-12-20
feat: [+onHookCall](http://vike.dev/onHookCall)

nitedani:
Adds a new cumulative global config `onHookCall` that wraps all hook calls, enabling instrumentation for monitoring tools like Sentry or OpenTelemetry.

### Usage

```ts
// +onHookCall.ts
import type { OnHookCall } from 'vike/types'

export const onHookCall: OnHookCall = async (hook, context) => {
  console.log('before', hook.name, hook.filePath)
  const result = await hook.call()
  console.log('after', hook.name)
  return result
}
```

### Details

- **Hook object** contains: `name`, `filePath`, `call()`
- **Context** is `pageContext` for page hooks, `globalContext` for global hooks
- **Wrappers applied at call time** via `execHookWithOnHookCall()` in `execHook.ts` — not at hook creation time
- **HMR-safe**: Wrappers are read fresh from `pageConfigGlobal` on each hook call, so changes take effect immediately
- **Prevents infinite recursion**: `onHookCall` hooks are not wrapped by themselves
- **Cumulative**: Multiple `onHookCall` wrappers can be defined and will be applied in order

### Implementation

- Added `OnHookCall` and `OnHookCallHook` types to `Config.ts`
- Added `execHookWithOnHookCall()` function in `execHook.ts` that applies wrappers at call time
- Updated all hook execution sites to use `execHookWithOnHookCall()`
- Extended `GlobalContextPrepareMinimum` to include `_pageConfigGlobal` for accessing wrappers
- Added documentation page at `/onHookCall` with use cases for Sentry, OpenTelemetry, and logging
- Added backlinks in `/hooks`, `/onError`, and `/error-tracking` pages

### Type Signature

```ts
type OnHookCallHook = {
  name: string
  filePath: string
  call: () => unknown | Promise<unknown>
}

type OnHookCall = (hook: OnHookCallHook, context: unknown) => unknown | Promise<unknown>
```

https://github.com/vikejs/vike/issues/2977

### DIFF
--- a/docs/pages/onHookCall/+Page.mdx
+++ b/docs/pages/onHookCall/+Page.mdx
@@ -122,7 +122,7 @@ import type { Config } from 'vike/types'
 
 export const onHookCall: Config['onHookCall'] = async (hook, pageContext) => {
   const start = performance.now()
-  console.log(`[+${hook.name}] start`, hook.filePath)
+  console.log(`[+${hook.name}] start`)
   try {
     await hook.call()
     console.log(`[+${hook.name}] done in ${(performance.now() - start).toFixed(2)}ms`)


packages/vike/src/shared-server-client/hooks/getHook.ts
```diff
@@ -219,3 +229,31 @@ function getHookTimeoutDefault(hookName: HookNameOld): HookTimeout {
...
+      const prev = fn
+      fn = () => wrapper({ ...hook, call: prev }, context)
+    }
+    return fn()
+  }
+}
+
+function getOnHookCallWrappers(pageConfigGlobal: PageConfigGlobalRuntime): OnHookCall[] {
+  const configValue = pageConfigGlobal.configValues['onHookCall']
+  if (!configValue?.value || !isArray(configValue.value)) return []
```
brillout: Let's avoid silently failing. Let's skip validation to save client-side KBs.

packages/vike/src/shared-server-client/hooks/getHook.ts
```diff
@@ -219,3 +229,31 @@ function getHookTimeoutDefault(hookName: HookNameOld): HookTimeout {
...
+  hookFilePath: string,
+  pageConfigGlobal: PageConfigGlobalRuntime,
+): HookFn<HookArg> {
+  return function (this: unknown, ...args: unknown[]) {
+    const wrappers = getOnHookCallWrappers(pageConfigGlobal)
+    const callFn = () => (hookFn as Function).apply(this, args)
+    if (wrappers.length === 0) return callFn()
+    // First argument is pageContext for page hooks, globalContext for global hooks
+    const context = args[0]
+    const hook = { name: hookName, filePath: hookFilePath, call: callFn }
```
brillout:
```suggestion
    const hookPublic = { name: hookName, filePath: hookFilePath, call: callFn }
```

packages/vike/src/shared-server-client/hooks/getHook.ts
```diff
@@ -219,3 +229,31 @@ function getHookTimeoutDefault(hookName: HookNameOld): HookTimeout {
...
   globalObject.isPrerendering = true
 }
+
+function wrapHookFn<HookArg>(
+  hookFn: HookFn<HookArg>,
+  hookName: HookNameOld,
+  hookFilePath: string,
+  pageConfigGlobal: PageConfigGlobalRuntime,
+): HookFn<HookArg> {
+  return function (this: unknown, ...args: unknown[]) {
```
brillout: We don't need `this`, let's remove it

packages/vike/src/shared-server-client/hooks/getHook.ts
```diff
@@ -219,3 +229,31 @@ function getHookTimeoutDefault(hookName: HookNameOld): HookTimeout {
...
+    const wrappers = getOnHookCallWrappers(pageConfigGlobal)
+    const callFn = () => (hookFn as Function).apply(this, args)
+    if (wrappers.length === 0) return callFn()
```
brillout: I think we can remove this line, let's do it if we can to save client-side KBs

packages/vike/src/shared-server-client/hooks/getHook.ts
```diff
@@ -125,10 +133,12 @@ function getHook<HookArg = HookArgDefault>(
   hookName: HookNameOld,
   hookFilePath: string,
   hookTimeout: HookTimeout,
+  pageConfigGlobal: PageConfigGlobalRuntime,
 ): Hook<HookArg> {
   assert(hookFilePath)
   assertHookFn<HookArg>(hookFn, { hookName, hookFilePath })
-  const hook = { hookFn, hookName, hookFilePath, hookTimeout }
+  const wrappedHookFn = wrapHookFn(hookFn, hookName, hookFilePath, pageConfigGlobal)
```
brillout:
```suggestion
  hookFn = wrapHookFn(hookFn, hookName, hookFilePath, pageConfigGlobal)
```

packages/vike/src/shared-server-client/hooks/getHook.ts
```diff
@@ -80,24 +84,28 @@ function getHookFromPageContextNew(hookName: HookName, pageContext: PageContextC
...
-    hooks.push(getHook(hookFn, hookName, hookFilePath, hookTimeout))
+    hooks.push(getHook(hookFn, hookName, hookFilePath, hookTimeout, pageConfigGlobal))
   })
   return hooks
 }
-function getHookFromPageConfig(pageConfig: PageConfigRuntime, hookName: HookNamePage): null | Hook {
+function getHookFromPageConfig(
+  pageConfig: PageConfigRuntime,
+  hookName: HookNamePage,
+  pageConfigGlobal: PageConfigGlobalRuntime,
```
brillout: I think we can pass `globalContext` instead (and then use `globalContext._pageConfigGlobal`, let's do this if possible (I think so) — it's more future proof

test/vike-react/pages/+onHookCall.ts
```diff
@@ -0,0 +1,8 @@
+import { OnHookCall } from 'vike/types'
```
brillout: Let's move it to `test/playground/` — that's the default place to try things out. Bonus: it defines a *lot* of hooks :)

packages/vike/src/shared-server-client/hooks/getHook.ts
```diff
@@ -219,3 +229,31 @@ function getHookTimeoutDefault(hookName: HookNameOld): HookTimeout {
...
+    let fn = callFn
+    for (const wrapper of wrappers) {
+      const prev = fn
+      fn = () => wrapper({ ...hook, call: prev }, context)
+    }
+    return fn()
+  }
+}
+
+function getOnHookCallWrappers(pageConfigGlobal: PageConfigGlobalRuntime): OnHookCall[] {
```
brillout: Either we inline it or we follow a convention and name it `getHooksOnHookCall` (not sure if there is already such convention for `getHook*` and wehther it's widely used)

brillout:
Looks great. I've nit-picked a bit because in my experience I always regret it when I don't highly polish Vike core code.

brillout:
Does it work for `vike-react-sentry`? :eyes: I can add you to https://www.npmjs.com/package/vike so you can pre-release it to easily try it out (cross-monorepo linking is often a pain).

Looking forward to merge this one :rocket:

brillout:
The failing CI seems interesting. I did't dig into it —so I ain't sure— but it does seem like a legit regression.

packages/vike/src/types/Config.ts
```diff
@@ -712,4 +726,19 @@ type ConfigBuiltInResolved = {
...
+  /** File path where the hook is defined */
+  filePath: string
+  /** Call the hook */
+  call: () => unknown
+}
+/** Wrapper for hook calls. Used for instrumentation (e.g. Sentry, OpenTelemetry).
+ *
+ * The `context` argument is `pageContext` for page hooks and `globalContext` for global hooks.
+ */
+type OnHookCall = (hook: OnHookCallHook, context: unknown) => unknown
```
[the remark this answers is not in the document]
nitedani: Using generics here wouldn't provide real benefit since the wrappers are resolved dynamically at runtime and all hooks have different return types.

nitedani:
@brillout Please check now

packages/vike/src/shared-server-client/hooks/execHook.ts
```diff
@@ -192,10 +197,33 @@ function execHookDirectSync<PageContext extends PageContextPrepareMinimum>(
...
+  globalContext: GlobalContextPrepareMinimum,
+  ...args: unknown[]
+): HookReturn {
+  const { hookFn, hookName, hookFilePath } = hook
+  // Don't wrap onHookCall itself to prevent infinite recursion
+  if (hookName === 'onHookCall') return hookFn(...args)
+  const configValue = globalContext._pageConfigGlobal.configValues['onHookCall']
+  if (!configValue?.value) return hookFn(...args)
+  const wrappers = configValue.value
+  if (!isArray(wrappers)) return hookFn(...args)
```
brillout: It's always `undefined | unknown[]` because it's cumulative — let's type cast it to that and remove the validation. (Please no unsupervised AI in case you do it unsupervised.)

packages/vike/src/types/index.ts
```diff
@@ -24,6 +24,8 @@ export type {
   ConfigMeta as Meta,
   ImportString,
   KeepScrollPosition,
+  OnHookCall,
```
brillout: Let's remove it — FYI we deprecated all hook types.

packages/vike/src/node/prerender/runPrerender.ts
```diff
@@ -413,7 +414,10 @@ async function callOnBeforePrerenderStartHooks(
         if (doNotPrerenderList.find((p) => p.pageId === pageId)) return
         const { hookName, hookFilePath } = hook
 
-        const prerenderResult = await execHookDirectWithoutPageContext(() => hook.hookFn(), hook)
+        const prerenderResult = await execHookDirectWithoutPageContext(
```
brillout: I think it's cleaner if we modify execHookDirectWithoutPageContext instead of using execHookWithOnHookCall here? It seems brittle.

packages/vike/src/server/runtime/renderPageServer/execHookOnError.ts
```diff
@@ -21,7 +22,7 @@ function execHookOnError(err: unknown, pageContext: PageContext_logRuntime) {
   const hooks = getHookFromPageConfigGlobalCumulative<unknown>(globalContext._pageConfigGlobal, 'onError')
   for (const hook of hooks) {
     try {
-      hook.hookFn(err, pageContext)
+      execHookWithOnHookCall(hook, globalContext, err, pageContext)
```
brillout: Let's define some kind of new execHook function (e.g. execHookVanilla) — someone *will* forget to call execHookWithOnHookCall. Ideally execHookWithOnHookCall should be internal to execHook.ts

brillout: Ideally, *all* hook calls should go through `execHook.ts`

brillout: (We could eventually export something for Vike extensions e.g. when they call `+onAfterRenderHtml`. But let's skip this for now.)

packages/vike/src/shared-server-client/hooks/execHook.ts
```diff
@@ -50,16 +51,20 @@ async function execHook<PageContext extends PageContextExecHook>(
...
   prepareForPublicUsage: (hookArg: HookArg) => HookArg,
 ) {
-  const hooks = getHookFromPageConfigGlobalCumulative(pageConfigGlobal, hookName)
+  const hooks = getHookFromPageConfigGlobalCumulative(globalContext._pageConfigGlobal, hookName)
   const hookArgForPublicUsage = prepareForPublicUsage(hookArg)
   await Promise.all(
     hooks.map(async (hook) => {
-      await execHookDirectAsync(() => hook.hookFn(hookArgForPublicUsage), hook, pageContext)
+      await execHookDirectAsync(
+        () => execHookWithOnHookCall(hook, globalContext, hookArgForPublicUsage),
```
brillout: If we can use execHookWithOnHookCall at only exactly one place, that'd be great — i.e. some kind new core execHook function that *every* hook call goes through. (We don't need to actually make sure that every hook call go through that core func —if we forget a couple of minor hooks that's okay— but let's evolve in that direction.)

packages/vike/src/shared-server-client/hooks/execHook.ts
```diff
@@ -74,7 +79,7 @@ async function execHookDirect<PageContext extends PageContextPrepareMinimum>(
   const hooksWithResult = await Promise.all(
     hooks.map(async (hook) => {
       const hookReturn = await execHookDirectAsync(
-        () => hook.hookFn(pageContextForPublicUsage),
+        () => execHookWithOnHookCall(hook, pageContext._globalContext, pageContextForPublicUsage),
```
brillout: Maybe just a execHookBase that basically just calls execHookWithOnHookCall. I don't know, let's double chekc that and spend a bit of time on this — very well worth it.

packages/vike/src/shared-server-client/hooks/execHook.ts
```diff
@@ -192,10 +197,33 @@ function execHookDirectSync<PageContext extends PageContextPrepareMinimum>(
...
 }
 
+function execHookWithOnHookCall<HookReturn>(
+  hook: HookLoc & { hookFn: Function },
+  globalContext: GlobalContextPrepareMinimum,
+  ...args: unknown[]
+): HookReturn {
+  const { hookFn, hookName, hookFilePath } = hook
+  // Don't wrap onHookCall itself to prevent infinite recursion
+  if (hookName === 'onHookCall') return hookFn(...args)
```
brillout: I think this can be an `assert()` (I don't see the recursion)

brillout: `assert(hookName!=='onHookCall')`

packages/vike/src/types/Config.ts
```diff
@@ -431,6 +439,12 @@ type ConfigBuiltIn = {
    */
   onCreateGlobalContext?: ((globalContext: GlobalContext) => void) | ImportString | null
 
+  /** Hook called when any config function is invoked. Useful for instrumentation (Sentry, OpenTelemetry, etc.).
```
brillout:
```suggestion
  /** Called when any hook is called. Useful for instrumentation (Sentry, OpenTelemetry, etc.).
```

packages/vike/src/types/Config.ts
```diff
@@ -712,4 +726,31 @@ type ConfigBuiltInResolved = {
 
 type ConfigMeta = Record<string, ConfigDefinition>
 type ImportString = ImportStringVal | ImportStringVal[]
+
+/** Info about the hook being called */
```
brillout:
```suggestion
/** Information about the hook being called */
```

packages/vike/src/types/Config.ts
```diff
@@ -712,4 +726,31 @@ type ConfigBuiltInResolved = {
...
+type OnHookCallHook = {
+  /** Name of the hook, e.g. 'onRenderHtml', 'data', 'guard' */
+  name: string
+  /** File path where the hook is defined */
+  filePath: string
+  /** Call the hook and return its result */
```
brillout:
```suggestion
  /** Trigger the hook call */
```

packages/vike/src/types/index.ts
```diff
@@ -24,6 +24,8 @@ export type {
...
+  OnHookCallHook,
```
brillout: We can also remove this and document `Config['onHookCall']` instead like the other hooks

docs/pages/onHookCall/+Page.mdx
```diff
@@ -0,0 +1,123 @@
...
+<ConfigSpec
+  env="server, client"
+  cumulative
+  global
+  isTypeOneLiner
+>
+`(hook: OnHookCallHook, context: unknown) => unknown | Promise<unknown>{:ts}`
+</ConfigSpec>
+
+Hook called whenever a Vike hook is invoked. Useful for instrumentation (e.g. Sentry, OpenTelemetry).
```
brillout:
```suggestion
Hook called whenever a Vike hook is called. Useful for instrumentation (e.g. Sentry, OpenTelemetry).
```

docs/pages/onHookCall/+Page.mdx
```diff
@@ -0,0 +1,123 @@
+import { Link } from '@brillout/docpress'
+import { ConfigSpec } from '../../components'
+
+<ConfigSpec
+  env="server, client"
+  cumulative
+  global
+  isTypeOneLiner
+>
+`(hook: OnHookCallHook, context: unknown) => unknown | Promise<unknown>{:ts}`
```
brillout: That isn't useful

brillout: The user doesn't know what OnHookCallHook when reading that page

docs/pages/onHookCall/+Page.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+export const onHookCall: OnHookCall = async (hook, context) => {
+  console.log('before', hook.name, hook.filePath)
+  const result = await hook.call()
+  console.log('after', hook.name)
+  return result
+}
+```
+
+> **Important:** The wrapper must return the result of `hook.call()` to preserve the hook's return value.
```
brillout: Let's avoid the user being able to modify the return value. (Maybe it's tricky to achieve given the current implementation because of the promise chaining, let's discuss this if it indeed is tricky).

Note that we'll soon deprecate returning values from all hooks.

docs/pages/onHookCall/+Page.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+The `hook` object contains:
+- `hook.name` — Name of the hook being called (e.g. `'onRenderHtml'`, `'data'`, `'guard'`)
+- `hook.filePath` — File path where the hook is defined
+- `hook.call()` — Function to call the actual hook
+
+The `context` argument is <Link href="/pageContext">`pageContext`</Link> for page hooks and <Link href="/globalContext">`globalContext`</Link> for global hooks.
```
brillout: Let's add a link to https://vike.dev/config#global

docs/pages/onHookCall/+Page.mdx
brillout: > (Maybe it's tricky to achieve given the current implementation because of the promise chaining, let's discuss this if it indeed is tricky).

While we're at it: I think it's more idiomatic if all onHookCall are exec'd in parallel instead of sequentially like it currently is. But I won't nit-pick on that, I can't think of why it would make a practical different either way. Whatever is easier to implement (with a slight preference for parallel).

docs/pages/onHookCall/+Page.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+
+## Use case: error tracking
```
brillout: `Sentry.startSpan` seems to be about Instrumentation: https://docs.sentry.io/platforms/javascript/tracing/instrumentation/

Let's call this section accoridngly: `## Use case: Instrumentation with Sentry`

brillout: Or better: `## Use case: Instrumentation` (fits better the sectoin below).

brillout:
Nice nice, getting closer to mergeability

docs/pages/onHookCall/+Page.mdx
nitedani: It's tricky but doable, the sync and async wrapping need separate paths

nitedani:
I separated sync and async paths of wrapping, the new types:
```ts
type OnHookCallContext = {
  pageContext?: PageContext
  globalContext?: GlobalContext
}

type OnHookCall = {
  sync?: (hook: { name: string; filePath: string; call: () => unknown }, context: OnHookCallContext) => void
  async?: (
    hook: { name: string; filePath: string; call: () => Promise<unknown> },
    context: OnHookCallContext,
  ) => Promise<void>
}
```

The expected return type is void, the original hook's return value is captured and returned automatically.
The separation of sync/async is needed because we can't wrap sync hooks with an async onHookCall.

brillout:
I see, make sense. Hm, I wonder what direction we should take. Let me think it through.

brillout:
I feel like `sync` doesn't add much(/any?) value.

The best would be this:

```js
export const onHookCall = async (hook, context) => {
  console.log('before', hook.name, hook.filePath)
  await hook.call()
  console.log('after', hook.name)
}
```

So wouldn't simply removing the sync version do the trick here? AFAICT the sync variant is useless.

Regarding the implementation, maybe the trick is to capture the return value of the original function while discarding & ignoring the return value of `+onHookCall` — so no chaining at all.

Couple of notes:
 - IIRC only `+onBeforeRoute` is sync, all other ones are async (or can be made async).
 - We can revert to the yesterday version, label +onHookCall as experimental, and then remove the `return` warning once we always forbid returning values from hooks in the next major version. But let's do this only as last resort.

nitedani:
> Regarding the implementation, maybe the trick is to capture the return value of the original function while discarding & ignoring the return value of `+onHookCall`

The code is doing this already.


> So wouldn't simply removing the sync version do the trick here? AFAICT the sync variant is useless.

How would you wrap a sync function with an async function, while keeping the execution sync? It's not possible. 
It *can* work in an edge case, only if the first await inside the async wrapper executes the original sync function itself. 

For example, this would break the sync hooks:
```ts
export const onHookCall = async (hook, context) => {
  await 1;
  await hook.call()
}
```

Explanation:
```ts
let v = 1

function increaseV() {
  v = 2
}

async function f() {
  await increaseV()
}

console.log(v) // prints 1
f()
console.log(v) // prints 2
```

```ts
let v = 1

function increaseV() {
  v = 2
}

async function f() {
  await 1;
  await increaseV()
}

console.log(v) // prints 1
f()
console.log(v) // prints 1
```

brillout:
By ignoring the promise returned by +onHookCall and skipping the whole promise chaining. Let me re-read your code.

brillout:
Ah, wait, I see you're right. Hm.

brillout:
Let me think

packages/vike/src/shared-server-client/hooks/execHook.ts
```diff
@@ -192,10 +202,76 @@ function execHookDirectSync<PageContext extends PageContextPrepareMinimum>(
...
 }
 
+function execHookVanillaSync<HookReturn>(
+  hookFnCaller: () => HookReturn,
+  hook: Omit<Hook, 'hookTimeout' | 'hookFn'>,
+  globalContext: GlobalContextPrepareMinimum,
+  pageContext: PageContextPrepareMinimum | null = null,
+): HookReturn {
+  const ret = execHookWithOnHookCall(hookFnCaller, hook, globalContext, pageContext, true)
+  assert(!isPromise(ret))
```
brillout:
```suggestion
```

I think this assert fails if the user defines both +onBeforeRoute and +onHookStart — I think we can & should remove the assertion anyways as it reduces client-side KBs.

brillout:
The CI fails because of the following error, see https://github.com/brillout/docpress?tab=readme-ov-file#detype.

Easy fix: two code blocks instead of one.

```
[10:22:57.435][/docs/.test-preview.test.ts][pnpm run preview][stderr] /home/runner/work/vike/vike/docs/some-dummy-filename.ts: Identifier 'onHookCall' has already been declared. (8:13)

   6 | /* @detype: empty-line=2 */
   7 | // ✅ This works for both sync and async hooks
>  8 | export const onHookCall: Config['onHookCall'] = async (hook, context) => {
     |              ^
   9 |   await hook.call() // Called before any await
  10 |   await someAsyncOperation()
  11 | }
[10:22:57.436][/docs/.test-preview.test.ts][pnpm run preview][stderr] Failed to transform the code block in: /home/runner/work/vike/vike/docs/pages/onHookCall/+Page.mdx.
```

brillout:
The sync check and promise chaining had issues. I went ahead refactored it. For some of the rational, see the removed commnets at e92ab635e963750122d9e50453cf990fcbe8b804.

packages/vike/src/shared-server-client/hooks/execHook.ts
```diff
@@ -192,10 +204,69 @@ function execHookDirectSync<PageContext extends PageContextPrepareMinimum>(
...
+    originalCalled = true
+    originalReturn = hookFnCaller()
+    return originalReturn
+  }
+  for (const onHookCall of configValue.value as Function[]) {
+    const hookPublic = { name: hookName, filePath: hookFilePath, sync, call }
+    // (It would be simpler to define a single hookFnCaller() wrapper instead of chaining call() functions, but it would break the async hooks of the vike-react-sentry extension.)
+    call = () => {
+      // +onHookCall should call hookPublic.call() (the previous call() function) => chaining
+      onHookCall(hookPublic, context)
```
nitedani: If we don't await this, then when an async hook is wrapped like this:

```js
async onHookCall(hook){
 await 1;
 await hook.call()
}
```

Then `originalReturn` will be undefined

nitedani: What do you mean "promise chaining had issues"? Please give an example, I'd like to understand the issue also.

brillout: > Then `originalReturn` will be undefined

Hm, good catch.

> What do you mean "promise chaining had issues"? Please give an example, I'd like to understand the issue also.

Problems with the version down below:

1. assertUsage() should be called synchronously (but it's called after awaiting the promise instead) — we want to ensure `hook.call()` was called before any `await`.
2. ~~While the `+onHookCall` hooks are *called* sequentially (i.e. they're chained), the promises aren't *awaited* sequentially: Vike (`await execHookVanilla()`) only awaits the promise of the last `+onHookCall` hook.~~ Ah, actually, I'm wrong: the promise of each +onHookCall *are* awaited sequentially. All good here ✅
3.  By design, `+onHookCall` cannot change the return value of a hook. In the same spirit, I'd argue that Vike shouldn't await the promise of `+onHookCall`.

https://github.com/vikejs/vike/blob/a6b912aa1c19b004f34571029fe52b8af54bbd50/packages/vike/src/shared-server-client/hooks/execHook.ts#L261-L268

brillout: I'm flexible about `3.`, the only problem is `1.`.

So, I think the following version would work? (You can ignore this comment: `ACTUALLY: I'm wrong, the promises aren't awaited sequentially`.)

https://github.com/vikejs/vike/blob/4f9967329b94346e19c9ab0b13800a7b4edf3a55/packages/vike/src/shared-server-client/hooks/execHook.ts#L258-L269

brillout: If we want `3.` then I guess we can go for this: 9e5f31273ddcff13bda776c9185aea55d7e213ec

nitedani: Okay. I agree, and at first glance, I think that will work.

nitedani: I think actually awaiting the wrapper is more predictable for the user: 

If you only await the original hook’s promise, then after that promise is resolved, modifications to the context object in the wrapper can cause  unexpected behavior.

Because after the hook.call() line, *async* modifications to the context object would be lost/cause context state fragmentation.

brillout: > more predictable

Good point. I also like that it's more succinct. Let me revert to that version and fix the CI.

brillout: How about this: https://github.com/vikejs/vike/pull/2978/commits/5cf58d094650f2c01eab32b332161e8abe71f97e ?

But I think we still have an issue with `sync` because it still resolves a promise (because the wrapped `call()` are async). Ideas?

brillout: As for the CI: 0c3831e98ed35e92232cf8e9082a5d798ad4edc8. WDYT?

nitedani: This is expected for me, swallowing the original hook’s redirect, etc.. errors break things.

The question is whether we should allow them to be swallowed.

There are ways to code the wrapping so it can’t swallow the original hook’s errors specifically.

nitedani: > How about this: [5cf58d0](https://github.com/vikejs/vike/commit/5cf58d094650f2c01eab32b332161e8abe71f97e) ?
> 
> But I think we still have an issue with `sync` because it still resolves a promise (because the wrapped `call()` are async). Ideas?

I think that’s fine and it would behave as intended on all runtimes.


1. Promise Executor Runs Synchronously

From the ECMAScript specification (Section 27.2.3.1):
“The executor function is called synchronously when the Promise constructor is called, before the Promise constructor returns the created object.”

2. Async Functions Execute Immediately (Until First Await)

From the ECMAScript specification (Section 27.7.5):
“When an async function is called, a new Promise is returned. The function body begins execution synchronously, executing until the first await expression is encountered.”

brillout: The issue is that it returns `Promise<typeof originalReturn>` — Vike internally consumes the wrapped async call, not the original call.

brillout: I reproduced the issue with this: https://github.com/vikejs/vike/pull/2978/commits/43ae4f467adad549e0309c89748bfc83971b3584

You can see the CI becomes red because of this warning:

```
[08:23:45.195][/test/playground/test-dev.test.ts][npm run dev][stderr] 8:23:45 AM [vike][request-1][Warning] The onBeforeRoute() hook /+onBeforeRoute.ts returned a promise, but asynchronous routing is deprecated and will be removed in the next major release, see https://vike.dev/route-function#async
```

I tried to fix it by completely removing the async awaiting (I'm aware of the issue your raised yesterday) with this implementation https://github.com/vikejs/vike/pull/2978/commits/36fe0840e44d0908b4f40e25c2362e5cf72c483f but this swallows the `throw redirect()` — I think because `await hook.call()` turns `throw redirect()` into a rejected promise (resulting in the `throw redirect()` to be swallowed from Vike's perspective).

Let me try something else. (Feel free to try as well if you want.)

brillout: How about this: a821d8cb6f8228e55045a8df45e44a011538d695 ?

But note that:
1. It doesn't await the `+onHookCall` promise (unpredictable behavior as you pointed out yesterday)
2. It forces `+onHookCall` to always run `hook.call()` before any `await`

I agree that 1. is a DX regression.

About `2.`, after thinking more about it, I actually now believe it's too brittle to allow `await` before `hook.call()`. Because extensions won't test their extension against `+onBeforeRoute` => when an extension user adds `+onBeforeRoute` to his app => boom :boom:

I'd suggest we address the limitations of `1.` and `2.` only after we get a complaint from a user. If we find a working solution, I'd suggest we just go with it (we can also ship fast and label `+onHookCall` as experimental). But I'm happy to discuss further and feel free to experiment on your own. No rush on my side (although I can't wait to add `+onHookData` to our nice suite of hooks).

nitedani: >[08:23:45.195][/test/playground/test-dev.test.ts][npm run dev][stderr] 8:23:45 AM [vike][request-1][Warning] The onBeforeRoute() hook /+onBeforeRoute.ts returned a promise, but asynchronous routing is deprecated and will be removed in the next major release, see https://vike.dev/route-function#async


I reproduced this warning. 
But the assertion is technically incorrect, because Promise and async function bodies are also executed synchronously.

Explanation:
```ts
const util = require('util')

const p = new Promise((resolve) => resolve(42))
console.log(isFulfilled(p)) // logs true

function isFulfilled(promise) {
  return util.inspect(promise).startsWith('Promise {') && !util.inspect(promise).includes('<pending>')
}
```

```ts
const util = require('util')

const a = async () => {
    return 42
}
console.log(isFulfilled(a())) // logs true

function isFulfilled(promise) {
  return util.inspect(promise).startsWith('Promise {') && !util.inspect(promise).includes('<pending>')
}

```

nitedani: >I'd suggest we address the limitations of 1. and 2. only after we get a complaint from a user.

Okay. Let's go with this.

nitedani:
Let's go with your version.

packages/vike/src/shared-server-client/hooks/execHook.ts
brillout: > But the assertion is technically incorrect

I see, makes sense. Unfortunately there doesn't seem to be an (easy) way to synchronously get value of a fulfilled promise, otherwise it would work indeed.

brillout:
There were some issues with error handling — it's now fixed: https://github.com/vikejs/vike/pull/2978/changes/ab958dae1536c59ebdcde7ddbcc3275955e923c2.

@nitedani I'll do one last round of polishing (the only noticeable change is that I'll remove `gloablContext` from the second parameter because the user can access it via [`getGlobalContext()`](https://vike.dev/getGlobalContext)) and then we can merge?

brillout:
@nitedani Done. LGTM. Let me know if it's LGTM for you as well and I'll merge.

nitedani:
Okay. Let's do it

brillout:
Pre-released as `0.4.250-commit-6161324` — let me know if you want me to release `0.4.251`.

PR #2980, opened 2025-12-21
enforce explicit import strings

brillout:


brillout:
@copilot fix it (see CI error) — i.e. make sure import strings are explicit

PR #3040, opened 2026-01-17
chore: only run discord webhook step if webhook url defined

AgentEnder:
Disables the step to run the discord curl call if the webhook URL is undefined

### DIFF
--- a/.github/workflows/discord.yml
+++ b/.github/workflows/discord.yml
@@ -14,7 +14,7 @@ jobs:
       DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
     steps:
       - name: Push the notification using Discord webhook
-        # Ideally this would be combined with the prior if, but the env isnt accessible at that level
+        # Skip pull requests. (This condition cannot be moved to the `if:` above because the secret isn't available there.)
         if: env.DISCORD_WEBHOOK != ''
         env:
           # 🟢 has poor compatibility: https://www.unicompat.com/1F7E2


brillout:
Curious: what's the motivation here?

AgentEnder:
> Curious: what's the motivation here?

External contributors get a github notification from this job running on their fork without a url

brillout:
:+1: Makes sense. Can it be hoisted to the `if:` above? I guess the env is set before the `if:`?

AgentEnder:
I dont think so? I know you can't directly access the secret in the if block, but I think env isnt set there as well

.github/workflows/discord.yml
```diff
@@ -8,18 +8,20 @@ on:
 jobs:
   send-discord-message:
     # 'skipped' means the `if: ` condition wasn't fulfilled
-    if: ${{ github.event.workflow_run.conclusion != 'skipped' }}
+    if: ${{ github.event.workflow_run.conclusion != 'skipped'  }}
```
brillout:
```suggestion
    if: ${{ github.event.workflow_run.conclusion != 'skipped' }}
```

brillout:
> I know you can't directly access the secret in the if block

Hm, I wonder why that is.

Let's add a comment explaing the second `if:` then we can merge.

.github/workflows/discord.yml
```diff
@@ -10,16 +10,18 @@ jobs:
     # 'skipped' means the `if: ` condition wasn't fulfilled
     if: ${{ github.event.workflow_run.conclusion != 'skipped' }}
     runs-on: ubuntu-latest
+    env:
+      DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
     steps:
       - name: Push the notification using Discord webhook
+        if: env.DISCORD_WEBHOOK != ''
```
AgentEnder, line 18:
```suggestion
        # Ideally this would be combined with the prior if, but the env isnt accessible at that level
        if: env.DISCORD_WEBHOOK != ''
```

.github/workflows/discord.yml
```diff
@@ -10,16 +10,19 @@ jobs:
     # 'skipped' means the `if: ` condition wasn't fulfilled
     if: ${{ github.event.workflow_run.conclusion != 'skipped' }}
     runs-on: ubuntu-latest
+    env:
+      DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK }}
     steps:
       - name: Push the notification using Discord webhook
+        # Ideally this would be combined with the prior if, but the env isnt accessible at that level
```
brillout:
```suggestion
        # Skip pull requests. (This condition cannot be moved to the `if:` above because the secret isn't available there.)
```

brillout:
Thank you :green_heart:

PR #3096, opened 2026-02-11
Add vavite example

dajinchu:


brillout:
@dajinchu I can't make vavite work in prod/preview mode. Any ideas why it doesn't work? Does it work on your side if you use `vike@0.4.253-commit-6cfeda2` (includes both fix I just wrote to fix the vavite regression)?

brillout:
Nervermind, it's fixed.

brillout:
By the way, we're collecting testimonials — contribution welcome to share 1-2 sentences about why you chose Vike.

Issue #3107, opened 2026-02-18
Improve warning Omitting Vike's Vite plugin (inside your vite.config.js) is deprecated


robin-ln:
### Description

i add the plugin vike but i get a warning : 

```
17:35:06 [vike][Warning] Omitting Vike's Vite plugin (inside your vite.config.js) is deprecated — make sure to always add Vike's Vite plugin https://vike.dev/vite-plugin
➜ Listening on: http://127.0.0.1:8080/
Server running at http://127.0.0.1:8080/
```

Reproduction warning : 
```
// vite.config.ts
import type { PluginOption } from 'vite'
import { defineConfig } from 'vite'

const plugins = async (): Promise<PluginOption[]> => {
  const vue = (await import('@vitejs/plugin-vue')).default
  const vike = (await import('vike/plugin')).default
  return [vue(), vike()]
}

// https://vitejs.dev/config/
export default defineConfig({
  plugins: [plugins()],
})
```

can you improve this function [findVikeVitePlugin](https://github.com/vikejs/vike/blob/79aa32ddb43731043a345d6c356e4fea2b627cb5/packages/vike/src/node/api/resolveViteConfigFromUser.ts#L150-L163) ?

brillout:
Why do you need this?

brillout:
I'm inclined to think that Vite should be awaiting your config.

But a workaround might be possible on Vike's side.

brillout:
> But a workaround might be possible on Vike's side.

I tried, but couldn't find a workaround.

Closing as per https://github.com/vikejs/vike/pull/3108#issuecomment-3923625755.

@robin-ln I'd recommend to create an issue over at Vite's repository. You can try to define vite.config.js as an async module, but I don't know if Vite supports that.

PR #3108, opened 2026-02-18
[failed-attempt] fix: support async vite.config.js (#3107)

brillout:
https://github.com/vikejs/vike/issues/3107


brillout:
Closing as the root cause seems to be on Vite's side: Vite's `loadConfigFromFile()` seems to swallow the user defined plugin promise. I don't see how Vike could work around this.

PR #3156, opened 2026-03-18
fix: duplicate URL verification on pre-rendering

tperale:
When a single 'onBeforePrerenderStart' hook pass duplicated URL entries the part of the 'callOnBeforePrerenderStartHooks' that verify duplicated URL by iterating through the current 'pageContexts' array don't complain with an error.

After inspecting the content of the 'pageContexts', I noticed the array would still be empty when checking for existing URLs. The concurrent access to the 'pageContexts' and the context switching happening before this array is populated made this array appear empty for a number of page entries that tries to check its content.

Before this commit the URL generation, URL duplication check and pageContext generation was handled in a single Promise. With this commit the URL generation is handled in a separate set of Promises. Based on the result of the URL generation the duplicate URLs are checked synchronously. The 'pageContext' are then generated on another separate Promise which populates the 'pageContexts' array.

This allows to not check the content of the 'pageContexts' array while populating it.

I didn't find any place where I can add tests for the duplicate URL. Please let me know if I can add one somewhere.

### DIFF
--- a/packages/vike/src/node/prerender/runPrerender.ts
+++ b/packages/vike/src/node/prerender/runPrerender.ts
@@ -448,11 +448,11 @@ async function callOnBeforePrerenderStartHooks(
   // If duplicate URL is found an error will be issued
   const pageContextsByUrl: Record<string, PageContextPrerendered> = {}
   for (const pageContext of prerenderContext.pageContexts) {
+    assert(pageContext._providedByHook)
     const urlNormalized = normalizeUrl(pageContext.urlOriginal)
     const pageContextSameUrl = pageContextsByUrl[urlNormalized]
     if (pageContextSameUrl) {
       assert(pageContextSameUrl._providedByHook)
-      assert(pageContext._providedByHook)
       const { hookName, hookFilePath } = pageContext._providedByHook
       const providedTwice =
         hookFilePath === pageContextSameUrl._providedByHook.hookFilePath


brillout:
It'd be nice to fix it indeed. How about we make this PR more minimal? Would it be sufficient to read `prerenderContext.pageContexts` after they're all collected? Not need to refactor that much?

Currently, it's a lot of changes for merely fixing the `assertUsage()`, and it breaks `concurrencyLimit`.

tperale:
@brillout to be sure I understand you correctly. You would prefer that I just move the duplicate URL checking out of the Promises block and do the check over the PageContext array after it has been completely populated ?

Regarding the concurrencyLimit it could be re-introduced by making the `allPrerenderResults ` array external

```diff
diff --git a/packages/vike/src/node/prerender/runPrerender.ts b/packages/vike/src/node/prerender/runPrerender.ts
index e857bd1c3..63465331a 100644
--- a/packages/vike/src/node/prerender/runPrerender.ts
+++ b/packages/vike/src/node/prerender/runPrerender.ts
@@ -415,12 +415,13 @@ async function callOnBeforePrerenderStartHooks(
     !doNotPrerenderList.find((p) => p.pageId === pageId)
   )

-  const allPrerenderResults = await Promise.all(
+  const allPrerenderResults = []
+  await Promise.all(
     toPreRenderList.map(async ({ pageId, ...hook }) => {
       const { hookName, hookFilePath } = hook
       const prerenderResult = await execHookSingleWithoutPageContext(hook, globalContext, () => hook.hookFn())
       const result = normalizeOnPrerenderHookResult(prerenderResult, hookFilePath, hookName)
-      return { hookName, hookFilePath, result }
+      allPrerenderResults.push({ hookName, hookFilePath, result })
     })
   )
```

brillout:
Maybe I'm missing something but AFAICT it'd be a good minimal fix, yes.

(I ain't against refactoring but it should have a motivation on its own instead of merely being a side effect of fixing assertUsage().)

tperale:
Updated

PR #3166, opened 2026-03-23
landing page: fix section nav progress + optimize blur dot images

richard-unterberg:
yes


brillout:
Any reason you closed the PR? Can I merge it?

brillout:
I think the following is weird and that's why I made the cut off earlier (I understand clicking on a USP doesn't lead a clean progress bar then).
image

brillout:
Perfect solution would be the progress bar to be stuck at `0` for some time, but not sure how to implement that.

brillout:
Not only did I revert but I doubled down and went for `70%`...

Because, actually, the following *is* the accurate progress bar: it's the amount of section already *visible*. So it's perfect IMO.

image

PR #3206, opened 2026-04-13
feat: auto inject @universal-deploy/netlify when @netlify/vite-plugin is detected

magne4000:


magne4000:
I'm thinking that this should probably directly be part of `@universal-deploy/vite` auto plugin 🤔

brillout:
> I'm thinking that this should probably directly be part of `@universal-deploy/vite` auto plugin 🤔

Good point, as you wish.

PR #3218, opened 2026-04-16
feat: support server string

rtritto:
Add string support to server in vike config.

Related: https://github.com/vikejs/vike/discussions/3201#discussioncomment-16576569

### DIFF
--- a/packages/vike/src/node/vite/plugins/pluginUniversalDeploy/getServerConfig.ts
+++ b/packages/vike/src/node/vite/plugins/pluginUniversalDeploy/getServerConfig.ts
@@ -14,7 +14,7 @@ function getServerConfig(vikeConfig: VikeConfigInternal) {
   // universal-deploy support must be manually enabled
   const serverConfig: boolean =
     // +config.js > `export default { server: true }`
-    vikeConfig.config.server ||
+    !!vikeConfig.config.server ||
     // +server.js exists
     !!vikeConfig._pageConfigGlobal.configValueSources.server ||
     false
@@ -26,6 +26,9 @@ function getServerConfig(vikeConfig: VikeConfigInternal) {
     assert(serverFilePath)
     serverEntryId = serverFilePath
     serverEntryVike = serverFilePath
+  } else if (!vikeConfig._pageConfigGlobal.configValueSources.server) {
+    serverEntryId = vikeConfig.config.server
+    serverEntryVike = vikeConfig.config.server
   } else {
     serverEntryId = catchAllEntry
     serverEntryVike = 'vike/fetch'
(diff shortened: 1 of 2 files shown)


brillout:
Can you try to use an import string instead? AFAICT that should work.

rtritto:
@brillout I already did a try, it doesn't work.
With the current implementation
https://github.com/vikejs/vike/blob/601b60c664feec73646366a935df7d85375b7152/packages/vike/src/node/vite/plugins/pluginUniversalDeploy.ts#L34
`universalDeploy` will no receive the option parameter `{ node: { importer: serverEntryId } }` and [ud:node:entry](https://github.com/universal-deploy/universal-deploy/blob/9016de0dc6f2c61425380282a010da80eb0ebbe3/packages/adapter-node/src/vite.ts#L27) goes on as default using `@universal-deploy/node/serve`.
This is a next step (maybe can be done also in this PR).

---

This PR:
- allows to pass the server as string (absolute path)
- should fail only because `vikeConfig.config.server` overlaps types (`boolean | undefined` instead of `boolean | string | undefined`)

rtritto:
@magne4000 WDYT?

rtritto:
> This PR:
> 
>     * allows to pass the server as string (absolute path)
> 
>     * should fail only because `vikeConfig.config.server` overlaps types (`boolean | undefined` instead of `boolean | string | undefined`)

@brillout is the overlap types correct?

Please can you explain why did you close this PR?

brillout:
If you can explain in a succinct and clear way why you want this, I might re-consider this.

rtritto:
@brillout If you pass an absolute path of the server file (string value) as importer at `universalDeploy` (`universalDeploy({ node: { importer: serverEntryId } })`), the plugin [ud:node:entry](https://github.com/universal-deploy/universal-deploy/blob/9016de0dc6f2c61425380282a010da80eb0ebbe3/packages/adapter-node/src/vite.ts#L27) will use that value instead of the default `@universal-deploy/node/serve`.

I already tested this logic by patching `@universal-deploy/node` (with https://github.com/universal-deploy/universal-deploy/pull/23#issuecomment-4260569899 and the absolute path) and the vike build works.

---

The main obstacle (the purpose of this PR) is how to pass (I thinked from `vikeConfig.config.server`) the absolute path to `getServerConfig` (where if it's defined will be returned as `serverEntryId`).

How can I do with vike?

---

FYI @magne4000 @nitedani

brillout:
I ain't that familiar with UD.

From Vike's perspective (that's where I'm most familiar with), what exactly are you trying to achieve with this PR? What is currently happening, what should happen instead?

PR #3224, opened 2026-04-20
Update title for Universal Deploy heading

NilsJacobsen:
I was looking for the release blog for "+server.js" but couldn't find it. Then I noticed that it is called "Universal Deploy" good framing but I was serching for "+server.js" so I couldn't find it at first.

### What I changed

- Changed the headline in the navigation

Introducing Universal Deploy -> Introducing Universal Deploy `+server.js`

### DIFF
--- a/docs/headings.ts
+++ b/docs/headings.ts
@@ -758,7 +758,7 @@ const headings = [
   },
   {
     level: 2,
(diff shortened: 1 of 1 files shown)


brillout:
Makes sense!

Issue #1815, opened 2024-08-18
Duplicated CSS


brillout:
### Description

Original issue: https://github.com/vikejs/vike/discussions/1811.

Reproduction: https://github.com/brillout/vike-reprod-duplicated-css-1815.

The root cause is that Vite builds the client- and server-side in two separate builds. Vike has to merge the CSS of the client- and server-side, while avoiding duplication based on the CSS file hashes but this doesn't work if Rollup bundles the CSS differently between the client- and server-side.

The root cuase will probably be fixed once Vite uses Rolldown. (Assuming Rolldown will build both environments at the same time.)

In the meantime, the only workaround I can think of is to disable some of the CSS bundling. See the reproduction's readme.

brillout:
Reproduction: https://github.com/brillout/vike-reprod-duplicated-css-1815.

The root cause is the fact that Vite builds the client-side and the server-side in two separate builds. This will probably be fixed once Vite uses Rolldown, assuming Rolldown builds both environments at the same time.

In the meantime, the only workaround I can think of is to disable CSS bundling. See the reproduction's readme.

brillout:
> the only workaround I can think of is to disable some of the CSS bundling

AFAICT CSS bundling needs to be entirely disabled:
- Each CSS module needs to be included in the exact same CSS bundle with the same content.
  - Determining CSS bundles based on the JavaScript importers doesn't work as it can lead to different bundles: a CSS module may have different JavaScript importers on the client-side than on the server-side.

princescar:
Hi @brillout, this is still happening with a setup of latest Vike + Hattip(not sure if it's related). While I import bootstrap css in root layout, it produces two entry in the entry.mjs field for css.

Is the issue completely fixed or it gets improved by now?

princescar:
One workaround for me is to import the global css in +onBeforeRenderHtml instead of +Layout or +Page.

brillout:
@princescar Minimal reproduction welcome.

princescar:
I did playaround with a minimal Vike + Hattip setup, it seems the issue is not with Vike. It only exists when I set server build target to "esnext" (in order to use top-level await in server code).

```
export default defineConfig({
  plugins: [
    react(),
    vike(),
    hattip({
      serverConfig: {
        build: {
          target: "esnext",
        },
      },
    }),
  ],
});
```

Then the hash would be different for client and server build:

```
// Client
[...]

PR #3230, opened 2026-04-22
fix: skip `manualChunks` if Vite 8's `codeSplitting` is used (#3227)

brillout:
See also:
- https://github.com/vikejs/vike/issues/1815
- https://github.com/orgs/vikejs/discussions/3227


brillout:
Pre-released in `0.4.258-commit-e168602`.

PR #3235, opened 2026-04-26
fix: eagerly expose prod.* options from +server.ts on virtual:ud:catch-all

### LINKED ISSUE #3232: `+server.ts` `prod.onCreate` / `prod.onReady` hooks and module-level init code never run at startup
## Description

When using `+server.ts` with a server framework (e.g. Hono), the `prod.onCreate` and `prod.onReady` hooks documented in [the `+server` guide](https://vike.dev/server#settings) are **never invoked**. Additionally, any module-level `await` expressions (i.e. startup initialization code) inside `+server.ts` are deferred until the first HTTP request, not run at server startup.

## Minimal Reproduction

https://github.com/YKDZ/vike-server-hooks-repro

```sh
git clone https://github.com/YKDZ/vike-server-hooks-repro
cd vike-server-hooks-repro
npm install
npm run build
npm run preview
```

## Root cause

Looking at the generated `dist/server/index.mjs` (from `@universal-deploy/node`):

```js
// virtual:ud:catch-all
var __map = { "m0": () => import("./chunks/chunk-ifPvdNCz.js") };

var virtual_ud_catch_all_default = {
  async fetch(request, ...args) {
    // lazy-loads entry module only on first request
    return assertFetchable(await __map["m0"]()).fetch(request, ...args);
  }
};

async function startServer() {
  const server = serve({
    ...virtual_ud_catch_all_default,  // only has `fetch` — no onCreate/onReady/port/hostname
    manual: true,
  });

  virtual_ud_catch_all_default.onCreate?.(server);   // ← always undefined, never called
  server.serve();
  await server.ready();
  virtual_ud_catch_all_default.onReady?.(server);   // ← always undefined, never called
}
```

`virtual_ud_catch_all_default` is a hand-crafted object with only a `fetch` method. The actual `+server.ts` export (which contains `prod.onCreate`, `prod.onReady`, `prod.port`, `prod.hostname`, etc.) lives inside the lazy entry chunk and is **only loaded on the first incoming HTTP request**.

As a result:

1. **`prod.onCreate` / `prod.onReady` are never called** — `virtual_ud_catch_all_default.onCreate` is `undefined`.
2. **Module-level startup code runs lazily** — any top-level `await` in `+server.ts` (e.g. database migrations, initializing connections) is deferred until the first request, not at server startup.
3. **`prod.port` and `prod.hostname` are not forwarded** to `srvx`'s `serve()` — only the default values are used.

## Expected output (`node dist/server/index.mjs`)

```
[+server.ts] module-level code executed
[onCreate] called! node?.server: true
Listening on: http://localhost:4000
[onReady] called! url: http://localhost:4000
```

## Actual output

```
Listening on: http://localhost:3000
```

No hooks are called, port `4000` is ignored, server listens on `3000`.

## Environment

- vike: `0.4.258`
- `@universal-deploy/node`: `0.1.6`
- `@vikejs/hono`: `0.2.0`
- hono: `4.12.15`
- vike-vue: `0.9.11`
- vite: `8.0.10`
- Node.js: `24`

## Expected fix

`startServer()` in `@universal-deploy/node` should eagerly load the entry module to obtain its full export (including `onCreate`, `onReady`, `port`, `hostname`) before calling `serve()` and `server.serve()`. The `virtual:ud:catch-all` module could expose these properties by eagerly importing the server entry, or the generator could emit them directly onto `virtual_ud_catch_all_default`.

YKDZ:
## Summary

>  This PR was generated with AI assistance as my temporary local workaround. Keeping it as a draft since the approach may not be the most suitable for vike.

Fixes #3232.

---

### Problem

`@universal-deploy/node`'s `startServer()` calls `virtual_ud_catch_all_default.onCreate?.(server)` and `.onReady?.(server)` at startup. However, the `prod.*` hooks defined in `+server.ts` are only reachable *after* the lazy entry chunk is loaded on the first HTTP request — so `onCreate` and `onReady` are always `undefined` at server startup.

The same applies to `prod.port` and `prod.hostname`, which are never forwarded to `srvx`'s `serve()`.

### Root cause

`@universal-deploy/vite`'s `catchAll()` plugin generates `virtual:ud:catch-all` as:

```js
export default {
  async fetch(request, ...args) {
    // lazy-load the actual handler on first request
    return (await import('./chunks/entry.js')).fetch(request, ...args)
  }
}
```

The `prod.*` options live inside the lazy chunk, not on this top-level object.

### Fix

Add a new Vite plugin **`pluginCatchAllProdOptions`** that transforms `virtual:ud:catch-all` (post-load) to eagerly import the user's `+server.ts` and spread its exports onto the default export:

```js
// generated after patch
import __vikeServerMod from '/path/to/+server.ts?wrapper_...'  // eager
export default {
  ...__vikeServerMod,   // port, onCreate, onReady, hostname, …
  async fetch(request, ...args) { /* routing */ }
}
```

`pluginUnwrapProdOptions` already wraps `serverFilePath` so that its default export is `{ ...originalDefault, ...originalDefault.prod }`, so no additional spread is needed.


packages/vike/src/node/vite/plugins/pluginUniversalDeploy.ts
```diff
@@ -60,5 +61,6 @@ function pluginUniversalDeploy(vikeConfig: VikeConfigInternal): Plugin[] {
     pluginServerEntryInject(serverFilePath ?? serverEntryId),
     pluginServerEntryAlias(serverFilePath),
     !serverFilePath ? null : pluginUnwrapProdOptions(serverFilePath),
+    !serverFilePath ? null : pluginCatchAllProdOptions(serverFilePath),
```
rtritto:
```suggestion
    ...serverFilePath ? [pluginUnwrapProdOptions(serverFilePath), pluginCatchAllProdOptions(serverFilePath)] : [],
```

brillout:
@magne4000 I guess this fix should live in UD instead of Vike?

magne4000:
@brillout Indeed it should, but it will be a bit more subtle. `catch-all` currently dynamically imports UD entries in the `fetch` handler. We need eagerly import UD entries when only one is found so that it can be re-exported.
I should have time to fix it either on sunday or beginning of next week.
In the mean time, this minimal plugin is a totally valid temporary workaround that could be implemented directly in `vite.config.ts`.

magne4000:
Fixed in `@universal-deploy/vite@0.1.8`

PR #3264, opened 2026-05-18
docs: start using `+choices`

phonzammi:
Let's start using `+choices` for vike.dev. 
@brillout, I haven't applied it to all pages yet, but please have a look and let me know what you think.

### DIFF
--- a/docs/+docpress.tsx
+++ b/docs/+docpress.tsx
@@ -53,11 +53,11 @@ const config = {
       default: 'Hono',
     },
     uiFramework: {
-      choices: ['React', 'Solid', 'Vue'],
+      choices: ['React', 'Vue', 'Solid'],
       default: 'React',
     },
     uiFrameworkVikeExtension: {
-      choices: ['vike-react', 'vike-solid', 'vike-vue'],
+      choices: ['vike-react', 'vike-vue', 'vike-solid'],
       default: 'vike-react',
     },
   },
--- a/docs/+docpress.tsx
+++ b/docs/+docpress.tsx
@@ -47,4 +47,18 @@ const config = {
     fontSize: 21,
     marginLeft: 0,
   },
+  choices: {
+    server: {
+      choices: ['Hono', 'Express.js', 'Fastify', 'H3', 'Elysia', 'Other'],
+      default: 'Hono',
+    },
+    uiFramework: {
+      choices: ['React', 'Solid', 'Vue'],
+      default: 'React',
+    },
+    uiFrameworkVikeExtension: {
+      choices: ['vike-react', 'vike-solid', 'vike-vue'],
+      default: 'vike-react',
+    },
+  },
 } satisfies Config
--- a/docs/+docpress.tsx
+++ b/docs/+docpress.tsx
@@ -56,9 +56,5 @@ const config = {
       choices: ['React', 'Vue', 'Solid'],
       default: 'React',
     },
-    uiFrameworkVikeExtension: {
-      choices: ['vike-react', 'vike-vue', 'vike-solid'],
-      default: 'vike-react',
-    },
   },
 } satisfies Config
--- a/docs/components/DeprecatedServer.mdx
+++ b/docs/components/DeprecatedServer.mdx
@@ -1,7 +1,7 @@
 import { Warning } from '@brillout/docpress'
 
 <Warning>
-  Because {props.name} doesn't follow modern JavaScript standards (such as `fetch`), it currently doesn't work with the following:
+  Because {props.name} doesn't follow modern JavaScript standards, it currently doesn't work with the following:
    - Bun
    - Deno
    - Cloudflare
--- a/docs/package.json
+++ b/docs/package.json
@@ -7,7 +7,7 @@
   },
   "dependencies": {
     "@batijs/elements": "^0.0.64",
-    "@brillout/docpress": "0.16.35-commit-3867953",
+    "@brillout/docpress": "^0.16.36",
     "@classmatejs/react": "^0.2.1",
     "@gsap/react": "^2.1.2",
     "@tailwindcss/vite": "^4.1.18",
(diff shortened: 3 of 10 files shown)


brillout:
Little bug:
image

We can merge first and fix later. As you wish.

brillout:
> @brillout, I haven't applied it to all pages yet, but please have a look and let me know what you think.

It's great, I love it.

brillout:
Little thing: when I go to http://localhost:3000/server#get-started and refresh the page, the dropdowns aren't visible for 1-2 seconds (I guess until the page is hydrated). If it's easy to do, then I'd say let's show the dropdowns as early as possible (maybe we can use our FOUC-avoiding mechanism).

brillout:
Btw. I think we should rename `Express.js` to `Express` (but we can do that later in another PR).

brillout:
For `/ClientOnly` I think we should re-use the choice "React" / "Vue" / "Solid"

brillout:
> "React" / "Vue" / "Solid"

In this order: `Solid` should be last (it's the least popular choice).

brillout:
Issue:
image

Ideas:
- Easy fix: we duplicate the code for Solid
- Potential future-long term fix: a code block can be valid for multiple choices.

I'd say the long term fix only if it's a quick-win (but *really* only if it's a quick-win). I think the easy fix is probably better here.

brillout:
Actually, wait, Solid doesn't implement `bodyHtmlEnd` so we should skip `Solid` here (I guess we can still re-use the same +choices group though).

brillout:
Hiding sections like `### Vue` at `/Head` is too implicit & brittle. The best would be to show a new `<Tabs>` here, but not sure how it would work (the choice is "Vue" or "React / Solid"). Maybe we need a new feature here. In the meantime we can leave things as it is.

brillout:
Done with my review. Love it :heart_eyes: can't wait to merge :sparkles:

phonzammi:
> Little bug: image
> 
> We can merge first and fix later. As you wish.

Fixed in: https://github.com/brillout/docpress/commit/3867953fceca31858a9bba7c78759337f275f845, and updated to commit release in https://github.com/vikejs/vike/pull/3264/commits/b4364e497a6a3fd1cce8891219bd54b3cbc9c54b

phonzammi:
Just pushed a few commits. If everything looks good, let’s release the new docpress, update it, and merge this PR.

brillout:
> use `uiFramework` choice instead of `uiFrameworkExtension` for `/ClientOnly`

I'd say also for the other pages.

phonzammi:
> For `/ClientOnly` I think we should re-use the choice "React" / "Vue" / "Solid"

I have been thinking about `uiFramework` & `uiFrameworkVikeExtension` choices. 

What if we merge them into a single group like this?

```ts
uiFramework: {
      choices: ['React', 'Vue', 'Solid'],
      extensionChoices: ['vike-react', 'vike-vue', 'vike-solid'], // optional
      default: 'React',
    },
```

We can use code block meta and `containerDirective` attributes like `extension=true`.

We can add this feature later.

phonzammi:
> > use `uiFramework` choice instead of `uiFrameworkExtension` for `/ClientOnly`
> 
> I'd say also for the other pages.

Done: [87cadcc](https://github.com/vikejs/vike/pull/3264/commits/87cadcce17289740ef2c7162cff37eeecc045ffc)

brillout:
> What if we merge them into a single group like this?

What use case do you have in mind?

brillout:
Correct me if I missed something, but I think we can now do https://github.com/vikejs/vike/pull/3264/commits/22530b8df053ea6c7df2a1adc3d49a6b49a32c58.

brillout:
> Hiding sections like `### Vue` at `/Head` is too implicit & brittle. The best would be to show a new `<Tabs>` here, but not sure how it would work (the choice is "Vue" or "React / Solid"). Maybe we need a new feature here. In the meantime we can leave things as it is.

https://github.com/vikejs/vike/pull/3264/commits/c85aa15b8a816259a18e71ca51c9d71f323a9a08

brillout:
> Actually, wait, Solid doesn't implement `bodyHtmlEnd` so we should skip `Solid` here (I guess we can still re-use the same +choices group though).

https://github.com/vikejs/vike/pull/3264/commits/5beb3506f1364d32746cd12afdcd61d78e6c91af

brillout:
> Just pushed a few commits. If everything looks good, let’s release the new docpress, update it, and merge this PR.

LGTM, feel free to self merge. (FYI I had to do a couple of `intervention++` though.)

phonzammi:
Thank you. (Sorry for the `intervention++`, maybe there were a few misunderstandings).

brillout:
@phonzammi Shall we create an issue(s) for the remaining issues/improvements?

phonzammi:
Yes, that would be helpful for tracking.

brillout:
@phonzammi Can you do it?

phonzammi:
Sorry, I’m unsure which issues to create.

phonzammi:
These issues ?

> Little thing: when I go to http://localhost:3000/server#get-started and refresh the page, the dropdowns aren't visible for 1-2 seconds (I guess until the page is hydrated). If it's easy to do, then I'd say let's show the dropdowns as early as possible (maybe we can use our FOUC-avoiding mechanism).

> Btw. I think we should rename `Express.js` to `Express` (but we can do that later in another PR).

brillout:
I believe so, and this:

> Actually, wait, Solid doesn't implement `bodyHtmlEnd` so we should skip `Solid` here (I guess we can still re-use the same +choices group though).

I reverted the changes here. Ideas for a solution?

phonzammi:
> Actually, wait, Solid doesn't implement `bodyHtmlEnd` so we should skip `Solid` here (I guess we can still re-use the same +choices group though).

How about we don't show that choice or any other choice with this use case ? 
Previously (https://github.com/brillout/docpress/pull/128#issuecomment-4366156338), we already disabled it in the dropdown. How about we also remove it from the tab list and the dropdown?

brillout:
:+1: Sounds good.

PR #3293, opened 2026-05-25
fix: do not force exit when using universal-deploy

### LINKED ISSUE #3292: pnpm build does not populate .vercel/output
### Description

1. `pnpm create vike@latest --react --tailwindcss --google-analytics --vercel --eslint --sentry`
3. `pnpm install`
4. `pnpm build`
5. inspect output .vercel/

expected behavior:
.vercel/output should be populated

actual behavior:
.vercel/ directory is empty

- pnpm info create-vike version
0.0.628

magne4000:
Fixes #3292
UD can build other Vite environments. When a `server` is configured, we know that we are following UD's build path, so we disabled force exit.

### DIFF
--- a/packages/vike/src/node/prerender/runPrerenderEntry.ts
+++ b/packages/vike/src/node/prerender/runPrerenderEntry.ts
@@ -32,7 +32,10 @@ async function runPrerenderFromCLIPrerenderCommand(): Promise<void> {
   runPrerender_forceExit()
   assert(false)
 }
-async function runPrerenderFromAutoRun(viteConfig: InlineConfig | undefined): Promise<{ forceExit: boolean }> {
+async function runPrerenderFromAutoRun(
+  viteConfig: InlineConfig | undefined,
+  isServerConfig: boolean,
+): Promise<{ forceExit: boolean }> {
   try {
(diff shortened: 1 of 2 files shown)


brillout:
The reason Vike force exists is because after pre-rendering the Node.js process often doesn't close on its own because of some kind of pending work in the event loop — the Node.js process must be forcefully closed.

Can we (Vike or UD) forcefully close the process after UD is done?

magne4000:
> The reason Vike force exists is because after pre-rendering the Node.js process often doesn't close on its own because of some kind of pending work in the event loop — the Node.js process must be forcefully closed.
> 
> Can we (Vike or UD) forcefully close the process after UD is done?

Perhaps we can in a `enforce: 'post'` + `order: 'post'` `buildApp` hook. Let me try that

magne4000:
> Perhaps we can in a `enforce: 'post'` + `order: 'post'` `buildApp` hook. Let me try that

Doesn't seem to work. It even triggers only the `post` hook, and the existing `pre` hook is not even triggered (seems like a Vite bug)

magne4000:
Replaced by https://github.com/vikejs/vike/pull/3293

PR #3336, opened 2026-06-13
Deduplicate import statements across pages

brillout:
## Summary
Refactor import statement handling to deduplicate imports across pages. When the same module export is used by multiple pages (e.g., a shared `+route.ts` inherited by N pages), the generated code now reuses a single import variable instead of emitting duplicate import declarations.

## Key Changes
- **New `ImportStatements` class**: Replaces the simple `string[]` array with a dedicated class that tracks both the import statements and a deduplication map keyed by `${importPath}::${exportName}`.
  - `getExisting()`: Returns the import variable name if the same (importPath, exportName) pair was already imported
  - `add()`: Adds a new import statement and registers it in the deduplication map
  - `toArray()`: Returns the list of import statements for code generation

- **New `createImportStatements()` factory function**: Provides a clean API for instantiating the deduplication tracker

- **Updated `addImportStatement()` function**: Now checks for existing imports before creating new ones, reusing the import variable when available

- **Type signature updates**: Changed all `importStatements: string[]` parameters to `importStatements: ImportStatements` throughout the codebase

- **V1 design optimization**: Added early return in `generateVirtualFileGlobalEntryWithOldDesign.ts` when using V1 design to skip unnecessary glob operations

## Implementation Details
- The deduplication map uses a composite key format (`${importPath}::${exportName}`) to uniquely identify import sources
- Import variable names remain consistent (`import1`, `import2`, etc.) based on the order of first occurrence
- The change is backward compatible and transparent to consumers of the serialized config values

https://claude.ai/code/session_0194Javu51nwH7cthohuX9X5

### DIFF
--- a/packages/vike/src/node/vite/plugins/pluginVirtualFiles/generateVirtualFileGlobalEntry.ts
+++ b/packages/vike/src/node/vite/plugins/pluginVirtualFiles/generateVirtualFileGlobalEntry.ts
@@ -37,7 +37,7 @@ function getCode(
   const filesEnv: FilesEnv = new Map()
 
   if (!isForClientSide) {
-    importStatements.push(`import '${VIRTUAL_FILE_ID_constantsGlobalThis}';`)
+    importStatements.list.push(`import '${VIRTUAL_FILE_ID_constantsGlobalThis}';`)
   }
 
   lines.push('export const pageConfigsSerialized = [')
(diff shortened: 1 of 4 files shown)


brillout:
Closing: measured against the docs (351 pages) and the production build (`pnpm run build`) is bit-for-bit identical with and without this change — same content hashes on every JS file, same total byte count.

The dedup only matters in dev mode, where virtual modules are served raw. In a production build, Rollup already deduplicates imports at the bundler level, so N identical `import { x as import1 }` / `import { x as import2 }` statements in the virtual module source collapse to a single reference in the output bundle anyway.

---

PR #3337, opened 2026-06-13
perf: skip old-design glob imports for V1-design projects

brillout:
## What

`generateVirtualFileGlobalEntryWithOldDesign()` always emits up to 7 `import.meta.glob()` calls — covering `.page`, `.page.client`, `.page.server`, `.page.route` × (import + exportNames + extractAssets) — into the `global-entry` virtual module, regardless of design.

In a **V1-design** project (using `+config.ts` files) there are no `.page.*` files, so every one of those globs expands to an empty object. Zero benefit, but Vite still scans the project tree (glob root is `/`) for each pattern.

## Change

Check `isV1Design()` before the glob block and return early when the project uses only the new design. The empty-object exports (`pageFilesLazy`, `pageFilesEager`, etc.) are still emitted above the early return, so the module's public shape is unchanged.

Old-design and mixed-design projects are unaffected (`isV1Design()` is `false` for them).

## Context

Split out from #3336 (the import-deduplication change) so the two perf improvements can be reviewed and merged independently.

https://claude.ai/code/session_0194Javu51nwH7cthohuX9X5

---
_Generated by [Claude Code](https://claude.ai/code/session_0194Javu51nwH7cthohuX9X5)_

### DIFF
--- a/packages/vike/src/node/vite/plugins/pluginVirtualFiles/generateVirtualFileGlobalEntryWithOldDesign.ts
+++ b/packages/vike/src/node/vite/plugins/pluginVirtualFiles/generateVirtualFileGlobalEntryWithOldDesign.ts
@@ -17,7 +17,7 @@ import { version as viteVersion } from 'vite'
 import { type FileType, fileTypes } from '../../../../shared-server-client/getPageFiles/fileTypes.js'
 import path from 'node:path'
 import { generateVirtualFileGlobalEntry } from './generateVirtualFileGlobalEntry.js'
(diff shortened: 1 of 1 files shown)


brillout:
Closing: measured against the docs (351 pages) and the production build (`pnpm run build`) is bit-for-bit identical with and without this change — same content hashes on every JS file, same total byte count.

In a V1-design project the glob patterns match zero files, so they expand to empty objects `{}` at build time and Rollup tree-shakes them away entirely. The `import.meta.glob()` calls only have visible cost in dev mode (where Vite scans the filesystem for each pattern), not in the production bundle.

---

PR #3348, opened 2026-06-16
fix: load +config.server with pointer import

### LINKED ISSUE #3341: `server` in `+config` ignores the pointer import
### Description

In the `+config` using the `server` option with the value as [pointer import](https://vike.dev/config#pointer-imports) to import the server, the server isn't loaded by vike:

```ts
// pages/+config.ts

export default {
  server: 'import:../server/index.ts:default'
} satisfies Config
```

### Reproduction

- clone [minimal-server-vercel](https://github.com/templates-ecosystem/template-vike-solid-daisyui-hono/tree/minimal-server-vercel) (commit [79ac620](https://github.com/templates-ecosystem/template-vike-solid-daisyui-hono/commit/79ac620faadbd99a41077e6d87f35b3314811205))
- `pnpm install`
- `pnpm dev`
- `http://localhost:3000/api/hello`

### Actual

```sh
[vike][request-1] HTTP response ← /api/hello 404
```

### Excpeted
Route `/api/hello` correctly dispatched and no 404 in log

### Use case

#3340

rtritto:
Fix #3341

### Changes
- [x] resolve `+config.server` with pointer import
- [ ] load server app from `+config.server` instead of the default one

### DIFF
--- a/packages/vike/src/node/vite/shared/resolveVikeConfigInternal/metaBuiltIn.ts
+++ b/packages/vike/src/node/vite/shared/resolveVikeConfigInternal/metaBuiltIn.ts
@@ -353,7 +353,7 @@ const metaBuiltIn: ConfigDefinitionsBuiltIn = {
     global: true,
   },
   server: {
-    env: { server: true },
+    env: { server: true, config: true },
     global: true,
   },
   cli: {


rtritto:
@magne4000 I found the issue where `+config.server` is always undefined and fixed it with adding `config: true` in `metaBuiltIn.ts`. By this way `+config.server='import:../server/index.ts:default'` will load the server app in `vikeConfig.config.server`.
Now Vike doesn't replace the default server app with the `vikeConfig.config.server`.

magne4000:
Oh yeah I forgot that we removed that. But it causes the issue you're seeing with Cloudflare.
I'm not sure how we should fix that. @brillout for this particular case, we would want the pointer import to be resolved, but not imported:

  | config: false | config: true | what we need
-- | -- | -- | --
pointer imports parsed | ❌ | ✅ | ✅
pointer imports auto imported | ❌ | ✅ | ❌

We currently rely on this syntax to have Vike resolve the absolute path.

brillout:
Hm, I see. I guess we could make pointer imports always available at config time.

Let me ask Claude.

brillout:
https://claude.ai/code/session_013APNNzqkpqD16jWYfmgJ4B (should be public)

brillout:
(Claude here, answering @brillout's ping.)

The key distinction is **resolve vs. import** — two separate steps for a pointer import:

- **Resolve** (`import:../server/index.ts:default` → absolute path): pure `require.resolve`, **no code executed** (`resolvePointerImport.ts` → `requireResolve.ts`). This already runs for *every* pointer import regardless of `env`, and the result is stored on the config value source's `definedAt` (`pointerImport.fileExportPath`).
- **Import** (execute the file to read its export): gated by `shouldBeLoadableAtBuildTime() = !!env.config && !_valueIsFilePath` (`loadFileAtConfigTime.ts`). This is the step that runs the server entry → breaks Cloudflare.

So `config: true` on its own flips on *both* "surface at config time" *and* "execute" — that's why the auto-import shows up in @magne4000's table. But the "resolve-but-don't-import" behavior you want **already exists**: it's exactly how `+client` works:

```ts
// metaBuiltIn.ts
client: {
  env: { server: true, config: true },
  _valueIsFilePath: true, // ← surface the resolved path at config time, never execute
},
```

With `_valueIsFilePath`, `shouldBeLoadableAtBuildTime()` returns `false` (file never executed) and `getConfigValueSources()` sets the config-time value to the resolved path (the `if (configDef._valueIsFilePath)` branch in `resolveVikeConfigInternal.ts`).

Extending @magne4000's table:

| | `env:{server}` (current) | `+ config:true` (this PR) | `+ config:true` + resolve-only |
|---|:--:|:--:|:--:|
| resolved to abs. path | ✅ (only on `definedAt`) | ✅ | ✅ |
| surfaced in `config.server` | ❌ | ✅ (executed value) | ✅ (abs. path) |
| server entry executed | ❌ | ✅ → breaks CF | ❌ |

**Recommendation:** keep `config: true` (you do need it — otherwise `config.server` isn't populated at config time at all, since `isForConfig` only matches sources with `env.config`), but pair it with resolve-only instead of auto-import. Two ways:

1. **Reuse `_valueIsFilePath` on `server`.** Minimal. Caveat: the current branch assumes the value is *exactly one* pointer import, whereas `server` is `boolean | ImportString | ImportString[]` — so it'd need to (a) pass `boolean` values through unchanged and (b) handle the array form. `+client` never hits these because it's always a single path.
2. **A dedicated opt-in flag** (a "resolve pointer imports but don't import them" meta) if you'd rather not overload `_valueIsFilePath`'s semantics.

I'd avoid making it a *global* rule ("pointer imports are never imported at config time") — some `env.config` configs legitimately want the executed value at config time (e.g. a `+vite` plugin, or a `route` function, provided via pointer import). A per-config opt-in keeps those working.

**Downstream** this matches the PR's 2nd checkbox: the resolved entry is on `configValueSources.server[].definedAt.filePathAbsoluteFilesystem` (path) + `.fileExportName` (export), so `getServerConfig()` / vike-server can load the server app in the right environment instead of Vike executing it during config resolution.

Happy to push a PoC for option 1 if useful.

---

Issue #1525, opened 2024-02-27
Generate file with arbitrary content at arbitrary output path


iacore:
### Description

I hope vike can generate files that is not HTML.

The requirements are

- The generation script need to work inside vite context (need access to vite API like `import.meta.glob`)
- Work in development mode (can see the file on development server)
- Work in production build (the file is generated correctly with a chosen filename without .html)

Related: #1522 #1524 

## Use Cases

- Generate Atom feed from blogposts
- Generate sitemap #49

iacore:
`onRenderContent` is best, I think. Still need to offer an option to set the output path, though. It would be nice if the API allows access to `pageContext` from other files like page title, which will make sitemap generation easier.

> Also, what do other frameworks do to support these two use cases? Can someone dig into that?


https://www.11ty.dev/docs/plugins/rss/

[lume](https://lume.land/) allows generating arbitrary file content. lume has a [sitemap](https://lume.land/plugins/sitemap/) plugin.

To set the page URL in lume, you simply do

```ts
export const url = "/feed.json";
```

or 

```
---
url: /blog/example.md
---

# Page Content Here
```

brillout:
> Still need to offer an option to set the output path, though.

Or how about using Filesytem Routing as usual?

> It would be nice if the API allows access to `pageContext` from other files

You shouldn't do that in dev, see https://vike.dev/markdown#metadata. Upon pre-rendering you can already access that, although it's currently hidden behind an internal API: `pageContext._prerenderContext.pageContexts`.

It makes sense that SSGs like Lume and Eleventy supports that use case. I was more wondering about SSR frameworks such as Next.js or Nuxt.

iacore:
> Or how about using Filesytem Routing as usual?

How do I generate `dist/client/feed.json` but not `dist/client/feed.json/index.html`?

Also, I can't name a folder `feed.json` because vite complains about trying to load the folder as JSON. (you can create `feed.json/+Page.tsx` to test it out)

iacore:
> Vike wouldn't create an `.html` file when the content is generated with `onRenderClient()`.

I need a **static** page! `onRenderClient` won't generate JSON, or can it?

> You could use a Route String.

A route string of `/feed.json` generates `/feed.json/index.html`

brillout:
@iacore It *will* create a file, it just won't be a `.html` one. If you set the Route String to end with `.json` so will the generated file.

brillout:
[...]

Issue #341, opened 2022-06-17
Single File Route


brillout:
### Description

Being able to define the entire routing of the app in a single file.

brillout:
How about following design.

```bash
/pages/index/index.page.js
/pages/about/index.page.js
/pages/product/index.page.js
/pages/product-list/index.page.js
/routes.js
```

```js
// /routes.js

export default {
  routes: [
    '/pages/index/index.page.js',
    '/pages/about/index.page.js',
    {
      path: '/pages/product/index.page.js',
      route: '/product/@id'
      /* Or a Route Function:
      route: pageContext => {
        // ...
      }
      */
    },
    {
      path: '/pages/product-list/index.page.js',
      route: '/products'
    },
  ],
  // By default all routes need to be listed in `/routes.js` (if `/routes.js` exists).
  // Set to `true` to allow routes to be skipped (we recommend against this).
  allowUnlistedRoutes: false
}
```

```js
function Menu() {
  return <>
    <Link to="index" /> // => <a href="/" />
    <Link to="about" /> // => <a href="/about" />
    <Link to="product" id={42} /> // => <a href="/product/42" />
    <Link to="product-list" /> // => <a href="/products" />
  </>
}
```

Open question: how to make it work with TypeScript? So that TypeScript complains upon `<Link to="unknown-page" />` or `<Link to="product-list" id={"42"} />` (because `"42"` is a `string` but should be a `number`).

I can see negative reactions about this design. Feel free to criticize.

redbar0n:
One option could be to use a tagged template literal string. Which has [typescript support](https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html). It would also allow to eliminate the documentation /  comment in the aforementioned example, since it would be more direct.

```jsx
function Menu() {
  return <>
    <Link to=`/` />
    <Link to=`/about` />
    <Link to=`/product/${42}` />
    <Link to=`/products` />
  </>
}
```

redbar0n:
I see there’s some precedence to the API that I suggested: https://solito.dev/usage/link#wrapper-components

brillout:
> Open question: how to make it work with TypeScript? So that TypeScript complains upon `<Link to="unknown-page" />` or `<Link to="product-list" id={"42"} />` (because `"42"` is a `string` but should be a `number`).

Does Solito has such TypeScript support?

>  One option could be to use a tagged template literal string.

So far, I can't see how TypeScript's template literal strings would help.

[...]

PR #3356, opened 2026-06-21
feat: `+pages` for programmatically defining pages (closes #1691)

### LINKED ISSUE #1691: Add ability to programatically define a file route
### Description

Current Vike API doesn't provide any way how to programmatically create a file route for situations when +Page.ts are created externally or in other dynamic fashhion

brillout:
## What

Implements the **config-time** half of #1691: define pages from code, without a `+Page`/`+config` file on the filesystem.

```js
// pages/+config.js
import ProductPage from './ProductPage'
import LoginPage from './LoginPage'

export default {
  pages: [
    { route: '/product/@id', Page: ProductPage },
    { route: '/login',        Page: LoginPage },
  ],
}
```

This unblocks the "Single Route File" request (#341), is the foundation for extension-defined pages (e.g. `vike-authjs`/`vike-better-auth`), and is the registration primitive needed for the page-variant use cases in #1525 (OG images, `.md` for LLMs).

> [!NOTE]
> Per discussion, this is **only the list form**. The function/transformer form (`pages: (pages) => [...]`) is intentionally left for a follow-up. Runtime/post-build programmatic pages are also out of scope.

## How

Each `pages` entry is turned, at config-time, into a **synthetic `+config.js`** at a namespaced `locationId` — `{definingLocationId}/(programmatic)/{slug}` — so it flows through the **existing** config-resolution pipeline unchanged, with no duplicated logic:

- The entry `{ route, Page, … }` is structurally a `+config.js` default export, so it reuses `getPlusFileFromConfigFile()`. This means `Page` is a top-level pointer import and gets resolved + serialized into a real runtime `import` for free.
- Nesting under the defining location makes the page **inherit** the surrounding config (renderer via `extends`, `title`, etc.).
- The `(programmatic)` marker is a parenthesis route-group (already a first-class Vike concept): it's stripped from the Filesystem-Routing fallback and guarantees a synthetic page can never collide with — nor become the inheritance parent of — a real page.
- `slug` is derived from `id` (optional) or the `route` string; duplicates are asserted with a clear message.

Touchpoints:
- `types/Config.ts` — `pages?: ConfigPageEntry[]` + exported `ConfigPageEntry` type; registered as a built-in global config name.
- `metaBuiltIn.ts` — `pages: { env: { config: true }, global: true, cumulative: true }` (config-time only; never shipped to runtime; definable globally or by extensions).
- `getPlusFilesByLocationId.ts` / `addProgrammaticPages.ts` — `addProgrammaticPages()` runs right after the filesystem crawl.

## Verification

- `tsc --noEmit` clean for `packages/vike` and first-party playground files.
- Unit suite green: **180/180** (`vitest --project unit`).
- Added a programmatic page to `test/playground` (+ e2e test wired into `testRun.ts`) and verified **end-to-end**, browser-free:
  - **Build/prerender**: emits `dist/.../programmatic.html` with the rendered `Programmatic Page` heading, inherits the page title (`title: 'Big Playground'`), and the `Page` component is shipped in a client chunk (hydrates via the client-routing entry). `pageId === /pages/(programmatic)/programmatic`.
  - **Dev**: `GET /programmatic` → 200 with rendered + inherited content.

## Follow-ups (not in this PR)
- Function/transformer form `pages: (pages) => [...]` (the #1525 variant story).
- Docs page `https://vike.dev/pages`.

Closes part of #1691.

https://claude.ai/code/session_01FMtgpc4sFVWUa7PPo1U92z


suleimansh:
Small thing, the description says it exports PageConfigEntry, but the code
exports ConfigPageEntry. Worth aligning the two (and the upcoming docs) on one name.

brillout:
Good catch — aligned the PR description with the code. The exported type is `ConfigPageEntry` (the rename from `PageConfigEntry` landed in a later commit); that's the canonical name to use for the upcoming docs too. Thanks!

---

brillout:
## `config.pages` — known limitations (first iteration)

For reviewers/users, the current scope and intentional limitations of programmatically defined pages:

1. **Config-time only.** `config.pages` is resolved at config-time. The function/transformer form (`pages: (pages) => [...]`) and runtime/CMS-driven (post-build) pages are **out of scope** for now.
2. **Route Strings only.** A page entry's `route` must be a string. Route Functions are rejected with an assertion for now (`ConfigPageEntry.route` is typed `string`).
3. **`pages` is a global config.** Define it at a global location (e.g. `/pages/+config.js` or `/+config.js`, or via an extension). Defining it at a non-global/scoped location still works but emits the usual "global config at a non-global location" warning.
4. **Identity is positional.** A page's `pageId` is `…/(programmatic)/<index>` (the entry's index, scoped per defining location). Reordering/inserting/removing entries changes the `pageId`s of later entries — and thus `globalContext.pages` keys and client chunk filenames. **URLs and pre-rendered output are unaffected** (they're route-based). There is intentionally no per-entry `id`.
5. **No `ImportStringList` for `pages` or `route`.** `pages` must be an inline array (or a `+pages.js` value file); `route` must be an inline string.
6. **Config-time values via pointer import inside an entry** (e.g. an imported value for a `env: { config: true }` config) aren't loaded at config-time, so config-time readers see them as absent. Inline config-time values and all runtime imports (`Page`, `data`, hooks, …) work normally. (The main case — route functions — is already covered by #2.)
7. **`+config.js` `Page` resolution rules are unchanged.** A `Page` (or other component) referenced by an entry must be a pointer import — i.e. a `.jsx/.tsx/.vue` file (or any non-plain-script). A plain `.ts/.js` component isn't a pointer import (same as filesystem pages).
8. **`getVikeConfig().config.pages`** (build-time) reflects the raw entries (including internal import strings for `Page`); use `getVikeConfig().pages` (the resolved page map) for the resolved pages.

Everything else (config inheritance, `meta.effect`, computed values, precedence, routing, pre-rendering, client-side hydration, both `config.pages` and `+pages.js` forms, and extension-defined pages) goes through the same pipeline as filesystem pages.

---

PR #3376, opened 2026-06-22
sync-github-releases: rename `try`→`check`, sync all packages by default, run with Node

brillout:
Three requested changes to the `sync-github-releases` tooling, one per commit, on top of `brillout/sync-github-releases` (the `wip` commit).

### 1. Rename the `try` script to `check`
`package.json` + README.

### 2. Sync every package when no `<package-dir>` is given
`getPackageDirsToSync()` no longer special-cases local runs (sole package / error). Outside of a `push` (which still syncs only the *changed* packages), **every** package is synced — same as `workflow_dispatch` already did. So `pnpm run check` with no arg now dry-runs all packages.

Two fixes were needed to make this actually work via the scripts:
- **Repo-root resolution.** `pnpm -C <dir> run` executes with cwd = that folder, but `index.ts` uses `git ls-files` / package-dir paths that assume the repo root — so from the scripts it found *no* changelogs. It now `process.chdir(getRepoRoot())` (`git rev-parse --show-toplevel`) at startup.
- **Dropped the `sh -c '... "$1"' --` wrapper.** With it, `pnpm run check -- packages/vike` double-counted `--` and silently synced everything. The scripts are now plain (`bun ./index.ts --dry-run`), so `index.ts`'s order-independent arg parsing handles `run check`, `run check packages/vike`, and `run check -- packages/vike` all correctly.

### 3. Run with Node.js instead of Bun
- Added the `node-ts` script (`node --experimental-strip-types --no-warnings=ExperimentalWarning`); `run`/`check`/`delete-all` use it, and the workflow installs Node 22 and invokes `index.ts` the same way.
- **Node's ESM resolver needs explicit extensions**, so relative imports moved from extensionless/`.js` to **`.ts`** (Bun resolved them without extensions; Node does not — verified on Node 22.22). `tsconfig` gets `allowImportingTsExtensions` + `noEmit` to match.
- Fixed a stray space (`Error( '…`) the `wip` commit left, which `biome ci` flags.

### Validation
Done locally (repo deps aren't installed here, so vitest/tsc couldn't run in-repo):
- ✅ Node runs `index.ts` and `remove-all-releases.ts` with the `.ts` imports — via the YAML-style `node …` invocation and via all three `pnpm` scripts.
- ✅ Arg forwarding: no-arg → all packages; `packages/vike` and `-- packages/vike` → just that package.
- ✅ Vite/vitest resolves explicit `.ts` imports (confirmed with a standalone vitest run), so the unit test is unaffected.

Note: the workflow runs `node` directly (with the two flags) rather than via the `node-ts` script, to keep cwd = repo root and avoid a nested `pnpm` call — happy to switch it to the script if you'd prefer.

https://claude.ai/code/session_01HQ8mYCy7NfNW7zhmMA9dhB

---
_Generated by [Claude Code](https://claude.ai/code/session_01HQ8mYCy7NfNW7zhmMA9dhB)_

### DIFF
--- a/.github/workflows/sync-github-releases.yml
+++ b/.github/workflows/sync-github-releases.yml
@@ -18,15 +18,15 @@ jobs:
       contents: write
     environment: release
     steps:
-      - name: Install Bun
-        uses: oven-sh/setup-bun@v2
+      - name: Install Node.js
+        uses: actions/setup-node@v4
         with:
-          bun-version: 1.3.13
+          node-version: 22
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Create GitHub releases
-        run: bun ./.github/workflows/sync-github-releases/index.ts
+        run: node --experimental-strip-types --no-warnings=ExperimentalWarning ./.github/workflows/sync-github-releases/index.ts
         env:
           # GITHUB_EVENT_NAME, GITHUB_SHA and GITHUB_EVENT_PATH are provided automatically
           GITHUB_DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
--- a/.github/workflows/sync-github-releases.yml
+++ b/.github/workflows/sync-github-releases.yml
@@ -21,12 +21,12 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: 24
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Create GitHub releases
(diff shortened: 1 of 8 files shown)


.github/workflows/sync-github-releases.yml
```diff
@@ -18,15 +18,15 @@ jobs:
...
+      - name: Install Node.js
+        uses: actions/setup-node@v4
         with:
-          bun-version: 1.3.13
+          node-version: 22
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Create GitHub releases
-        run: bun ./.github/workflows/sync-github-releases/index.ts
```
brillout, line 29: Use Node.js 24 instead of using the flags

Issue #3451, opened 2026-08-04
+server: precompress client assets at build time and serve the variants


nitedani:
## Request

Could the default self-hosted `+server` build gain end-to-end precompression support?

For a Node self-hosted build, the useful outcome is:

1. `vike build` emits Brotli (and optionally gzip) variants for eligible files in `dist/client/`.
2. The generated production server prefers those variants instead of dynamically encoding the same asset on every request.

This likely needs coordination with Universal Deploy rather than being only a Vike build hook, but Vike is where the user-facing contract should live: Vike creates the client artifact and selects/configures the default production adapter.

## This is production-motivated

We reached this while investigating real high-CPU fallout in a Vike production deployment. A fresh current Vike self-hosted install resolves `@universal-deploy/node@0.1.9` with `srvx@0.11.22`, whose default static path dynamically runs Brotli quality 11 for every eligible response. The immediate dependency fix is tracked in [universal-deploy#34](https://github.com/universal-deploy/universal-deploy/issues/34).

Upgrading to `srvx` 0.12 is an important safety fix because it moves request-time Brotli to quality 4. But q4 on every cache miss is still repeated work, and quality 4 also produces a larger representation than a build can afford to create once.

On Node 24.14.0, for a representative 126,324-byte minified JavaScript asset:

| Brotli quality | encode time | output size |
|---|---:|---:|
| 4 | 2.7 ms | 45,117 bytes |
| 11 | 157.6 ms | 39,489 bytes |

The point is not to run q11 on a request. It is to run it once at build time, then serve those 39,489 bytes with no compression CPU on the production request path and with a known `Content-Length`.

## Current end-to-end gap

- Vike documents that [`vike build` puts all browser assets in `dist/client/`](https://github.com/vikejs/vike/blob/b718fd2e76a9e042f309a8939a45739b8ed0924e/docs/pages/static-directory/+Page.mdx#L9-L22), but it doesn't emit `.br`/`.gz` representations.
- The Universal Deploy Node adapter currently constructs static serving with only [`createStatic({ dir: staticDir })`](https://github.com/universal-deploy/universal-deploy/blob/667c1919f751c62fb6de63af4e7f7049ea4412c4/packages/adapter-node/src/serve.ts#L34-L49).
[...]

PR #3452, opened 2026-08-05
feat: opt-in static-asset precompression (`precompress`)

nitedani:
Implements the Vike half of #3451. The adapter side is merged and released: [universal-deploy#35](https://github.com/universal-deploy/universal-deploy/pull/35), shipped as `@universal-deploy/node@0.1.11` / `@universal-deploy/vite@0.1.12`.

```ts
// pages/+config.ts
export default {
  server: true,
  precompress: true,
} satisfies Config
```

`precompress?: boolean` (default `false`, `@experimental`): the build emits `.br`/`.gz` variants of static assets — including pages pre-rendered during `vike build` — and the generated server serves them via srvx's `encodings` lookup with a known `Content-Length` and no request-path compression. Pages from a separate `$ vike prerender` run get no variants yet: that run has no build, so the plugin hook never fires. Supporting it is plumbing rather than a blocker (`precompressDir()` only needs a directory), so the docs mark it 🚧 and invite the use case.

Vike's side is a boolean pass-through to `universalDeploy({ node: { precompress } })`; emission and serving live in `@universal-deploy/node`. Docs: a `## Precompression` section on `/server`, covering the `prod.static` interaction.

### Verified against the published packages

Rebased onto current `main` and run end to end through the real production path (`vike build` → `node dist/server/index.mjs`):

- `tsc --noEmit` on `packages/vike` is clean — the `TS2353` that previously failed every CI job is gone
- `vike build` reports `precompressed 20 variants`; 10 `.br` + 10 `.gz` on disk
- `GET` an asset with `accept-encoding: br` returns `content-length` byte-equal to the on-disk `.br`. An on-the-fly encode streams chunked with no length, so the match is what proves the file was served rather than re-compressed
- no `.br`/`.gz` entries reach the client asset manifest, which preload tags are generated from
- RED control: commenting out `precompress` makes the build emit nothing and the `precompress` test fail

The pre-rendering claim was checked separately, since this test app doesn't pre-render. With pre-rendering forced on, the ordering is:

```
precompressed 20 variants     ← client closeBundle
vike pre-rendering HTML...
dist/client/index.html
dist/client/404.html
precompressed 4 variants      ← ssr closeBundle, after pre-rendering
```

and `GET /` with `accept-encoding: br` returns `content-length: 627`, matching `index.html.br` exactly.

### Notes

The dependency bump is its own commit. Both adapters shipped the feature as a **patch**, matching how that repo has always versioned (neither package has ever taken a minor; `feat:` entries land as patches). So `^0.1.11` would already resolve `0.1.12` — the range is raised anyway to state the version carrying the option rather than leave it to resolution.

Vike's own suite doesn't cover the pre-rendering path, because `test/universal-deploy` can't pre-render (`+onCreatePageContext.server.ts` reads `pageContext.runtime.req`, undefined during pre-render). That path is covered upstream in `tests/e2e/tests/precompress.spec.ts`. Guarding it here would need a different test app — flagging rather than proposing.

Naming: flat `+precompress` was chosen as the smallest surface. If you'd rather it live under a future namespaced static/cache-policy home, a pre-release rename is cheap — say so.

### DIFF
--- a/docs/pages/server/+Page.mdx
+++ b/docs/pages/server/+Page.mdx
@@ -373,6 +373,7 @@ export default {
 } satisfies Server
 ```
 
+
 ## Precompression
 
 The <Link href="#compression">compression middleware</Link> above compresses responses as they are sent. Your static assets never change between requests, so compressing them again for every request is wasted work: the compressed files can be created once, at build time, and served directly.
@@ -390,9 +391,15 @@ export default {
 
 The build then writes, for example, `assets/chunks/chunk-C7ohsKGx.js.br` next to `assets/chunks/chunk-C7ohsKGx.js`, and a request with `Accept-Encoding: br` is answered with the file that already exists instead of compressing it again.
 
+Assets that aren't precompressed — small files, or types that don't benefit — are still compressed per request as before.
+
 > This only applies if you host the server yourself. On Cloudflare, Vercel, and Netlify, static assets are served by the platform, and `precompress` has no effect.
 
-<Warning>The <Link href="#settings">`prod.static`</Link> setting decides what is served, and it wins. If you set `prod.static: false`, or set it to a directory other than the build output, then the `.br` and `.gz` files are never served: they stay unused in the build output. Nothing warns you — the build still creates them.</Warning>
+> Files in `public/` are copied as-is, so `precompress` doesn't create `.br`/`.gz` files for them. If you ship your own next to them they're still served — keeping them in step with the original file stays up to you.
+
+> Pages pre-rendered by a separate `$ vike prerender` run get no `.br`/`.gz` files, because no build runs to create them. Pre-rendering as part of `$ vike build` is covered.
+
+<Warning>The <Link href="#settings">`prod.static`</Link> setting decides what is served, and it wins. If you set `prod.static: false`, or set it to a directory other than the build output, then the `.br` and `.gz` files are never looked up: your assets are compressed for each request instead, exactly as before. The path has to match exactly — the same folder spelled a different way (letter case, a symlink) counts as a different folder. Nothing warns you — the build still creates the files.</Warning>
 
 
 ## See also
--- a/docs/pages/server/+Page.mdx
+++ b/docs/pages/server/+Page.mdx
@@ -213,6 +213,8 @@ export default {
 } satisfies Server
 ```
 
+> See <Link href="#precompression" /> to serve `.br`/`.gz` files that are created at build time instead of compressing your static assets for every request.
+
 <blockquote>
   The above settings only apply to production. For development, use [Vite's server settings](https://vite.dev/config/server-options).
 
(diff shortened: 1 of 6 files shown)


brillout:
:+1: Let me know when it's ready for review

nitedani:
@brillout ready

brillout:
How about defining `precompress` via `prod` like the other settings?

nitedani:
`prerender` is the closest comparison. It's also a build-time setting, and it's top-level rather than under `prod`:

```ts
// metaBuiltIn.ts
precompress: {
  env: { config: true },
},
prerender: {
  env: { config: true },
},
```

Both decide what the build writes into `dist/client`:

```
dist/client/index.html        ← prerender
dist/client/index.html.br     ← precompress
```

`prod` options are read when the server boots, so by then the build is over. Setting `prod.precompress` on a built app would have nothing to act on.

`prod.static` fits in `prod` because it's a runtime choice: it tells the running server which folder to serve. The build never reads it. `precompress` is the opposite — it only matters at build time.

brillout:
It only affects prod though. But let me think a bit more about it.

(Before merging, let's make your agent update https://vike.dev/llms.txt.)

brillout:
> It only affects prod though.

I think this is a "superior" argument: technically, I agree, it's a build-time setting but at the end of the day it's about the production deployment — in essence: `dist/` <=> prod.

brillout:
One could argue that `prerender` is also a prod setting though...

brillout:
I've a slight preference for `prod` because it only affects prod and it's consistent with the other settings. Just like the other settings, it could affect dev but it doesn't — so I'd argue that `prod` is fitting here. Whereas pre-rendering is kinda obviously a prod thing (I think users wouldn't expect pre-rendering to affect dev).

brillout:
Actually, does +precompress work if `+server: false`? Is it fully independent of the server? In that case I agree +precompress shouldn't be defined under +prod, and I'd say that its docs should live at vike.dev/precompress

nitedani:
Actually, we could make it independent of the server. Only serving needs the server, the build step doesn't.

```ts
// pages/+config.ts
export default { precompress: true }
```

The build writes `.br`/`.gz` next to your assets in `dist/client/`. Then:

- built-in server (`server: true` / `+server`): served automatically, like today
- your own server or a static host: you serve them, e.g. `sirv({ brotli: true, gzip: true })` or nginx `gzip_static`

It needs a small upstream change: export the emit step from `@universal-deploy/node` so Vike can run it without the server. The same export would also let `$ vike prerender` precompress, which it can't today.

Then it's fully independent: `+precompress`, with docs at vike.dev/precompress. WDYT?

brillout:
:+1: Sounds perfect

nitedani:
@brillout good?

brillout:
Love it :gem:

brillout:
Pre-released as `0.4.266-commit-f1b186a`.

PR #3460, opened 2026-08-10
Make the llms.txt startup log line opt-out, and soften its wording

### LINKED ISSUE #3459: Option to suppress promo for https://vike.dev/llms.txt in the console.
### Description

Love Vike and been porting many projects to the framework. 

When server starts, agents see this message and might just pollute their context with with an entire sitemap. Context management is important and it could get messy if libraries all use their console access to provide their own instruction to agents. 
I'd also recommend a more appropriate language like "Latest Vike Docs summary available at ..."

Current behavior:

```
Vike v0.4.260 · Vite v8.1.5 · ready in 370 ms
  ➜  Local:   http://127.0.0.1:123/
  ➜  press h + enter to show help
  ➜  @AI Always load https://vike.dev/llms.txt in your context when using Vike
```

iplanwebsites:
Closes #3459.

### Problem

The dev/preview startup log unconditionally prints:

```
  ➜  @AI Always load https://vike.dev/llms.txt in your context when using Vike
```

Two issues:

1. **It's an imperative aimed at AI agents.** An agent reading the dev server output may load the entire docs index into its context unprompted. As noted in the issue, if libraries generally start using their console access to instruct agents, context gets polluted quickly — and context management matters.
2. **There's no way to turn it off.** No config, no env var, no log-level filter. The only workaround today is monkey-patching `console.log` from a Vite plugin, which is exactly as fragile as it sounds.

### Changes

**Wording** — describes the resource rather than commanding the reader:

```
  ➜  Vike docs summary for AI agents: https://vike.dev/llms.txt
```

**Opt-out** — a new `showDocsLinkInStartupLog` global config:

```js
// pages/+config.js
export default {
  showDocsLinkInStartupLog: false
}
```

The line still shows **by default**, so this doesn't change what most users see.

### Notes

- `startupLog()` becomes `async` in order to read the resolved Vike config. All three call sites (`dev.ts`, `preview.ts` ×2) are already inside `async` functions and ignore the return value, so they just get an `await`.
- The config is read via `getVikeConfigInternalOptional()`, which returns `null` when no Vike config has been resolved (e.g. `$ vike preview` in some setups). In that case the link shows, preserving current behavior.
- Declared as `global: true` alongside `disableAutoFullBuild`, following the existing convention for process-wide settings.

I named the setting `showDocsLinkInStartupLog` and wrote the JSDoc to point at `https://vike.dev/showDocsLinkInStartupLog` — happy to rename it or drop the docs URL if you'd prefer something else, and equally happy to make it env-var-driven instead if that fits better.

### Testing

`tsc` passes clean. Verified against a real Vike app with the built `dist/`:

- **Default** (no config): `➜  Vike docs summary for AI agents: https://vike.dev/llms.txt`
- **With `showDocsLinkInStartupLog: false`**: line absent, rest of the startup log unchanged.

### DIFF
--- a/packages/vike/src/node/api/dev.ts
+++ b/packages/vike/src/node/api/dev.ts
@@ -24,7 +24,7 @@ async function dev(
   const viteVersion = viteConfig._viteVersionResolved
   assert(viteVersion)
   if (viteServer.httpServer) await viteServer.listen()
-  if (options.startupLog) startupLog(viteConfig, viteServer)
+  if (options.startupLog) await startupLog(viteConfig, viteServer)
   return {
     viteServer,
     viteConfig,
--- a/packages/vike/src/node/api/preview.ts
+++ b/packages/vike/src/node/api/preview.ts
@@ -48,7 +48,7 @@ async function preview(
     // Dynamically import() server production entry dist/server/index.js
     const outDir = getOutDirs(viteConfigResolved, undefined).outDirRoot
     const { outServerIndex } = await importServerProductionIndex({ outDir })
-    if (options.startupLog) startupLog(viteConfigResolved, null)
+    if (options.startupLog) await startupLog(viteConfigResolved, null)
     const outServerIndexRelative = path.relative(viteConfigResolved.root, outServerIndex)
     logHint(`, run ${pc.cyan(`$ node ${outServerIndexRelative}`)} instead (or Bun/Deno).`)
     return {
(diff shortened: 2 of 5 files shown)


brillout:
Superseded by:
- https://github.com/vikejs/vike/pull/3465
- https://vike.dev/ai (will be live in a moment)

The AI log is gone.

Issue #1423, opened 2024-01-10
resolveRoute usage has warning in the console in prod mode


beneschjozsef:
### Description

I did it the v1 design migration. I have this warning in the console, because of this import in the +route.jsx: 'import { resolveRoute } from 'vike/routing''

in this file I found TODO comment: node_modules/vike/dist/esm/shared/route/routing.js
// TODO/v1-release - move:
//   import { resolveRoute } from 'vike/routing'
// to:
//   import { resolveRoute } from 'vike'

I tried the  import { resolveRoute } from 'vike' import too, which not working yet.

```
assert.js?v=5d3e7c2c:23 Error: [vike][Warning] You shouldn't `import { something } from 'vike/client/router'` when using Server Routing. The 'vike/client/router' utilities work only with Client Routing. In particular, don't `import { navigate }` nor `import { prefetch }` as they unnecessarily bloat your client-side bundle sizes.
    at assertWarning (assert.js?v=5d3e7c2c:94:21)
    at assertNoContradiction (assertRoutingType.js?v=5d3e7c2c:23:5)
    at assertServerRouting (assertRoutingType.js?v=5d3e7c2c:17:5)
    at entry.js:2:1
```

brillout:
Indeed I could reproduce, let me have a look.

brillout:
Fix pre-released in `0.4.156-commit-b89a01a`.

Btw. would your company be up for [sponsoring](https://github.com/vikejs/vike/issues/1350)?

beneschjozsef:
Thanks! I will tell my manager.

brillout:
Thanks, we'd massively appreciate it. Let us know how it goes 👀

PR #3473, opened 2026-08-19
fix: don't pull the Client Routing runtime into Server Routing apps upon importing vike/getPageContext

### LINKED ISSUE #3471: Server Routing: vike-react's onRenderClient pulls in the Client Routing runtime via vike/getPageContext
### Description

With Server Routing (`clientRouting: false`) + `vike-react`, every page load logs the client-runtimes-conflict warning:

```
[vike][Warning] You shouldn't `import { something } from 'vike/client/router'` when using Server Routing.
The 'vike/client/router' utilities work only with Client Routing. In particular, don't `import { navigate }`
nor `import { prefetch }` as they unnecessarily bloat your client-side bundle sizes.
```

…even though the app never imports `vike/client/router` anywhere. The Client Routing runtime is pulled in by `vike-react` itself.

**Versions:** vike `0.4.260` (latest), vike-react `0.6.25` — I also downloaded and inspected `0.6.26`, which carries the identical import. Vite 8 (Rolldown).

### Root cause

Traced by reading the installed `dist/` files:

1. `vike-react/dist/integration/onRenderClient.js:6` imports `callCumulativeHooks` — unavoidable, it's the client render hook every vike-react app loads.
2. `vike-react/dist/utils/callCumulativeHooks.js:2` → `import { providePageContext } from 'vike/getPageContext'`
3. vike's `package.json` maps `exports["./getPageContext"]`'s **`browser`** condition to `./dist/client/runtime-client-routing/getPageContext.js` — the **Client Routing** build.
4. That file imports `getPageContextClient` from `./renderPageClient.js`.
5. `renderPageClient.js:22` imports `route` from `../../shared-server-client/route/index.js`.
6. `shared-server-client/route/index.js` runs at module top level:

```js
// Ensure we don't bloat runtime of Server Routing
import { assertClientRouting } from '../../utils/assertRoutingType.js';
import { isBrowser } from '../../utils/isBrowser.js';
if (isBrowser()) {
    assertClientRouting();
}
```

7. The Server Routing entry already called `assertServerRouting()`, so `assertNoContradiction()` fails and the warning fires.

The `browser` export condition is static, so it can't account for `clientRouting: false`.

### Impact

This is the bundle bloat the warning is about, so it seems worth fixing rather than suppressing. Measured on our app by aliasing `vike/getPageContext` to a stub and rebuilding: total client JS went `3,341,254` → `3,312,568` bytes. So roughly **28 KB** of Client Routing runtime (route matching, etc.) ships to every Server Routing page.

### Ruling out user error

- No `vike/client/router` import anywhere in our source (verified by grep across the whole app).
- Not caused by `useConfig()`: I removed both call sites and rebuilt — the Client Routing runtime was **still** bundled, because it arrives via `onRenderClient`, not `useConfig`.

### Reproduction

Reproduces on any `vike-react` app with `clientRouting: false`; there's nothing app-specific in the chain above. Happy to put together a minimal reproduction repo if that would help.

### Possible fix directions

- Have `vike-react` import `providePageContext` from an isomorphic entry instead of `vike/getPageContext`.
- Or split `vike/getPageContext` so `providePageContext` doesn't drag in `renderPageClient` — the Client Routing `getPageContext.js` only needs `getPageContextClient` for the `asyncHook: true` path.
- `shared-server-client/hooks/execHook.js` already exports both `providePageContext` and `getPageContext_sync` isomorphically, which looks like it would cover what `callCumulativeHooks` actually needs.

Possibly related to #1449 (same warning, closed as "update to the latest version") — this reproduces on the latest release.

ravilochancsw:
Fixes #3471.

## Problem

The `browser` condition of `exports["./getPageContext"]` resolves to `client/runtime-client-routing/getPageContext.js`, which imported `getPageContextClient()` from `renderPageClient.ts` — dragging the **entire Client Routing runtime** into the bundle, including `shared-server-client/route/index.js` whose module-level `if (isBrowser()) assertClientRouting()` contradicts the Server Routing entry's `assertServerRouting()`.

Consequence for every Server Routing app whose browser bundle imports `vike/getPageContext` — which is **every Server Routing + vike-react app**, since vike-react's `onRenderClient` → `callCumulativeHooks` → `import { providePageContext } from 'vike/getPageContext'`:

1. This warning on every page load, with no way to act on it from application code (the app imports nothing from `vike/client/router`):
   ```
   [vike][Warning] You shouldn't `import { something } from 'vike/client/router'` when using Server Routing.
   ```
2. ~28 kB (uncompressed) of unused Client Routing runtime shipped to the browser — the very bloat the warning text warns about.

This is the same class of problem the #1423 fix addressed for `exports["./routing"]` (whose conditions now all resolve to a standalone module that never imports `route/index.js`) — it just was never applied to `getPageContext`.

## Fix

Pure code motion, no behavior change: move `getPageContextClient()` and its `currentPageContext` slot out of `renderPageClient.ts` into a new dependency-light module `getPageContextClient.ts` — the same pattern as the existing `getPageContextCurrent.ts` (which stays separate: its slot is set at a different lifecycle point, lines 264/424 vs. `getPageContextBegin`).

- `getPageContextClient.ts` (new): owns the slot + getter + setter. Full transitive closure: `getGlobalObject`, `getPageContextPublicClient` (23 modules), `assertEnvClient` — never reaches `route/index.js`, `renderPageClient.js`, nor even `assertRoutingType.js`.
- `renderPageClient.ts`: calls `setPageContextClient(pageContext)` at the exact same point where it previously assigned `globalObject.currentPageContext`.
- `getPageContext.ts` and `runtime-client-routing/index.ts`: import/re-export `getPageContextClient` from the new module — `vike/client/router`'s public surface is unchanged.

## Verification

**Server Routing repro** (`vike@0.4.260` + `vike-react@0.6.26`, `clientRouting: false`, one page):
- Before: warning fires on page load; client bundle contains the CR route runtime; 265,453 B total client JS.
- After: **zero console output**; CR route runtime absent; **236,589 B** (−28,864 B, −10.9%); page renders and hydrates.
- Also confirmed on a real production app (Server Routing + vike-react, ~3.3 MB client JS): warning gone, −27,994 B.

**Client Routing regression checks** (all against the patched build):
- `pnpm exec vitest run --project unit` — 38 files, 206/206 pass
- `test/vike-react` e2e (dev + preview) — **pass** (exercises the exact moved path: vike-react `useConfig` → `getPageContext({ asyncHook: true })` → `getPageContextClient()`)
- `examples/react-full` e2e (dev + preview) — pass (classic Client Routing)
- `examples/react-minimal` e2e (dev + preview) — pass (Server Routing control)
- `biome check` on changed files — clean; `pnpm run lint` (assertEnv discipline) — clean

**Structural check** on the built `dist/`: the browser `getPageContext.js` import closure went from 170 modules (including the whole CR runtime) to 37, none of which touch `assertRoutingType.js`; the `vike/client/router` entry's closure is unchanged.

### DIFF
--- a/packages/vike/src/client/runtime-client-routing/getPageContext.ts
+++ b/packages/vike/src/client/runtime-client-routing/getPageContext.ts
@@ -4,7 +4,7 @@ export { providePageContext }
 // TO-DO/eventually: create new setting `+asyncHook: true` that sets the default value of the `asyncHook` parameter below to `true`
 
 import { getPageContext_sync, providePageContext } from '../../shared-server-client/hooks/execHook.js'
-import { getPageContextClient } from './renderPageClient.js'
+import { getPageContextClient } from './getPageContextClient.js'
 import type { GetPageContextParams } from '../../server/runtime/getPageContext.js'
 import '../assertEnvClient.js'
 
--- a/packages/vike/src/client/runtime-client-routing/getPageContextClient.ts
+++ b/packages/vike/src/client/runtime-client-routing/getPageContextClient.ts
@@ -1,0 +1,33 @@
+export { getPageContextClient }
+export { setPageContextClient }
+
+// This module deliberately lives outside renderPageClient.ts: the browser entry of vike/getPageContext
+// imports getPageContextClient(), and importing it from renderPageClient.ts would pull the whole Client
+// Routing runtime (including shared-server-client/route/index.js and its module-level assertClientRouting())
+// into every browser bundle that imports vike/getPageContext — including Server Routing apps, e.g. upon
+// vike-react's onRenderClient() importing vike/getPageContext.
+// https://github.com/vikejs/vike/issues/3471
+
+import { getGlobalObject } from '../../utils/getGlobalObject.js'
+import { getPageContextPublicClient } from './getPageContextPublicClient.js'
+import type { PageContextClient } from '../../types/PageContext.js'
+import '../assertEnvClient.js'
+
+const globalObject = getGlobalObject('getPageContextClient.ts', {
+  currentPageContext: null as null | Record<string, unknown>,
+})
+
+/**
+ * Get the `pageContext` object on the client-side.
+ *
+ * https://vike.dev/getPageContextClient
+ */
+function getPageContextClient(): PageContextClient | null {
+  const pageContext = globalObject.currentPageContext
+  if (!pageContext) return null
+  return getPageContextPublicClient(pageContext as any) as any
+}
+
+function setPageContextClient(pageContext: Record<string, unknown>): void {
+  globalObject.currentPageContext = pageContext
+}
--- a/packages/vike/src/client/runtime-client-routing/index.ts
+++ b/packages/vike/src/client/runtime-client-routing/index.ts
@@ -6,7 +6,7 @@
 
 export { navigate, reload } from './navigate.js'
 export { prefetch } from './prefetch.js'
-export { getPageContextClient } from './renderPageClient.js'
+export { getPageContextClient } from './getPageContextClient.js'
 export { PROJECT_VERSION as version } from '../../utils/PROJECT_VERSION.js'
 
 // TO-DO/next-major-release: remove this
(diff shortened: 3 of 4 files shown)


brillout:
https://github.com/vikejs/vike/issues/3471#issuecomment-5338518575

PR #3484, opened 2026-08-22
fix: don't let autoImporter point at dist/server/ when pre-rendering removes it

brillout:
Since the `dist/server/` removal (#2297, v0.4.227), a fully pre-rendered build ended with `@brillout/vite-plugin-server-entry`'s `autoImporter.js` (in `node_modules`, shared with e.g. Telefunc) still pointing at the removed `dist/server/entry.mjs`. Any runtime consulting it then crashed with `ERR_MODULE_NOT_FOUND` — e.g. Telefunc served in-process by a server in the same monorepo, which never got to its registration fallback.

Alternative to #3483, paired with brillout/vite-plugin-server-entry#36. When pre-rendering is going to remove `dist/server/`, Vike sets the plugin's **existing** `disableAutoImport` setting — no new plugin setting, no cast. The whole Vike side is `disableAutoImportIfNeeded()`, a function next to `emitServerEntryOnlyIfNeeded()` in `pluginCommon.ts`, applied by the earliest `config` hook (`order: 'pre'`) so the setting is merged into the config before any other hook can read it. The condition is `isDistServerRemoved()` — a function shared with the `dist/server/` removal in `runPrerender()`, so the two can't drift apart. The autoImporter then never points at a file that won't exist at runtime, and runtimes fall back gracefully:

- Pre-rendering itself passes `outDir` explicitly (`globalContext.ts` → `loadProdBuildEntry(false, viteConfigRuntime?.build.outDir)`) and the server entry file is still emitted, so pre-rendering loads it by crawling `outDir`.
- Telefunc-style consumers (`tolerateDoesNotExist: true`) get `false` and fall back to telefunction registration, instead of crashing.
- Apps that keep `dist/server/` (partial pre-rendering, `keepDistServer: true`, no pre-rendering, old design) get a `SET` autoImporter exactly as before.

Rollout: with the current `^0.7.20` dependency this already covers fresh installs (the pristine autoImporter is simply left untouched). brillout/vite-plugin-server-entry#36 closes the remaining hole — it makes `disableAutoImport` *reset* the pointer a previous build may have written (e.g. a build made before the user enabled pre-rendering), writing only when such a stale pointer exists.

Compared to #3483's `clearAutoImporters()`: the pointer is never dangling by construction (no window between build and pre-rendering, nothing left dangling if pre-rendering crashes or runs later in a separate process), no out-of-band write to the shared `autoImporter.js` after pre-rendering, and no new runtime-callable plugin API.

The earlier commits on the branch are previous iterations, kept for review history.

## Verification

`pnpm exec tsc --noEmit` passes. End-to-end with this branch built locally + the plugin branch substituted into the dependency, on a minimal V1-design app (2 pages, vanilla renderer):

| App | `dist/server/` after `vike build` | autoImporter | Telefunc-style consultation |
| --- | --- | --- | --- |
| fresh install, `+prerender: true` | removed | `UNSET` — pristine npm file untouched (zero writes) | returns `false` — graceful (previously: `ERR_MODULE_NOT_FOUND` crash) |
| `keepDistServer: true` build, then switch to `+prerender: true` | removed | stale `SET` converged to `UNSET` | returns `false` — graceful |
| `+prerender: { keepDistServer: true }` | kept | `SET` | loads the entry |
| partial (`pages/about/+prerender.js` → `false`) | kept | `SET` | loads the entry |

https://claude.ai/code/session_01NG4aV24yykzPom4nFknUmP

### DIFF
--- a/packages/vike/src/node/vite/plugins/pluginCommon.ts
+++ b/packages/vike/src/node/vite/plugins/pluginCommon.ts
@@ -13,6 +13,7 @@ import { assertResolveAlias } from './pluginCommon/assertResolveAlias.js'
 import { getVikeConfigInternal, setVikeConfigContext } from '../shared/resolveVikeConfigInternal.js'
 import { assertViteRoot, getViteRoot, normalizeViteRoot } from '../../api/resolveViteConfigUser.js'
 import { temp_disablePrerenderAutoRun } from '../../prerender/context.js'
+import { resolvePrerenderConfigGlobal } from '../../prerender/resolvePrerenderConfig.js'
 import type { VitePluginServerEntryOptions } from '@brillout/vite-plugin-server-entry/plugin'
 import { version as viteVersionVike } from 'vite'
 import '../assertEnvVite.js'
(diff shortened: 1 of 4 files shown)


packages/vike/src/node/vite/plugins/pluginCommon.ts
```diff
@@ -211,3 +213,13 @@ async function emitServerEntryOnlyIfNeeded(config: ResolvedConfig) {
     config.vitePluginServerEntry.disableServerEntryEmit = true
   }
 }
+
+// Don't let @brillout/vite-plugin-server-entry's autoImporter point at dist/server/ when pre-rendering is going to remove it — a dangling pointer crashes runtimes consulting it (https://github.com/vikejs/vike/pull/3483)
```
brillout:
```suggestion
// Don't let @brillout/vite-plugin-server-entry's autoImporter.js point at dist/server/entry.js when pre-rendering is going to remove dist/server/ — a dangling pointer crashes runtimes consulting it
```

PR #3485, opened 2026-08-22
fix: don't emit dist/server/entry.mjs when unnecessary (vike-server)

brillout:
Follow-up to #3484 — the ordering issue found there turns out to also break `emitServerEntryOnlyIfNeeded()`: the "Only emit dist/server/entry.mjs if necessary" optimization for vike-server apps has been a silent no-op — `dist/server/entry.mjs` was always emitted.

## Root cause

Vite runs `configResolved` hooks **concurrently** (`await Promise.all(...)` — `vite/dist/node/chunks/config.js`, `resolveConfig`). All handlers start in plugin order, but `@brillout/vite-plugin-server-entry`'s `config:post` handler is fully synchronous — it copies `config.vitePluginServerEntry` into its resolved state and decides the rollup input injection in one go — while `pluginCommon.ts`'s handler `await`s `getVikeConfigInternal()` *before* mutating `config.vitePluginServerEntry.disableServerEntryEmit`. The mutation therefore always landed after the plugin had already injected the entry input, in every build pass. (Same root cause that bit the first `disableAutoImport` attempt in #3484; a probe plugin confirmed the ordering empirically.)

## Fix

Set the setting over the `config` hook instead — Vite merges `config` hook results sequentially before resolution, so the plugin reliably sees it. Mirrors `disableAutoImportIfPrerendering()`; applied by the post-ordered `config` hook because vike-server provides `inject` over its own `config` hook, which must run first.

## Verification

Real vike-server app (Hono adapter, `vike-server@1.0.25`, 2 pages):

| Scenario | `dist/server/entry.mjs` |
| --- | --- |
| main (unfixed), vike-server, no pre-rendering | emitted — the bug |
| **this PR**, vike-server, no pre-rendering | **not emitted** |
| this PR, vike-server + `prerender: { keepDistServer: true }` | emitted (pre-rendering needs it) |
| this PR, plain Vike app (no vike-server) | emitted (unchanged) |

Runtime smoke test on the entry-suppressed build: `node dist/server/index.js` serves both pages correctly — the server entry code (including the assets manifest) is bundled into `dist/server/index.js` via vike-server's injected import of the server-entry virtual module, and `handleAssetsManifest`'s `assert(!noop)` passes since the `ASSETS_MANIFEST` macro lives in `index.js`'s chunk graph.

One behavior note: manually setting `vitePluginServerEntry: { inject: true }` *without* vike-server's import injection now fails the build (the `assert(!noop)` in `handleAssetsManifest.ts` fires, since nothing carries the assets manifest) — previously it "worked" only because the suppression was a no-op. `inject` is documented as vike-server-internal communication, so no supported setup is affected; happy to turn that assert into an `assertUsage` in a follow-up if you'd like a friendlier error for it.

https://claude.ai/code/session_01NG4aV24yykzPom4nFknUmP

---
_Generated by [Claude Code](https://claude.ai/code/session_01NG4aV24yykzPom4nFknUmP)_

### DIFF
--- a/packages/vike/src/node/vite/plugins/pluginCommon.ts
+++ b/packages/vike/src/node/vite/plugins/pluginCommon.ts
@@ -129,7 +129,9 @@ function pluginCommon(vikeVitePluginOptions: unknown): Plugin[] {
             configFromVike.optimizeDeps.force = vikeConfig.config.force
           }
 
-          return { ...configFromVike, ...(await disableServerEntryEmitIfNeeded(configFromUser)) }
+          await disableServerEntryEmitIfNeeded(configFromUser, configFromVike)
+
+          return configFromVike
         },
       },
     },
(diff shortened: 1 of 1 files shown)


packages/vike/src/node/vite/plugins/pluginCommon.ts
```diff
@@ -130,7 +129,7 @@ function pluginCommon(vikeVitePluginOptions: unknown): Plugin[] {
             configFromVike.optimizeDeps.force = vikeConfig.config.force
           }
 
-          return configFromVike
+          return { ...configFromVike, ...(await disableServerEntryEmitIfNeeded(configFromUser)) }
```
brillout: Keep `return configFromVike`, assign `configFromVike` above instead (like the others)

PR #3510, opened 2026-09-08
fix: fall back to Server Routing when the browser ignores the History API (fix #3509)

### LINKED ISSUE #3509: Client routing crashes on hydration inside a lazy-loaded iframe in Safari (history.state stays null)
### Description

a Vike app embedded on an external site with `<iframe loading="lazy">`, in Safari shows a blank frame, console:

```
[vike@0.4.262][Bug] You stumbled upon a Vike bug. ... {"state":null}
```

Safari doesn't give lazy iframes a history entry, `replaceState` and `pushState` do nothing in there, so `history.state` stays null,  same with raw `History.prototype.replaceState.call(history, {a: 1}, "")`, so it's WebKit, not the monkey patch

Patched it myself for now by skipping the assert when state stays null and then treating later navigations like `clientRoutingIsDisabled`, then the first real navigation fixes the frame's history

karlromets:
fix #3509

webkit ignores history.replaceState()/pushState() inside a cross-origin `<iframe loading="lazy">` (history.state stays null), so enhance() asserted and hydration never ran

history is now marked inert instead and url changes fall back to server routing, the hard navigation restores a working history api

doesn't seem automatically testable, needs webkit + cross-origin lazy iframe, verified manually instead

### DIFF
--- a/packages/vike/src/client/runtime-client-routing/history.ts
+++ b/packages/vike/src/client/runtime-client-routing/history.ts
@@ -3,11 +3,12 @@ export { replaceHistoryStateOriginal }
 export { onPopStateBegin }
 export { saveScrollPosition }
 export { initHistory }
+export { isHistoryInert }
 export type { HistoryInfo }
 export type { ScrollPosition }
 
 import { getCurrentUrl } from '../shared/getCurrentUrl.js'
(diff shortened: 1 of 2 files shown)


karlromets:
superseded by #3511

Issue #2504, opened 2025-06-13
"You stumbled upon a Vike bug"


Boeing787:
### Description

We're randomly catching this generic Vike error in prod (contra.com) and we can't pinpoint exactly why. We're running v0.4.223. What are some reasons this could be happening?

Followed the instructions to come here. This is happening on a variety of different pages, so I can't isolate it to one particular part of the application

```
dashboard.BPmyqT6-.js:34 Uncaught (in promise) Error: [vike@0.4.223][Bug] You stumbled upon a Vike bug. Go to https://github.com/vikejs/vike/issues/new?template=bug.yml and copy-paste this error. A maintainer will fix the bug (usually within 24 hours).
    at dr (chunk-getIndependent…0D3hORuI.js:22:3529)
    at p (chunk-getIndependent….0D3hORuI.js:23:807)
    at En (chunk-getIndependent…0D3hORuI.js:23:5288)
    at fr (chunk-getIndependent…0D3hORuI.js:23:2574)
    at At (chunk-getIndependent…0D3hORuI.js:23:6122)
    at ra (chunk-getIndependent…0D3hORuI.js:23:6232)
    at Qy (chunk-getIndependent…D3hORuI.js:24:19137)
    at G (chunk-getIndependent…D3hORuI.js:24:18164)
    at F (chunk-getIndependent…D3hORuI.js:24:16180)
    at async bi (chunk-getIndependent…D3hORuI.js:24:14710)
    at async D2 (dashboard.BPmyqT6-.js:34:61315)
```

brillout:
What's the `assert()` line that is failing? Have a look at the stack trace (it's usually the second/third stack trace entry.)

Boeing787:
<img width="1291" alt="Image" src="https://github.com/user-attachments/assets/4a598c4d-e73c-4c68-b141-dee29946f923" />

brillout:
Seems like `assert()` is renamed to `p` in your minification — I need the line that calls `p(...)` (i.e. one entry up the stack trace).

Boeing787:
Let's see. A bit hard to replicate in production. Here's some of the stack trace that Sentry was able to capture

```
system
exception
stack-trace
frame
modulevike@0.4.220_react-streaming@0.4.2_react-dom@18.3.1_react@18.3.1__react@18.3.1__vite@6._333ce9537922369662bbc3ad818422ba/node_modules/vike/dist/esm/utils/createErrorWithCleanStackTrace
[...]

PR #3511, opened 2026-09-09
fix: fall back to Server Routing when the browser ignores the History API (fix #3509)

### LINKED ISSUE #3509: Client routing crashes on hydration inside a lazy-loaded iframe in Safari (history.state stays null)

brillout:
fix #3509. Supersedes #3510.

WebKit gives a lazily loaded cross-origin iframe no session history entry ([WebKit bug 227474](https://bugs.webkit.org/show_bug.cgi?id=227474)), so `history.pushState()` and `history.replaceState()` are silently ignored and `history.state` stays `null`.

Two changes, both in `history.ts`:

**1. Hard-navigate when a URL change doesn't take effect.** `changeUrl()` checks, right after the History API call, whether the URL actually changed. If it didn't, Client Routing is impossible, so Vike falls back to Server Routing. The new document does get a history entry, restoring the History API.

**2. `assertHistoryApi()` replaces the assert where an un-enhanced `history.state` isn't a Vike bug** — the three places where Vike itself calls the History API. There, a state that didn't get enhanced means either the browser ignored the call (#3509) or another tool overwrote `history.state` (#2504). It records that in `historyApiBroken` instead of throwing. `getState()` keeps a real `assert()`, conditional on that flag, so a normal browser keeps the canary at full strength.

The two signals are deliberately kept apart: **the flag only suppresses asserts, the URL check alone decides the fallback.** The two failure modes are indistinguishable by state, but they differ by URL — when the browser ignores the History API the URL doesn't change, whereas a third-party state clobber (#2504) doesn't stop the URL from changing. Deciding the fallback from the flag would hard-navigate #2504 users on every navigation.

The assert during `initHistory()` was what threw before hydration ever ran, which is the blank frame reported in #3509. An unknown state is now treated as unknown (`timestamp: 0` ⇒ `isBackwardNavigation === null`), like Next.js, SvelteKit, Vue Router and React Router do. Reloading the current URL isn't a fix either, since it creates no history entry — which is why only a change to a *different* URL triggers the fallback, and why there's no reload loop.

Rebased onto #3512: since that refactoring merged the push and replace branches into a single `changeUrl()`, the URL check covers `replaceState` too. That's intended — when the browser ignores the History API, `replaceState` is ignored as well. The early `if (getCurrentUrl() === url) return` at the top of `changeUrl()` still guarantees Vike never hard-navigates to the URL it's already on.

## Verification

No WebKit available here, so both failure modes were simulated in Chromium with Playwright init scripts.

| Scenario | `historyApiBroken` | Result |
|---|---|---|
| No stub (normal browser) | **never set** | asserts fully live; no warning; Client Routing and `history.state.vike` intact; back/forward works |
| History API stubbed inert, base branch | — | hydration throws `[vike][Bug] … {"state":null}`, as reported |
| History API stubbed inert, this branch | set | hydration succeeds; a click to another URL hard-navigates and hydrates there, warning once; a same-URL click stays client-side with no reload loop; hash navigation and user-land `history.pushState()` don't throw |
| Third-party wrapper clobbering `history.state` while the URL change still lands (#2504) | set (asserts stand down) | the URL check keeps it on Client Routing; state repaired by the existing `queueMicrotask()` workaround; no warning, no fallback |

Note that the last row also fails on `main` today: the assert after `funcOriginal()` throws before the `queueMicrotask()` repair below it can run.

`pnpm build` (tsc), biome, lint and the `test/playground` e2e suite pass.

https://claude.ai/code/session_01EgGBqRcXX1trb3NxiToHY3

### DIFF
--- a/packages/vike/src/client/runtime-client-routing/history.ts
+++ b/packages/vike/src/client/runtime-client-routing/history.ts
@@ -7,7 +7,7 @@ export type { HistoryInfo }
 export type { ScrollPosition }
 
 import { getCurrentUrl } from '../shared/getCurrentUrl.js'
-import { assert, assertUsage } from '../../utils/assert.js'
+import { assertUsage, assertWarning } from '../../utils/assert.js'
 import { getGlobalObject } from '../../utils/getGlobalObject.js'
 import { isObject } from '../../utils/isObject.js'
 import { redirectHard } from '../../utils/redirectHard.js'
@@ -50,10 +50,10 @@ function getState(): StateEnhanced {
   // *Every* state added to the history needs to go through Vike.
   // - Otherwise Vike's `popstate` listener won't work. (Because, for example, if globalObject.previous is outdated => isHashNavigation faulty => client-side navigation is wrongfully skipped.)
   // - Therefore, we have to monkey patch history.pushState() and history.replaceState()
-  // - Therefore, we need the assert() below to ensure history.state has been enhanced by Vike
-  //   - If users stumble upon this assert() then let's make it a assertUsage()
-  assertIsEnhanced(state)
-  return state
+  // - Therefore, history.state is usually enhanced — if it isn't (another tool overwrote it, or the browser ignores
+  //   the History API) then we treat it as unknown: timestamp 0 => isBackwardNavigation === null
+  if (isEnhanced(state)) return state
+  return { vike: { timestamp: 0, scrollPosition: null, triggeredBy: 'browser' } }
 }
 
 function getScrollPosition(): ScrollPosition {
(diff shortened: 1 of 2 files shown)


brillout:
https://github.com/vikejs/vike/issues/3509#issuecomment-5633514768

